PR #44, opened 2017-06-07
Skip EC testcase in case a mechanism is not supported

ifranzki:
When running with a CEXxP card that does not support a newer
mechanism, such as ECDSA with SHA256, the testcase is now
skipped and does no longer fail.

The second patch is for a problem with RSA PKCS PSS testcase failing with the ICA token, since the current version of libica does not allow to request zero bytes of random when libica is built with FIPS support. 


dodys:
This PR was merged, find below the commit ids:
2de0d27
b99196b

Thanks

PR #58, opened 2017-07-26
Fix ock_test to not check if root is in group pkcs11

dodys:
Root user should be able to build and run opencryptoki without being part of
group pkcs11. Only non-root users should be in the pkcs11 group.

Signed-off-by: Eduardo Barretto <ebarretto@linux.vnet.ibm.com>


pvital:
Testing 123

PR #59, opened 2017-07-26
Update pkcsslotd.service.in

markkp:
Update pkcsslotd.service.in to use After=local-fs.target instead of syslog.target. The syslog target doesn't exist any more.


dodys:
Please sign your commit

PR #61, opened 2017-08-02
Fix bug when parsing the ep11 config file

ifranzki:
When the ep11 config file contains an APQN_WHITELIST,
but the list is not terminated with the END keyword,
then pkcsconf -t terminates with a segmentation fault.
This fix checks for end-of-file at various states during
parsing and returns an appropriate error.


dodys:
There are multiple changes needed. 
1. You are using your fork's master branch to do development. Don't do it, create a specific branch at your fork for development. 
2. You are using tabs instead of spaces. 
Please review https://github.com/opencryptoki/opencryptoki/blob/master/CONTRIBUTING.md
You will need to open a new pull request from your new branch.

PR #63, opened 2017-08-14
Remove unneeded '#include "slotmgr.h"' in sample

ifranzki:
The '#include "slotmgr.h"' is not needed in the sample code, as well
as 'Slot_Mgr_Shr_t  *shmp = NULL;'. File slotmgr.h is not installed
into /usr/include/opencryptoki, so compiling this sample would not
be possible outside of the source tree.


dodys:
Nice catch, thanks!

PR #64, opened 2017-08-22
Support for Multi-Token-Instances

tuchscherer:
Opencryptoki supports one token type concurrently today.
Since the underlaying crypto hardware is able to provide multiple
crypto engines/domains, it's desirably to exploit that within
opencryptoki.
With this new extension opencryptoki enables multiple user to
participate from the underlaying hardware and isolate different users
from each other. Hence the resource allocation of domains can be
regulated in a more detailed granularity.
This feature allows the use of multiple instances of the same token
type. Each token can be configured with a specific configuration file
to define hardware resources and the token directory name. Thus token
objects are physically separated and secured by individual PINs.


pvital:
Since we already have a V2, would not be better close this one?

dodys:
Closing this Pull request as it was merged in #66 and #71

PR #66, opened 2017-09-20
Support for Multi-Token-Instances - v2

dodys:
I've rebased the patches of #64 (which I refer to as version 1) in a way it allow us to perform a git bisect and won't break building. The rebase resulted in 85 patches against the 86 of version 1 as one of the patches (patch 12) was integrated to other patch. I've kept the author and the original sign-off since all I've done was rebasing.
If the reviewers agree with the rebase, then before committing I will fix the number of each patch so they are linear, I didn't do this yet to make it easier to compare with version 1.

Follow description of version 1:
"Opencryptoki supports one token type concurrently today.
Since the underlaying crypto hardware is able to provide multiple
crypto engines/domains, it's desirably to exploit that within
opencryptoki.
With this new extension opencryptoki enables multiple user to
participate from the underlaying hardware and isolate different users
from each other. Hence the resource allocation of domains can be
regulated in a more detailed granularity.
This feature allows the use of multiple instances of the same token
type. Each token can be configured with a specific configuration file
to define hardware resources and the token directory name. Thus token
objects are physically separated and secured by individual PINs."


hfreude:
On 09/20/2017 02:47 PM, Eduardo Barretto wrote:
> @dodys requested your review on: opencryptoki/opencryptoki#66 Support for Multi-Token-Instances - v2.
>
I'm sorry. I didn't have time to review all theses patches.
However, I will try to spend some hours and have a look into these changes.
regards
Harald Freudenberger

dodys:
Just fixed the patches comments to be serial for merge.

PR #71, opened 2017-10-06
MTI additional patches

dodys:
The first patch ist just a update for the ep11 README document with an example configuration.
The second patch fix the access permission when creating new token directories.
The last 4 patches enable individual token locking instead having a single lock for all tokens. This is very useful when dealing with hundreds of tokens.

Developed by Ingo


dodys:
Reviewed-by: Eduardo Barretto <ebarretto@linux.vnet.ibm.com> 

I've reviewed it previously on pull request #64 but as that pull request was rebased, I've created a separate pull request for this additional features, making it more clear to understand.

PR #75, opened 2017-10-17
 Fix CKA_ALWAYS_AUTHENTICATE attribute for priv. keys

tuchscherer:
The PKCS#11 standard requires the CKA_ALWAYS_AUTHENTICATE attribute
to be present in private key objects. Actually this attribute is not
set as one of the default attributes. Further, it will be rejected
when it's declared within the application defined key template.

This fix will add the CKA_ALWAYS_AUTHENTICATE attribute by default
to any private key object. The value is set to 'false'. The calling
application is free to overrule by setting the attribute value
within the key template to 'true'.


dodys:
There are errors on the latest commit, and the other commits shouldn't be included in this pull request.

PR #88, opened 2017-11-02
Token private data field in STDLL_TokData_t for multi-instance-token support

ifranzki:
Introduce a field private_data in structure STDLL_TokData_t that allows
a token to anchor a pointer to a private control block to store all its
variables that are supposed to be global on a per token instance basis.

Instead of adding token specific fields for this purpose to the STDLL_TokData_t
structure (which is part of common code), the private_data field provides a
token implementation independent way of storing information on a per token
instance basis.

This patch also removes the EP11 specific field target_list from STDLL_TokData_t
into a EP11 specific private data block to keep STDLL_TokData_t free from any
token implementation specific fields.

### DIFF
--- a/usr/lib/pkcs11/common/host_defs.h
+++ b/usr/lib/pkcs11/common/host_defs.h
@@ -300,7 +300,7 @@ typedef struct _STDLL_TokData_t {
 	CK_STATE 	global_login_state;;
 	LW_SHM_TYPE	*global_shm;
 	TOKEN_DATA	*nv_token_data;
-	uint64_t	*target_list;		// pointer to adapter target list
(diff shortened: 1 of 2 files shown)


ifranzki:
For discussion: I guess we also should take a closer look at the other tokens in regards of its global variables in order to support the multi-instance-token support. For example, I am not sure if the ICA token needs changes in regards to global variable adapter_handle. Also the ICSF and TPM tokens seem to have some global variables that might need to be moved into a token private data structure.

dodys:
Adding everyone as a reviewer, but actually to discuss the points that @ifranzki brought up.

PR #91, opened 2017-11-07
Multi token instance - test tool / description

tuchscherer:
Updated opencryptoki.conf man page
Added utility to run ock test cases to multiple tokens in parallel.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -623,6 +623,7 @@ AC_CONFIG_FILES([Makefile usr/Makefile \
 	  testcases/pkcs11/Makefile \
 	  testcases/login/Makefile \
 	  testcases/misc_tests/Makefile \
+	  testcases/multi_token_tests/Makefile \
 	  man/Makefile \
 	  man/man1/Makefile \
 	  man/man1/pkcsconf.1 \
--- a/man/man5/opencryptoki.conf.5.in
+++ b/man/man5/opencryptoki.conf.5.in
@@ -55,6 +55,15 @@ minor version number (the hundredths portion of the version).
 If the slot is associated with a token that has its own configuration file,
 this option identifies the name of that configuration file.
 For example, confname=ep11tok.conf
+.TP
+.BR tokname
+If a token want to have its own token directory name that is different from the
+default name, especially if multiple tokens of the same type are configured,
+this option defines the name of the token individual directory.
+For example, tokname=ep11tok01
+
+Note: This key-value pair is optional: If only one token per token type is used,
+you don't need that entry. In that case the default directory name is used.
 
 .SH Notes
 The pound sign ('#') is used to indicate a comment.
--- a/testcases/Makefile.am
+++ b/testcases/Makefile.am
@@ -1,4 +1,4 @@
-SUBDIRS=common crypto pkcs11 login misc_tests
+SUBDIRS=common crypto pkcs11 login misc_tests multi_token_tests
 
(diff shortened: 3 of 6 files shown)


dodys:
Ingo, I see four problems in this testcase that I don't think that it is actually testing MTI:
1. From what I understood of the test, you pass the -s argument with a number, specifying the number of slots. But then when you call the test that you want to run, this number is actually used in the for loop, where in this loop you will use the iterator as the slot id for the test, like this: 
../crypto/aes_tests -slot 1 > aes_tests_rslt_slot_1.txt
So if you pass the -s 3, you will end up running:
../crypto/aes_tests -slot 1 > aes_tests_rslt_slot_1.txt
../crypto/aes_tests -slot 2 > aes_tests_rslt_slot_2.txt
../crypto/aes_tests -slot 3 > aes_tests_rslt_slot_3.txt

This means that you will execute aes for slot 1 which is ICA, slot 2 which is CCA and slot 3 which is Soft token.
So you are not testing the Multi token Instance. The idea of the MTI is that you can have 3 instances of the same token and run different tests on each instance at the same time right? 

2. I believe that to test MTI you would need to first change opencryptoki.conf, so the token instances are created. There will be a slot for each token instance. This is not being done.

3. The execution doesn't seem to be running in parallel, but sequentially, at least in the test I've done by changing the test to execute on only one token. I would rather prefer that the same test run on different instances at the same time. Or different tests are running in parallel in different instances. What do you think?

4. The second problem is that it doesn't check if the token is present or initiated. So if the token is not present, the test will simply remain as executing to the user, no error is being returned.

tuchscherer:
Hi Eduardo,
1. and 2.
Yes, the parameter 's' represent the last slot number. All slots (from 1 to 's') are subject to test.
Well, the opencryptoki.conf that is provided within the package is imho a configuration example. The user need to ensure that the slot/token configuration (opencryptoki.conf) will match to the system underneath or at least the required test scenario. Not every system (hardware platform) or every specific setup do support all kind of tokens. At the end the slot/token configuration is a very individual part of the setup. To keep things simple I just intended the sequential range of tokens under test. BTW: It doesn't matter what kind of tokens are configured. This multi token test utility should also work with different token types. We could of course improve the parameter list to hand over a list of token numbers, but this would also not prevent that users specify mixed type token numbers.

3. You're right, the parent process is waiting for the childs sequentially before forking a new one. Something went wrong during rebuild the test tool. I'll fix that and send you an updated version.
BTW: It was intended to run the same test to different token instances. Running different test cases to different token instances would be more complex.

4. Right, I see the problem. The child is not terminating. I'll fix that and send you an update.

dodys:
Hi Ingo, 

Although you made the fix for the child processes, I don't see a reason to have this testcase integrated since we already have ock_tests.sh (testcases/ock_tests.sh.in) which can be changed really quick to make the tests run in parallel (right now it is executed in sequence, for one token or for all available tokens) and also be tuned to execute just one testcase if the user wants to, instead of running all the crypto tests available.
I believe that is a better option than having similar testcases.

I am still not happy with not having an automated testcase for multiple instances of the same token, but as you mentioned this will depend on user's setup environment. So this will continue to be a manual test, as we will have to setup the other instances manually by changing the opencryptoki.conf regarding to our environment.

I will do some thinking on getting this automated in a generic way. 

So I will create another PR with the changes needed for ock_tests.sh.in, and include your man page commit on the same PR.

dodys:
Closing this pull request as this was merged on pull request #94

PR #92, opened 2017-11-09
Fix problem with EP11 token loading

ifranzki:
A tester reported a problem with the EP11 token when using the 3.8.1 release of OpenCryptoki. He got the following error when loading the EP11 token:

pkcsconf[55687]: apiutil.c DL_Load: dlopen() failed for [libpkcs11_ep11.so]; dlerror = [/usr/lib64/opencryptoki/stdll/libpkcs11_ep11.so: undefined symbol: dsa_sign]

This is caused by unresolved symbols in the EP11 token shared library.

### DIFF
--- a/usr/lib/pkcs11/common/mech_dsa.c
+++ b/usr/lib/pkcs11/common/mech_dsa.c
@@ -52,7 +52,7 @@ dsa_sign( STDLL_TokData_t     * tokdata,
       if (rc == CKR_OBJECT_HANDLE_INVALID)
           return CKR_KEY_HANDLE_INVALID;
       else
-        return rc
+        return rc;
    }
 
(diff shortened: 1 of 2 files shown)


usr/lib/pkcs11/common/mech_dsa.c
```diff
@@ -52,7 +52,7 @@ dsa_sign( STDLL_TokData_t     * tokdata,
       if (rc == CKR_OBJECT_HANDLE_INVALID)
           return CKR_KEY_HANDLE_INVALID;
       else
-        return rc
+        return rc;
```
dodys: Apparently this file was not being compiled or used by any token up to now, as the compilation was not breaking on it

PR #111, opened 2018-02-20
Fix token reinitialization

dodys:
During a token reinitialization, to set up a SO PIN or a User PIN again,
the operation would fail mentioning that the hash of the old PIN doesn't
match the entered old PIN. This happens because during InitToken
operation we are resetting the NV data, and with this we are loosing the
hash and md5 of the old PIN.
To fix this, instead of calling init_token_data on every InitToken, we
do a load_token_data, that will check if we have a NV data already
created. If yes, it just loads it. If not, it calls init_token_data.

Signed-off-by: Eduardo Barretto <ebarretto@linux.vnet.ibm.com>

### DIFF
--- a/usr/lib/pkcs11/icsf_stdll/icsf_specific.c
+++ b/usr/lib/pkcs11/icsf_stdll/icsf_specific.c
@@ -595,19 +595,9 @@ CK_RV reset_token_data(STDLL_TokData_t *tokdata, CK_SLOT_ID slot_id,
 
 	/* Reset token data and keep token name */
 	slot_data[slot_id]->initialized = 0;
-	init_token_data(tokdata, slot_id);
+    load_token_data(tokdata, slot_id);
 	init_slotInfo(&tokdata->slot_info);
 	tokdata->nv_token_data->token_info.flags |= CKF_TOKEN_INITIALIZED;
-
-	/* Reset SO pin to default and user pin to invalid */
-	pin_len = strlen((pin = "87654321"));
-	if (compute_sha1(tokdata, pin, pin_len,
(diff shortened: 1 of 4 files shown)


usr/lib/pkcs11/icsf_stdll/icsf_specific.c
```diff
@@ -595,20 +595,10 @@ CK_RV reset_token_data(STDLL_TokData_t *tokdata, CK_SLOT_ID slot_id,
 
 	/* Reset token data and keep token name */
 	slot_data[slot_id]->initialized = 0;
-	init_token_data(tokdata, slot_id);
+    load_token_data(tokdata, slot_id);
 	init_slotInfo(&tokdata->slot_info);
 	tokdata->nv_token_data->token_info.flags |= CKF_TOKEN_INITIALIZED;
 
-	/* Reset SO pin to default and user pin to invalid */
```
dvelarde: I don't understand this part of the patch.

dodys: ICSF token, specifically (the other tokens don't do that), was resetting the SO PIN to the default "87654321" and the User PIN to an invalid PIN. 
According to the PKCS#11 this shouldn't be done. 
If you are reinitializing a token, the SO PIN that you will use during this reinitialization must be the PIN that was set before the reinitialization. 
That's why I removed this piece of code as this is not compliant to the standard.

dvelarde: Thanks or the explanation.

PR #114, opened 2018-03-08
Fix Hardware Feature Object validation and tests

dodys:
Monotonic Counters have read-only attributes. If during CreateObject the
supplied template specifies a value for any of the read-only attributes,
then the attempt should fail with the error code CKR_ATTRIBUTE_READ_ONLY.
Fixed tests that created Monotonic counters objects.

### DIFF
--- a/testcases/misc_tests/obj_mgmt.c
+++ b/testcases/misc_tests/obj_mgmt.c
@@ -1162,251 +1162,260 @@ CK_RV do_CreateTokenObjects(void)
 }
 
 /*
- * do_HW_Feature_Search Test:
+ * do_HWFeatureSearch Test:
  *
- * 1. Create 5 objects, 3 of which are HW_FEATURE objects.
+ * 1. Create 4 objects, 2 of which are HW_FEATURE objects (1 of them is a
+ *    monotonic counter).
  * 2. Search for objects using a template that does not have its
  *    HW_FEATURE attribute set.
  * 3. Result should be that the other 2 objects are returned, and
  *    not the HW_FEATURE objects.
  * 4. Search for objects using a template that does have its
  *    HW_FEATURE attribute set.
- * 5. Result should be that the 3 hardware feature objects are returned.
+ * 5. Result should be that the only hardware feature objects that is not a
+ *    monotonic counter should be returned.
  *
  */
-
 CK_RV do_HWFeatureSearch(void)
 {
-	unsigned int            i;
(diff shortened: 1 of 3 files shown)


testcases/misc_tests/obj_mgmt.c
```diff
@@ -1162,251 +1162,260 @@ CK_RV do_CreateTokenObjects(void)
...
+     * since it is a read only feature.
+     */
+    rc = funcs->C_CreateObject(h_session, counter1_template, 6, &h_counter1);
+    if (rc != CKR_ATTRIBUTE_READ_ONLY) {
+        testcase_fail("C_CreateObject() rc = %s", p11_get_ckr(rc));
+        goto destroy_2;
+    }
+
+    rc = funcs->C_CreateObject(h_session, clock_template, 4, &h_clock);
+    if (rc != CKR_OK) {
```
dvelarde: I don't really understand the hard coded numbers (5, 6, 4) in the calls to C_CreateObject.  I do see other testcases that also use similar numbers.  But I am not understanding how you know which ones to use.

testcases/misc_tests/obj_mgmt.c
```diff
@@ -1162,251 +1162,260 @@ CK_RV do_CreateTokenObjects(void)
...
+     * in v2.11, it should return all objects *except* HW_FEATURE
+     * objects.
+     */
+    rc = funcs->C_FindObjectsInit(h_session, NULL, 0);
+    if (rc != CKR_OK) {
+        testcase_fail("C_FindObjectsInit() rc = %s", p11_get_ckr(rc));
+        goto destroy;
+    }
+
+    rc = funcs->C_FindObjects(h_session, obj_list, 10, &find_count);
```
dvelarde: Similar question here, why 10?

testcases/misc_tests/obj_mgmt.c
dodys: The third parameter that goes in a C_CreateObject call refers to the number of attributes in the template. For more information on this you can check on PKCS#11 standard.
In this case if you look at clock_template, it contains 4 attributes.

testcases/misc_tests/obj_mgmt.c
dodys: The third parameter in a C_FindObjects refers to the maximum number of objects handles to be returned, in this case being the size of the obj_list array.
For more information on this, please check page 96 of pkcs11-base-v2.40.pdf.

PR #123, opened 2018-06-04
Fix logic and warnings

dodys:
Fix logic for getting and releasing locks when DestroyObject is called and fix some compiler warnings.

### DIFF
--- a/usr/lib/pkcs11/common/loadsave.c
+++ b/usr/lib/pkcs11/common/loadsave.c
@@ -615,7 +615,7 @@ CK_RV save_private_token_object(STDLL_TokData_t *tokdata, OBJECT * obj)
 	CK_BYTE *clear = NULL;
 	CK_BYTE *cipher = NULL;
 	CK_BYTE *ptr = NULL;
-	CK_BYTE fname[100];
+	CK_BYTE fname[266];
 	CK_BYTE hash_sha[SHA1_HASH_SIZE];
 	CK_BYTE *key = NULL;
 	CK_ULONG key_len = 0L;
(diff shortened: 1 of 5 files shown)


ifranzki:
Looks good.

One note to commit "Fix logic on picking locks when deleting an object":
So when pthread_rwlock_wrlock() fails, you remember this and do not unlock afterwards. So far so good.
If another thread holds the lock already, pthread_rwlock_wrlock() will block until the lock is released by the other thread (OK, as expected). However, when the current thread already holds the lock, we assume that pthread_rwlock_wrlock() fails with a return code. This seems to be the case as our testing has shown, although the man page does not guarantee this:
See https://linux.die.net/man/3/pthread_rwlock_wrlock

> The calling thread may deadlock if at the time the call is made it holds the read-write lock (whether a read or write lock). 

Or http://pubs.opengroup.org/onlinepubs/7908799/xsh/pthread_rwlock_wrlock.html

> Results are undefined if the calling thread holds the read-write lock (whether a read or write lock) at the time the call is made.

dodys:
Hi Ingo, 
Yes, I've been through those docs before and for our case the pthread_rwlock_rdlock() or pthread_rwlock_wrlock() fails if the same thread already has a wrlock, returning the EDEADLK error code.
So even though the docs say it may or may not deadlock, I saw from the past week of testing that we always hit a deadlock in our scenario. So I think we are good to go with this implementation. 
I will continue to test and think of a better solution for this affter this release, as we are in a short window. It is in my TODO list.
So do you approve to merge this solution for now?

ifranzki:
I approve.
Thanks.

PR #127, opened 2018-07-06
Minor fixes in EC testcases.

jschmidb:


testcases/crypto/ec_func.c
```diff
@@ -381,7 +381,7 @@ CK_RV run_DeriveECDHKey()
 
                 for (m=0; m<NUM_SHARED_DATA; m++) {
 
-                    testcase_new_assertion();
+                    testcase_begin("Starting with ec=%lu, kdf=%lu, keylen=%lu, shared_data=%lu", i,j,k,m);
```
dodys: I believe this is not what you want, removing testcase_new_assertion() will decrease the number of testcases. You can add testcase_begin() but should keep the testcase_new_assertion.
Please check file testcases/regress.h and see the output of ec_tests to check what I am talking about.

jschmidb: You are right ... I added testcase_new_assertion again. I saw a second place in run_DeriveECDHKeyKAT where a  testcase_new_assertion was also missing.

dodys:
Please see #128

PR #129, opened 2018-07-18
Indentation

dodys:
This is a big patch set with major rewrites to fix the coding style.
Take your time to review it.
Thanks

PR #131, opened 2018-08-08
Identation 2

p-steuer:
Rebased PR https://github.com/opencryptoki/opencryptoki/pull/129 and integrated @ifranzki  s feedback so @dodys dont has to do more work ; -)


testcases/crypto/des3.h
```diff
@@ -862,50 +856,50 @@ static struct des3_test_vector des3_ofb64_tv[] = {
 #endif
```
ifranzki: According to the new coding style the "# define NUM_OF_PUBLISHED_TESTSUITES" should probably be without a blank.
I should have made it the correct was with my patch already, sorry....

p-steuer:
thanks ingo, fixed

p-steuer:
merged.

dodys:
Thanks @p-steuer!

PR #133, opened 2018-08-09
Fix CCA version migration tool

ifranzki:
This fix was triggered by a customer problem. They reported problems when migrating the CCA token objects from OCK version 2 to version 3. The CCA migration tool pkcscca ended successfully, but AES keys stored in token objects were not usable afterwards.

### DIFF
--- a/usr/lib/pkcs11/common/object.c
+++ b/usr/lib/pkcs11/common/object.c
@@ -308,6 +308,7 @@ CK_RV object_flatten(OBJECT * obj, CK_BYTE ** data, CK_ULONG * len)
     CK_ULONG tmpl_len, total_len;
     CK_ULONG offset;
     CK_ULONG_32 count;
+    CK_OBJECT_CLASS_32 class32;
     long rc;
 
     if (!obj) {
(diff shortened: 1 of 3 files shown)


usr/lib/pkcs11/common/object.c
```diff
@@ -329,7 +330,8 @@ CK_RV object_flatten(OBJECT * obj, CK_BYTE ** data, CK_ULONG * len)
 
     offset = 0;
 
-    memcpy(buf + offset, &obj->class, sizeof(CK_OBJECT_CLASS_32));
+    class32 = obj->class;
+    memcpy(buf + offset, &class32, sizeof(CK_OBJECT_CLASS_32));
```
p-steuer: sizeof(class32) resp. sizeof(*class32) if CK_OBJECT_CLASS_32 is a typedef'ed pointer (i dont hope so).

usr/lib/pkcs11/common/object.c
```diff
@@ -641,7 +644,8 @@ CK_RV object_restore_withSize(CK_BYTE * data, OBJECT ** new_obj,
 
     memset(obj, 0x0, sizeof(OBJECT));
 
-    memcpy(&obj->class, data + offset, sizeof(CK_OBJECT_CLASS_32));
+    memcpy(&class32, data + offset, sizeof(CK_OBJECT_CLASS_32));
```
p-steuer: same as above

PR #134, opened 2018-08-17
Updates for SEC1810 (Optimize single calls for EP11) and misc fixes

ifranzki:


testcases/include/regress.h
```diff
@@ -117,7 +119,8 @@ int get_user_pin(CK_BYTE_PTR);
 
 #define testcase_setup(total)   \
     t_total = 0;                \
-	t_errors = 0;
+    t_errors = 0;               \
+    gettimeofday(&timev3, NULL);
```
p-steuer: When you add a second statement,  please add do {...} while(0) for this function-like macro.

usr/lib/pkcs11/ep11_stdll/ep11tok.conf
```diff
@@ -34,6 +34,11 @@
 #
 #      CPFILTER /etc/opencryptoki/ep11cpfilter.conf
 #
+# To enable optimization of single part Sign/Verify and Encrypt/Decrypt 
+# operations specify the following option: 
+# 
+#      OPTIMIZE_SINGLE_PART_OPERATIONS
+#
```
p-steuer: Why is there a config option at all? Shouldnt onehot sign/verify enc/dec be the default i.e. done if its possible?
If it would be the default one could think of a config option to disable it, but then again is there even a reason one would like to do so ?

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
```diff
@@ -255,6 +255,7 @@ typedef struct {
     size_t control_points_len;
     int strict_mode;
     int vhsm_mode;
+    int optimize_single_ops;
```
p-steuer: i feel "one-shot" singn/verify resp. enc/dec is commonly used to decribe non-init-update-final apis. This applies to all namings of functions variable etc. However, calling it "single op" would still be okay for me since its internal names only.

p-steuer:
Done with my review so far.

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
```diff
@@ -4919,67 +4919,52 @@ CK_RV ep11tok_get_mechanism_info(STDLL_TokData_t * tokdata,
...
-            return CKR_MECHANISM_INVALID;
-    }
-
-    if (check_cps_for_mechanism(ep11_data->cp_config,
-                                type, ep11_data->control_points,
-                                ep11_data->control_points_len) != CKR_OK) {
-        TRACE_INFO("%s Mech '%s' banned due to control point\n",
-                   __func__, ep11_get_ckm(type));
-        return CKR_MECHANISM_INVALID;
-    }
```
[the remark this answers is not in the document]
ifranzki: I don't quite understand what you mean.  The checks below (i.e. the switch) is still there. The switch does not modify rc, so its fine to return with CKR_OK in any case if we get to the switch.

testcases/include/regress.h
ifranzki: Fixed.

usr/lib/pkcs11/ep11_stdll/ep11tok.conf
ifranzki: Defering the Init call to a later time may change the behaviour in regards of return codes. I.e. the init call may no longer return an error, but the error will then apear at the next call, when the defered init call is done. Since this may be seen as a behaviour that is not conform with the PKCS#11 standard, Reinhard wanted to enable this explicitely with an option.

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
ifranzki: The EP11 library functions uses the term "single", as in m_SignSingle().

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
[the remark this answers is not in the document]
p-steuer: okay

Issue #11, opened 2017-05-25
CCA library name changed


dodys:
The configure.ac states cca library to check for is csulsapi. But this is no longer true.
In csulcca-4.1.0-06, library name is csulcca.

This issue was migrated from: https://sourceforge.net/p/opencryptoki/bugs/147/

PR #135, opened 2018-08-20
some fixes 1

p-steuer:


usr/lib/pkcs11/api/api_interface.c
```diff
@@ -3004,6 +3004,12 @@ CK_RV C_OpenSession(CK_SLOT_ID slotID,
 
     TRACE_INFO("C_OpenSession  %lu %lx %p %p %p\n", slotID, flags,
                pApplication, Notify, phSession);
+
+    if (!(flags & CKF_SERIAL_SESSION)) {
+        TRACE_ERROR("%s\n", ock_err(ERR_SESSION_PARALLEL_NOT_SUPPORTED));
+        return CKR_SESSION_PARALLEL_NOT_SUPPORTED;
+    }
+
```
ifranzki: I hope that all our testcases set this flag?

usr/lib/pkcs11/common/object.c
```diff
@@ -144,28 +140,48 @@ CK_RV object_create(STDLL_TokData_t * tokdata,
...
         rc = build_attribute(CKA_NEVER_EXTRACTABLE, &flag, sizeof(CK_BYTE),
                              &extractable);
         if (rc != CKR_OK) {
             TRACE_DEVEL("build attribute failed.\n");
             goto error;
         }
+
         template_update_attribute(o->template, sensitive);
         template_update_attribute(o->template, extractable);
+        /* fall-through */
```
ifranzki: Hmm, so we require that CKA_LOCAL is already there, otherwise we fail.
If it is there, we ignore its value, and set it to FALSE in any case....
Sounds somewhat strange to me. Can't we just do template_update_attribute with the attribute set to FALSE in any case?

Function key_object_set_default_attributes (in key.c) already adds and sets CKA_LOCAL to FALSE. This is called from template_add_default_attributes for all key types inside object_create_skel which is called just above. So the code you added might not be needed at all.

usr/lib/pkcs11/api/apiutil.c
```diff
@@ -623,12 +623,17 @@ int DL_Load_and_Init(API_Slot_t *sltp, CK_SLOT_ID slotID)
 // representation
 void CK_Info_From_Internal(CK_INFO_PTR dest, CK_INFO_PTR_64 src)
 {
-    memset(dest, 0, sizeof(*dest));
-
     dest->cryptokiVersion = src->cryptokiVersion;
-    memset(dest->manufacturerID, '\0', 32);
-    memcpy(dest->manufacturerID, src->manufacturerID, 32);
+
+    memset(dest->manufacturerID, ' ', sizeof(dest->manufacturerID));
```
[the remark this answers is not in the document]
p-steuer: true

usr/lib/pkcs11/common/utility.c
```diff
@@ -495,8 +496,6 @@ void init_tokenInfo(TOKEN_DATA *nv_token_data)
     token_info->hardwareVersion.minor = 0;
     token_info->firmwareVersion.major = 1;
     token_info->firmwareVersion.minor = 0;
-
-    memset(token_info->utcTime, ' ', sizeof(token_info->utcTime));
```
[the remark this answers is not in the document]
p-steuer: it was moved to the top of the function, where all other fields are initialized to ' '

usr/lib/pkcs11/common/utility.c
```diff
@@ -449,14 +449,15 @@ void init_tokenInfo(TOKEN_DATA *nv_token_data)
...
     memset(token_info->manufacturerID, ' ', sizeof(token_info->manufacturerID));
     memset(token_info->model, ' ', sizeof(token_info->model));
     memset(token_info->serialNumber, ' ', sizeof(token_info->serialNumber));
+    memset(token_info->utcTime, ' ', sizeof(token_info->utcTime));
 
-    memcpy(token_info->label, label, strlen((char *) label));
-
-    memcpy(token_info->manufacturerID, manuf, strlen((char *) manuf));
-    memcpy(token_info->model, model, strlen((char *) model));
+    memcpy(token_info->label, label, strlen((char *)label));
```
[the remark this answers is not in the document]
p-steuer: since model manuf label are globals i assumed the length check to be done when they are set eg when reading config. but i can do it just to be sure.

usr/lib/pkcs11/api/api_interface.c
p-steuer: I ran the test suite with softtok and everything was working

usr/lib/pkcs11/common/object.c
p-steuer: Have to look into the CKA_LOCAL thing ie if its already set to false here. Some attribute set clearly wrong here leading to test case failure , see my mail.

Problem was because the comment above (that i removed) said:
> The PKCS #11 spec doesn't impose this restriction however.

... but it does, at least for private and secret keys.

usr/lib/pkcs11/common/utility.c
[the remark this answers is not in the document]
p-steuer: checks are done on input. since memcpy and strlen are used in combination in a lot of places id prefer keep itlike that for now

usr/lib/pkcs11/common/object.c
p-steuer: CKA_LOCAL is already present and set to false, as you expected. fixed that

p-steuer:
integrated your feedback

ifranzki:
thanks.

PR #136, opened 2018-08-20
EP11: Optimize digest calls using CPACF/libica (SEC1703)

ifranzki:


usr/lib/pkcs11/ep11_stdll/ep11_specific.c
```diff
@@ -243,6 +244,84 @@ static CK_RV check_cps_for_mechanism(cp_config_t * cp_config,
 static CK_RV get_control_points(STDLL_TokData_t * tokdata,
                                 unsigned char *cp, size_t * cp_len);
 
+/* Definitions for libica.
+ * Derived from ica_api.h so that there is no direct build dependency on libica
```
p-steuer: why not add the libica dependency. i mean libica functionality is used , even per default.

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
```diff
@@ -2792,22 +2804,140 @@ CK_RV ep11tok_generate_key(STDLL_TokData_t * tokdata, SESSION * session,
...
+        TRACE_ERROR("%s Libica SHA failed. mech=%s rc=0x%lx\n", __func__,
+                    ep11_get_ckm(mech), rc);
+
+        switch (rc) {
+        case EINVAL:
+            return CKR_ARGUMENTS_BAD;
+        case ENODEV:
+            return CKR_DEVICE_ERROR;
+        default:
+            return CKR_FUNCTION_FAILED;
```
p-steuer: i dont like all the ock internal functions returning the pkcs11 CKR_ retcodes: makes it harder to decide if you may change the return codes or if the return code behavior is mandated by pkcs11. Since this is dont core too much at this time.

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
p-steuer: The code would build also with older header files (you dont need the new sha api prototypes). The new symbols could not be resolved but thats okay. Im in favor of adding an ep11 tok dependency to libica-devel package e.g. ica_api header.

ifranzki: If I remember correctly, I tried to include ica_api.h first, but this lead to many compile errors, because ica_api.h caused many things to be redefined. E.g. ica_api.h includes <stdbool.h> which defines true/false. This conflicts with variables named true/false in OCK code. This of course could be changed in OCK code, but that would be a larger patch with lots of somehow unrelated changes.
Also unresolved symbols might be a problem, since OCK build/link fails when there are unresolved symbols.

ifranzki: And finally, the ica_api.h header only defines the function prototypes, but not the typedefs for the function pointers. So those would have to be 'copied' anyway....

p-steuer: Opened PR https://github.com/opencryptoki/opencryptoki/pull/137 to fix the 'true' issue.

The typedefs are needed anyway. Its only about the structure definitions (the problem would not exist if the struct defns were opaque).

How can the build fail with unresolved symbols when libica is dlopen'ed. I mean does it resolve symbols it does not use ?

usr/lib/pkcs11/common/mech_ec.c
```diff
@@ -60,6 +60,12 @@ CK_BYTE brainpoolP512r1[] = {
     0x06, 0x09, 0x2B, 0x24, 0x03, 0x03,
     0x02, 0x08, 0x01, 0x01, 0x0D
 };
+
+CK_BYTE brainpoolP512t1[] = {
```
p-steuer: trailing whitespace

ifranzki: fixed

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
ifranzki: OK, I have rebased all patches on top of the last to PRs, and have changed the code to include ica_api.h to get rid of the definitions contained in ica_api.h.

PR #142, opened 2018-08-22
Fixes 2

p-steuer:
Tested with soft- and ica-tok

### DIFF
--- a/usr/lib/pkcs11/ep11_stdll/ep11_specific.c
+++ b/usr/lib/pkcs11/ep11_stdll/ep11_specific.c
@@ -6160,9 +6160,17 @@ static int read_adapter_config_file(STDLL_TokData_t * tokdata,
         }
 
         if (strchr(ep11_data->cp_filter_config_filename, '/') == NULL) {
-            snprintf(cfgname, sizeof(cfgname) - 1,
-                     "%s/%s", cfg_dir, ep11_data->cp_filter_config_filename);
-            cfgname[sizeof(cfgname) - 1] = '\0';
+            cfgname[0] = '\0';
+
+            if (strlen(cfg_dir) + 1
+                + strlen(ep11_data->cp_filter_config_filename)
+                <= sizeof(cfgname) - 1) {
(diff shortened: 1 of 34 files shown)


p-steuer:
> Aren't these fixes also required for the token-specific new_host.c?

Possibly yes. I can check it for ep11. Question is what to do for icsf.

ifranzki:
> Question is what to do for icsf.
Depends on how much different the new_host.c files of the tokes are compared to the common one. If the changes you made are at place that where copied 1:1, then you should be able to include your fixes there as well.

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
```diff
@@ -6160,9 +6160,17 @@ static int read_adapter_config_file(STDLL_TokData_t * tokdata,
...
-                     "%s/%s", cfg_dir, ep11_data->cp_filter_config_filename);
-            cfgname[sizeof(cfgname) - 1] = '\0';
+            cfgname[0] = '\0';
+
+            if (strlen(cfg_dir) + 1
+                + strlen(ep11_data->cp_filter_config_filename)
+                <= sizeof(cfgname) - 1) {
+                strcpy(cfgname, cfg_dir);
+                cfgname[strlen(cfg_dir)] = '/';
+                strcpy(cfgname + strlen(cfg_dir) + 1,
```
ifranzki: You could use strcat instead here ...

p-steuer:
Ported the common new_host.c changes to icsf and ep11.

Im done so far with the clean up:

- Warning-free build under -Wall.
- No crashes with address- and undefined-behavior-sanitizer (1)
- Fixes #146 and  #8
- Fixes a lot of pkcs11 compliance issues
- Passes pkcs11test suite (1)

(1) Tested soft- and ica-tok. Other tokens still need to be tested.

ifranzki:
I ran all testcases with soft, ica, cca and ep11 tokens.
Looks good to me.

Issue #30, opened 2017-06-01
Examine NODH define


pvital:
The NODH define is included in each token's tok_struct.h but most do not have the function in
their token specific file nor in their mech_list. Needs cleaning up.

Migrated from https://sourceforge.net/p/opencryptoki/bugs/111/

ifranzki:
I checked the tokens, and it seems that the oly tokens that support DH are the soft token and the EP11 token. The EP11 token does not use common code/functions in tok_struct.h. The soft token implements the DF functions using OpenSSL. 

The CCA token has NODH in its makefile, but also implements the 2 token specific DH functions, but they simply return CKR_FUNCTION_NOT_SUPPORTED. Also due to NODA in the makefile the tokstruct uses NULL for the 2 functions. And the mechlist does not contain the DH mechanisms.

The ICA and TPM tokens have NODH in its makefile, but also implements the 2 token specific DH functions, but those functions do not even compile (when NODH is removed). Also the mechlists do not contain the DH mechanisms.

I would suggest to remove the incomplete implementation of the 2 token specific DH functions for CCA, ICA and TPM tokens, and specify NULL in the tokstruct unconditionally.  That way only the soft token keeps its working implementation, as well as the EP11 token. Would save us some dead code in the other tokens.

Opinions?

p-steuer:
Fixed by #151 . Closed.

PR #151, opened 2018-08-29
Cleanup incomplete DH implementation for CCA, ICA, TPM and ICSF tokens

ifranzki:
This fixes https://github.com/opencryptoki/opencryptoki/issues/30

The NODH define as such is kept so far, since it controls if DH specific common code parts are included. This is similar to the NODSA, NOAES defines.


p-steuer:
Did you check if ./configure CFLAGS="-DNODH -DNODSA -DNOAES" build is working ?

ifranzki:
Hm, no this fails... let me look into this.

p-steuer:
This is why i suggested to remove disabling algorithms at compile time completely. openssl already has such defines on which you technically depend (e.g. your openssl rsa code has an #ifndef OPENSSL_NO_RSA). Introducing a second set of such compiler time options intruduces a lot of complexity (e.g. #if !defined(OPENSSL_NO_RSA) && !defined(NORSA) ) ..

ifranzki:
I have fixed the compile errors with ./configure CFLAGS="-DNODH -DNODSA -DNOAES"
We can still decide to remove NOxxx defines completely.... Not sure what the original rational was to introduce these defines.

PR #154, opened 2018-08-30
EP11: Fix return code handling with single optimization

ifranzki:
Fix the return code handling for the single-operations optimizations. Failures of the deferred init call do not cause the function to return with an error immediately.

### DIFF
--- a/usr/lib/pkcs11/ep11_stdll/new_host.c
+++ b/usr/lib/pkcs11/ep11_stdll/new_host.c
@@ -1701,8 +1701,10 @@ CK_RV SC_EncryptUpdate(STDLL_TokData_t * tokdata, ST_SESSION_HANDLE * sSession,
     if (sess->encr_ctx.init_pending) {
         rc = ep11tok_encrypt_init(tokdata, sess, &sess->encr_ctx.mech,
                                   sess->encr_ctx.key);
-        if (rc != CKR_OK)
+        if (rc != CKR_OK) {
             TRACE_DEVEL("ep11tok_encr_init() failed.\n");
+            goto done;
+        }
 
         sess->encr_ctx.init_pending = 0;
     }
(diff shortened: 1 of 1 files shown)


usr/lib/pkcs11/ep11_stdll/new_host.c
```diff
@@ -1701,8 +1701,10 @@ CK_RV SC_EncryptUpdate(STDLL_TokData_t * tokdata, ST_SESSION_HANDLE * sSession,
     if (sess->encr_ctx.init_pending) {
         rc = ep11tok_encrypt_init(tokdata, sess, &sess->encr_ctx.mech,
                                   sess->encr_ctx.key);
-        if (rc != CKR_OK)
+        if (rc != CKR_OK) {
             TRACE_DEVEL("ep11tok_encr_init() failed.\n");
+            goto done;
```
p-steuer: I dont know what ep11tok_encrypt_init returns in case of deferred init but note that if its not CKR_BUFFER_TOO_SMALL, the session will be cleaned up at 'done'.

p-steuer: ah okay. cleaning up is probably what you want here, because there will be no final.

p-steuer:
> Failures of the deferred init call do not cause the function to return with an error immediately.

If this describes the faulty behaviour which this PR fixes, i approve. I first understood this described, what the PR does ; -).

usr/lib/pkcs11/ep11_stdll/new_host.c
ifranzki: yes that's intended.

ifranzki:
Yes this describes it. Without the 'goto done' it would fall through to the update call which then would fail with a different return code overwriting the original return code.
The 'gioto done' in the final calls are just to not skip printing the trace message at the end of the function.

PR #159, opened 2018-09-03
Changes for SW crypt/digest routines

ifranzki:


ifranzki:
Fixed.

usr/lib/pkcs11/common/sw_crypt.h
```diff
@@ -23,4 +23,17 @@ CK_RV sw_des3_cbc(CK_BYTE *in_data,
                   CK_ULONG *out_data_len,
                   CK_BYTE *init_v, CK_BYTE *key_value, CK_BYTE encrypt);
 
+#define sw_aes_cbc_encrypt(clear, len, cipher, len2, iv, key, keylen) \
+ sw_aes_cbc(clear, len, cipher, len2, iv, key, keylen, 1)
```
p-steuer: add parantheses for macro params

usr/lib/pkcs11/common/sw_crypt.h
```diff
@@ -23,4 +23,17 @@ CK_RV sw_des3_cbc(CK_BYTE *in_data,
...
+
+#define sw_aes_cbc_decrypt(clear, len, cipher, len2, iv, key, keylen) \
+ sw_aes_cbc(clear, len, cipher, len2, iv, key, keylen, 0)
```
p-steuer: add parantheses for macro params

usr/lib/pkcs11/common/sw_crypt.c
```diff
@@ -74,3 +75,35 @@ CK_RV sw_des3_cbc(CK_BYTE *in_data,
...
+    // the aes decrypt will only fail if the data length is not evenly divisible
+    // by AES_BLOCK_SIZE
+    if (in_data_len % AES_BLOCK_SIZE) {
+        TRACE_ERROR("%s\n", ock_err(ERR_DATA_LEN_RANGE));
+        return CKR_DATA_LEN_RANGE;
+    }
+
+    // Encrypt or decrypt the data
+    if (encrypt) {
+        AES_set_encrypt_key(key_value, keylen * 8, &aes_key);
```
p-steuer: Not using evp interfaces means no guarantee for full platform specific hardware support (old cbc mode apis have full hw support on z though). Dont know if encrypting/decrypting master key is relevant wrt performance.

usr/lib/pkcs11/common/utility.c
```diff
@@ -848,14 +848,18 @@ CK_RV compute_sha(STDLL_TokData_t *tokdata, CK_BYTE *data, CK_ULONG len,
         TRACE_DEBUG("failed to create digest.\n");
         return rv;
     }
-    return sha_hash(tokdata, NULL, FALSE, &ctx, data, len, hash, &hash_len);
+    rv = sha_hash(tokdata, NULL, FALSE, &ctx, data, len, hash, &hash_len);
+
+    if (ctx.context)
+        free(ctx.context);
```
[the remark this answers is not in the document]
ifranzki: Agreed, but usually the context is freed in digest_mgr_cleanup() which is inside dig_mgr.c.  Only in the 2 cases when it is called from utility.c, this is not done that way. I can certainly add a function to mech_sha.c to free the context, and call that one from utility.c, but that would change it only for those 2 cases.

usr/lib/pkcs11/common/sw_crypt.c
ifranzki: I don't think is performance relevant. this code is only used for master key encryption, not for real data encryption. Real data encryption uses the token specific crypto functions.

usr/lib/pkcs11/common/mech_sha.c
```diff
@@ -38,356 +38,23 @@
 #include "tok_spec_struct.h"
 #include "trace.h"
 
-#define SHA_HARDWARE_THRESHHOLD 128000
+#include <openssl/sha.h>
```
p-steuer: Same comment regarding evp as in the aes case applies here. However, for sha1 old interfaces have hw support on z.

usr/lib/pkcs11/common/sw_crypt.h
p-steuer: your right. it doesnt make a difference here.

usr/lib/pkcs11/common/mech_sha.c
ifranzki: Also not performance relevant. In general the token provided SHA routines are used. Only if the token does not provide a SHA routine, and its SHA1, then it falls back to the SW implementation. All our tokens do provide SHA routines.

compute_sha1 is used for pin verification, private token object handling and master key handling only. 

compute_sha is used inside mech_rsa for emsa_pss encode and verify, and oaep. Only if SHA1 is used and the token does not provide a SHA implementation, then the SW fallback is used.

usr/lib/pkcs11/common/utility.c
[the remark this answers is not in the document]
ifranzki: do you want me to change the 2 cases, or should I leave it as is?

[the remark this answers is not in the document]
p-steuer: what about completing the internal sw_sha1* by an update and final function ? So openssl's sha api is not called one time from sha_hash* layer and the other time from the lower sw_sha1* layer i.e.: sha_hash* > sw_sha1* > SHA1*. Context (de)allocation would be at the sw_sha1 layer:  _sha1 and _sha1_final would free the context and set the pointer to NULL, such digest_mgr_cleanup would not do a double free (cleanup's free would only be needed with unsuccesful sha ops).

usr/lib/pkcs11/common/sw_crypt.c
p-steuer: okay im fine with it.

usr/lib/pkcs11/common/mech_sha.c
p-steuer: alright

p-steuer:
Approved , thanks. While your at it could you move the sha_init upwards so it is placed near the other sha* functions and not hidden between the hmac* funcs ?

ifranzki:
moved sha_init upwards

PR #160, opened 2018-09-04
Fix memory leak in compute_sha()

ifranzki:
compute_sha() leaks memory when using a token specific sha routine. The token specific sha init allocates a context, but leaves it to digest_mgr_cleanup() to free it. 

Usually the token specific sha routines are called by the digest manager, so digest manager also cares about freeing the context. Function compute_sha() does not use the digest manager, so it needs to free the context itself.

In general the token specific sha functions should free its context at end of token_specific_sha and token_specific_sha_final, but this is not how it is done in OCK. Also the token specific sign/verify and encrypt/decrypt routines do not free their contexts, but leave it to the sign/verify manager and the encrypt/decrypt manager to do so. So at least its consistent...

### DIFF
--- a/usr/lib/pkcs11/common/utility.c
+++ b/usr/lib/pkcs11/common/utility.c
@@ -848,7 +848,10 @@ CK_RV compute_sha(STDLL_TokData_t *tokdata, CK_BYTE *data, CK_ULONG len,
         TRACE_DEBUG("failed to create digest.\n");
         return rv;
     }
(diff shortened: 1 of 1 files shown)


p-steuer:
> In general the token specific sha functions should free its context at end of token_specific_sha and token_specific_sha_final

Agreed. Either the size of the context is implementation specific, so the corresponding init/final functions should handle memory management, or it is not, then common code should handle memory management. Approved anyway, since it is consistent with ither code.

PR #161, opened 2018-09-04
Fixes

p-steuer:
memleak related stuff


usr/lib/pkcs11/common/template.c
```diff
@@ -536,14 +536,12 @@ CK_RV template_flatten(TEMPLATE *tmpl, CK_BYTE *dest)
                     ptr += attr->ulValueLen;
                 }
             }
+            free(attr_32);
```
ifranzki: This fix should also go into usr/sbin/pkcscca/pkcscca.c since function template_flatten() is copied there.

p-steuer: thanks for pointing this out. fixed.

PR #164, opened 2018-09-07
EP11 fixes for pkcs11test suite

ifranzki:


usr/lib/pkcs11/ep11_stdll/ep11_specific.c
```diff
@@ -5668,7 +5669,10 @@ CK_RV ep11tok_get_mechanism_list(STDLL_TokData_t * tokdata,
         *pulCount = 0;
         for (i = 0; i < counter; i++) {
             if (ep11tok_is_mechanism_supported(tokdata, mlist[i]) == CKR_OK) {
-                pMechanismList[*pulCount] = mlist[i];
+                if (*pulCount >= size)
+                    rc = CKR_BUFFER_TOO_SMALL;
+                else
+                    pMechanismList[*pulCount] = mlist[i];
```
p-steuer: Id prefer to check *pulCount >= size directly after incrementing *pulCount below. If this is the case , set rc and break out of the for loop.
While you are at it, can you remove the empty else-case below ?

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
```diff
@@ -5160,6 +5162,7 @@ static CK_RV ep11_ende_crypt_init(STDLL_TokData_t * tokdata, SESSION * session,
         ctx->context = ep11_state;
         ctx->context_len = ep11_state_l;
         if (rc != CKR_OK) {
+            encr_mgr_cleanup(ctx);
```
p-steuer: This change doesnt match the patch desciption. Accident ?

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
```diff
@@ -5178,6 +5181,7 @@ static CK_RV ep11_ende_crypt_init(STDLL_TokData_t * tokdata, SESSION * session,
...
```
p-steuer: THis change doesnt match the patch decription. Accident ?

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
```diff
@@ -1532,6 +1532,9 @@ static CK_RV rawkey_2_blob(STDLL_TokData_t * tokdata, SESSION * sess,
                              &cslen, (uint64_t) ep11_data->target_list);
     RETRY_END(rc, tokdata, sess)
 
+    if (rc == 0x8001000d)
```
p-steuer: Is there no ep11 header that has a readable define for this . If not, should we add such a define?

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
p-steuer: Ah okay, you need to check it before, since the function does not check for pulCount=0 above.
So in this case i suggest to add a break statement to the if-case above, so you dont need an else-case anymore.

ifranzki: *pulCount is the index to the array of mechs. After increasing it, you don't know if there will be a next mech (because all further might not be supported). If there is no next, then no need to set rc. 

Also if you return CKR_BUFFER_TOO_SMALL you must also return back the required number of elements in *pulCount. So we need to continue the loop to find out about the number of elements we really need.

I will remove the empty else.

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
ifranzki: No unfortunately not.... EP11 defines lost of private return codes. Just adding this one seems to be strange.

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
p-steuer: > So we need to continue the loop to find out about the number of elements we really need.

I was missing this. On first thought you dont see a reason to continue the for loop if its already clear, the function failed .. Can you do 
```
if (*pulCount < size)
    pMechanismList[*pulCount] = mlist[i];
```
and set the rc after the for loop. So this becomes more clear.

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
p-steuer: okay

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
ifranzki: Done.

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
p-steuer: in general i would prefer the length checks were done first and output buffer would be left unmodified in case its too small. however, since pkcs11 doesnt mandate this, im okay with it.

ifranzki: The output buffer isn't changed (there is no output buffer), just the internal context. We could certainly setup the context only if rc==OK, and just free the ep11_state otherwise.
Do you want me to change this?

p-steuer: the comment was meant as a closing remark regarding the ep11_get_mechanism_list discussion. Somehow put it in the wrong place.

ifranzki: Ah I see. The ep11_get_mechanism_list would have to run through the loop 2 times to achieve this. Once to get the number of elements required, and a second time to fill the list. And theoretically the list of mechs reported by EP11 could have changed in-between...

PR #173, opened 2018-09-11
EP11 fixes 2

ifranzki:


usr/lib/pkcs11/ep11_stdll/ep11_specific.c
```diff
@@ -4635,6 +4682,7 @@ CK_RV ep11tok_sign_init(STDLL_TokData_t * tokdata, SESSION * session,
     RETRY_END(rc, tokdata, session)
 
     if (rc != CKR_OK) {
+        rc = ep11_error_to_pkcs11_error(rc);
```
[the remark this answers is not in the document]
ifranzki: Function ep11_error_to_pkcs11_error writes a trace with the original return code.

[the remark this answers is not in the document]
p-steuer: somehow missed that. approved.

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
```diff
@@ -1438,6 +1438,55 @@ static int read_adapter_config_file(STDLL_TokData_t * tokdata,
...
+#define CKR_IBM_READONLY            CKR_VENDOR_DEFINED + 0x0001000c
+#define CKR_IBM_STATIC_POLICY       CKR_VENDOR_DEFINED + 0x0001000d
+#define CKR_IBM_TRANSPORT_LIMIT     CKR_VENDOR_DEFINED + 0x00010010
+
+static CK_RV ep11_error_to_pkcs11_error(CK_RV rc)
+{
+    if (rc < CKR_VENDOR_DEFINED)
+        return rc;
+
+    TRACE_ERROR("%s ep11 specific error: rc=0x%lx\n", __func__, rc);
```
hfreude: maybe consider printing a textual version of the ep11 specific error code.

testcases/crypto/digest_func.c
```diff
@@ -1184,13 +1184,23 @@ CK_RV do_SignVerify_HMAC(struct HMAC_TEST_SUITE_INFO * tsuite)
...
         rc = funcs->C_SignInit(session, &mech, h_key);
         if (rc != CKR_OK) {
+            if (rc == CKR_KEY_SIZE_RANGE) {
+                testcase_skip("keysize %d is not supported in slot %ld",
+                              (unsigned int) key_len, slot_id);
+                continue;
+            }
             testcase_error("C_SignInit rc=%s", p11_get_ckr(rc));
             goto error;
         }
```
hfreude: Ingo, have a look into testcases/common/common.c
There is a function check_supp_keysize() in there. This function is used by the digest testcases and checks based on the mechanis for valid key size range. I think, this is the better way. The mechanism should carry this information and so we can also check that the mechanism info entries are correct.

ifranzki: The problem is that the EP11 library does not seem to report correct min/max key size requirements via C_GetMechInfo. If it would report correct values, the check_supp_keysize() would catch it already. But since it does not, it fails afterwards. Thus the additional return code checking. 

I know that's not nice, but I have no idea how to fix this otherwise. It might be considered as a bug of the EP11 library, but this needs further discussion with the EP11 people.

dvelarde: Harald, are you ok with the changes now?

hfreude: Well, it hides an issue which may appear at the customer. We should make up bugzillas (or similar) to the ep11 guys. However, I would recommend to 1) add the check_supp_keysize() at all places where such a limitation may exist 2) have Ingo's workaround with skip test when the actual call fails with returncode CKR_KEY_SIZE_RANGE.

ifranzki: There is already a call to check_supp_keysize()  inside do_SignVerify_HMAC(). However, it does not catch this because the EP11 library reports min/max key sizes different from what it seems to accept later on. 

Also EP11 may complain about the key size during C_CreateObject called from create_GenericSecretKey() when the key size is too small. There is no way to check the supported min/max key size, since C_CreateObject does not use/reference a mechanism. Looks like as if Ep11 simply has an overall minimum key size that it allows. This might be related to a control point setting on the EP11 card.

So yes, we should report this to the EP11 guys, but we still need a interim fix until the EP11 library is fixed (if it will be fixed at all).

ifranzki:
Please wait with merging this PR. 

In the meantime I am questioning if Haralds commit 4cdd5038e732ab895fae64a32face784fee000f8 is correct. He assumes that ulMinKeySize returned by C_GetMEchanismInfo is always in BITS (instead of BYTES).  I don't think this is true for HMAC. 

Accroding to the PKCS#11 standard it is mechanism dependent if its in BITS or BYTES. All asymetric mechs use BITS, but symetric mechs like DES, 3DES, AES use BYTES. It does not tell how it is for HMAC (as far as I can tell), but I have found references in the internet that it is also BYTES for HMAC. 

If that's the case, the Haralds commit should be reverted, and the CCA token must be changed to report BYTES instead for the HAMC mechs. Also my 3rd commit in this PR would then be superflous.

ifranzki:
The EP11 library reports totally inconsistent min/max key sizes for HMAC mechanisms, min is in bytes, but max is in bits. 

I added a patch to temporarily correct this until the EP11 library is being fixed.

With this, the patch to check for CKR_KEY_SIZE_RANGE is obsolete. The check using check_supp_keysize() is sufficient once the key sizes are consistently in bits.

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
```diff
@@ -5816,14 +5816,31 @@ CK_RV ep11tok_get_mechanism_info(STDLL_TokData_t * tokdata,
...
+    case CKM_SHA512_HMAC_GENERAL:
+    case CKM_SHA512_224_HMAC:
+    case CKM_SHA512_224_HMAC_GENERAL:
+    case CKM_SHA512_256_HMAC:
+    case CKM_SHA512_256_HMAC_GENERAL:
+        /*
+         * EP11 may report ulMinKeySize in bytes (but ulMaxKeySize in bits).
+         * Adjust this if ulMinKeySize appears to be way too small.
+         */
+        if (pInfo->ulMinKeySize < 80)
```
p-steuer: At this time min is always in bytes, no? If this changes in future ep11 versions, we should check the version number instead of making an assumptions here (that everything < 80 is in bytes).

ifranzki: Agreed, but as far as I know there currently is no easy way to determine the EP11 host library version at runtime....

Currently min is always reported in BYTES, and EP11 does not support keys < 80 bits. So the min value in bytes can not be less then 10 bytes (or 80 bits).  So when EP11 reports min as BYTEs this check will turn it into BITS. Once EP11 reports min as BITS, the check will leave min as is (i.e. as BITS), because min then is >= 80.

We could even add another check for max the same way.  Currently max is reported as BITS, so the check will leave max as is (i.e. in BITS). If EP11 decides in the future to report min and max both as BYTES, then the check will turn max also to BITS, which is what we want to report externally.

ifranzki:
> As i understood they will fix the inconsistency by reporting both values in bytes.

See my comment avove.

ifranzki:
Even if this fix this, you never know with which EP11 host library version it will be run at the customer site. So we need to turn it into BOTS no matter what EP11 host library version is used.

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
p-steuer: Lets handle units in our code as they are now i.e. min is interpreted as bytes, max as bits. If units change with future ep11 version they must offer a way to detect it e.g. provide version information so we can just wrap our code with "if (old_ep11_version) .. else ..".

ifranzki:
Changed 3rd patch as just discussed (unconditionally adjust min key size) and added a 4th patch to revert change of check_supp_keysize() in testcases.

PR #178, opened 2018-09-19
EP11: Support CKM_ECDH1_DERIVE in PKCS#11 flavor 

ifranzki:
The EP11 library supports CKM_ECDH1_DERIVE slightly different than the PKCS#11 standard. It receives the public data directly as mechanism parameter. In PKCS#11, the mechanism parameter specifies struct CK_ECDH1_DERIVE_PARAMS which then contains the public data, a KDF and shared data.

Add code to accept CK_ECDH1_DERIVE_PARAMS as parameter, but pass the public data directly to EP11. KDFs other than CKD_NULL and shared data is still not supported and will return CKR_MECHANISM_PARAM_INVALID if specified.

Also update the EC testcase and remove special handling for EP11.

### DIFF
--- a/testcases/crypto/ec_func.c
+++ b/testcases/crypto/ec_func.c
@@ -464,30 +464,29 @@ CK_RV run_DeriveECDHKey()
                     // key and B's public key
 
                     if (is_ep11_token(SLOT_ID)) {
-                        /* ep11token implements PKCS#11 v2.20 */
-                        mech.mechanism  = CKM_ECDH1_DERIVE;
-                        mech.ulParameterLen = extr2_tmpl->ulValueLen;
-                        mech.pParameter = extr2_tmpl->pValue;
-                    } else {
-                        /* icatoken implements PKCS#11 v2.40, other tokens
-                         * don't support derive */
-                        ecdh_parmA.kdf = kdfs[j];
-                        ecdh_parmA.pPublicData = extr2_tmpl->pValue;
-                        ecdh_parmA.ulPublicDataLen = extr2_tmpl->ulValueLen;
-                        ecdh_parmA.pSharedData =
-                            shared_data[m].length == 0 ?
-                                NULL : (CK_BYTE_PTR) &shared_data[m].data;
-                        ecdh_parmA.ulSharedDataLen = shared_data[m].length;
-
-                        if (kdfs[j] == CKD_NULL) {
-                            ecdh_parmA.pSharedData = NULL;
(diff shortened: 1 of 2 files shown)


testcases/crypto/ec_func.c
```diff
@@ -464,31 +464,30 @@ CK_RV run_DeriveECDHKey()
                     // key and B's public key
 
                     if (is_ep11_token(SLOT_ID)) {
-                        /* ep11token implements PKCS#11 v2.20 */
```
ifranzki: Please note that the removed code mentions that the EP11 library would implement PKCS#11 v2.20. This is not true as far as I can tell. I checked the PKCS#11 v2.20 standard, and it already uses the CK_ECDH1_DERIVE_PARAMS as mechanism parameter. CKM_ECDH1_DERIVE was added with v2.11 and uses CK_ECDH1_DERIVE_PARAMS since then.

So obviously EP11 did never comply with PKCS#11 with CKM_ECDH1_DERIVE.

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
```diff
@@ -3504,9 +3504,49 @@ CK_RV ep11tok_derive_key(STDLL_TokData_t * tokdata, SESSION * session,
...
+     * EP11 supports CKM_ECDH1_DERIVE slightly different than specified in
+     * PKCS#11. It expects the public data directly as mechanism param,
+     * not via CK_ECDH1_DERIVE_PARAMS. It also does not support KDFs and shared
+     * data.
+     *
+     * ATTENTION: Once the EP11 library supports CKM_ECDH1_DERIVE in PKCS#11
+     * way, the following code needs to be removed. Possibly a EP11 library
+     * version check needs to be added to conditionally remove the following
+     * code.
+     */
```
p-steuer: id just mention the code has to be adapted in this case. if it has to be removed or be part of a if-version>-else block will be determined when its time. could you mark the begin and end of the code block to which this comment refers or just say "the following if block". When you cite pkcs11 spec could you add the version to which you refer (2.2.0).

p-steuer:
So this PR makes ep11toks CKM_ECDH1_DERIVE mech pkcs11 2.20 compliant. Approved modulo comment wording.

p-steuer:
I mean we claim 2.20 so the question is: is it 2.20 compliant. As far as i understood, it is.

ifranzki:
Yes it is.

PR #181, opened 2018-09-21
Fix -Wextra and -std=c99 -pedantic warnings

p-steuer:
... and add symbol versioning for libopencryptoki and the tokens.

Please test with ep11tok.


configure.ac
```diff
@@ -33,8 +54,12 @@ AC_PATH_PROG([GROUPADD], [groupadd], [/usr/sbin/groupadd])
 AC_PATH_PROG([CAT], [cat], [/bin/cat])
 AC_PATH_PROG([CHMOD], [chmod], [/bin/chmod])
 AC_PATH_PROG([CHGRP], [chgrp], [/bin/chgrp])
-AC_PROG_SED
-AC_PROG_MKDIR_P
+AC_PROG_CXX
```
hfreude: CXX - braucht ock den einen c++ compiler ?

usr/lib/pkcs11/opencryptoki_tok.map
```diff
@@ -0,0 +1,69 @@
...
+        SC_VerifyFinal;
+        SC_VerifyInit;
+        SC_VerifyRecover;
+        SC_VerifyRecoverInit;
+        SC_VerifyUpdate;
+        SC_WaitForSlotEvent;
+        SC_WrapKey;
+        ST_Initialize;
+    local: *;
+};
```
hfreude: Patrick, do I understand this right ? This is the interface between ock base and the tokens ?
So every token builds some glue code around the token struct which on the upper side has all these functions ?

usr/lib/pkcs11/ep11_stdll/ep11_func.h
```diff
@@ -197,9 +197,9 @@ typedef unsigned int (*m_admin_t) (unsigned char *response1, size_t * r1len,
                                    const unsigned char *cmd, size_t clen,
                                    const unsigned char *sigs, size_t slen,
                                    uint64_t target);
-typedef unsigned int (*m_add_backend_t) (const char *name, unsigned int port);
-typedef unsigned int (*m_init_t) (void);
-typedef unsigned int (*m_shutdown_t) (void);
+typedef int (*m_add_backend_t) (const char *name, unsigned int port);
+typedef int (*m_init_t) (void);
+typedef int (*m_shutdown_t) (void);
```
ifranzki: These typedefs are derived from the EP11 library API functions. They must match those from the EP11 library. Are you sure that the EP11 library uses int?

usr/lib/pkcs11/cca_stdll/cca_specific.c
```diff
@@ -272,99 +272,99 @@ static CK_RV cca_resolve_lib_sym(void *hdl)
...
-    dll_CSUAPCV = (CSUAPCV_t) dlsym(hdl, "CSUAPCV");
-    dll_CSUAPRB = (CSUAPRB_t) dlsym(hdl, "CSUAPRB");
-    dll_CSUADHK = (CSUADHK_t) dlsym(hdl, "CSUADHK");
-    dll_CSUADHQ = (CSUADHQ_t) dlsym(hdl, "CSUADHQ");
-    dll_CSNDTBC = (CSNDTBC_t) dlsym(hdl, "CSNDTBC");
-    dll_CSNDRKX = (CSNDRKX_t) dlsym(hdl, "CSNDRKX");
-    dll_CSNBKET = (CSNBKET_t) dlsym(hdl, "CSNBKET");
-    dll_CSNBHMG = (CSNBHMG_t) dlsym(hdl, "CSNBHMG");
-    dll_CSNBHMV = (CSNBHMV_t) dlsym(hdl, "CSNBHMV");
+    *(void **)(&dll_CSNBCKI) = dlsym(hdl, "CSNBCKI");
```
hfreude: What is the benefit of casting all these to void* instead of using the type casts ?

usr/lib/pkcs11/common/ock_syslog.h
```diff
@@ -0,0 +1,39 @@
...
+
+static inline void _ock_syslog(int priority, const char *file,
+                               const char *fmt, ...) {
+    char buf[OCK_SYSLOG_MAX];
+    size_t off;
+    va_list ap;
+
+    snprintf(buf, sizeof(buf), "%s ", file);
+    off = strlen("%s ");
+
```
hfreude: off = strlen("%s ") is always 3, but you want off = strlen(file) + 1;

usr/lib/pkcs11/common/trace.c
```diff
@@ -206,41 +206,69 @@ CK_RV trace_initialize(void)
...
+    pbuf = buf + len;
+    buflen = sizeof(buf) - len;
+    va_start(ap, fmt);
+    vsnprintf(pbuf, buflen, fmt, ap);
+    va_end(ap);
+
+    /* serialize appends to the file */
+    pthread_mutex_lock(&tlmtx);
+    if (write(trace.fd, buf, strlen(buf)) == -1)
+        fprintf(stderr, "cannot write to trace file\n");
```
hfreude: This is a bad implementation: write may return with > 0 but < strlen(buf).
However, much better would be to use the stdio functions fopen/fwrite/fclose instead of the bare system calls. For example there is no mutex needed then.

usr/lib/pkcs11/common/trace.h
```diff
@@ -121,42 +121,28 @@ extern struct trace_handle_t trace;
 void set_trace(struct trace_handle_t t);
 CK_RV trace_initialize();
 void trace_finalize();
-void ock_traceit(trace_level_t level, const char *fmt, ...)
-    __attribute__ ((format(printf, 2, 3)));
```
hfreude: Removing the __attribute__((format(printf, 2, 3))
disables all the format warnings - you will not do this.

usr/sbin/pkcsep11_migrate/pkcsep11_migrate.c
```diff
@@ -37,16 +37,25 @@ CK_LONG adapter = -1;
...
+                                        size_t, const unsigned char *,
+                                        size_t, uint64_t);
+typedef long (*ep11a_cmdblock_t) (unsigned char *, size_t, unsigned int,
+                                  const struct ep11_admresp *,
+                                  const unsigned char *,
+                                  const unsigned char *, size_t);
+typedef long (*ep11a_internal_rv_t) (const unsigned char *, size_t,
+                                     struct ep11_admresp *, CK_RV *);
+
+m_get_ep11_info_t _m_get_ep11_info;
```
hfreude: static ?

usr/lib/pkcs11/cca_stdll/cca_specific.c
```diff
@@ -2337,7 +2337,7 @@ CK_RV ccatok_hmac(STDLL_TokData_t * tokdata, SIGN_VERIFY_CONTEXT * ctx,
     }
 
     if (sign && !sig_len) {
-        TRACE_ERROR("%s received bad argument(s)\n", __FUNCTION__);
+        TRACE_ERROR("%s received bad argument(s)\n", __func__);
```
hfreude: isn't the function name anyway part of each trace statement ?

usr/lib/pkcs11/api/apiutil.c
```diff
@@ -515,7 +515,7 @@ int DL_Load(Slot_Info_t *sinfp, API_Slot_t *sltp, DLL_Load_t *dllload)
         char *e = dlerror();
         OCK_SYSLOG(LOG_WARNING,
                    "%s: dlopen() failed for [%s]; dlerror = [%s]\n",
-                   __FUNCTION__, sinfp->dll_location, e);
+                   __func__, sinfp->dll_location, e);
```
hfreude: isn't the function name anyway part of the syslog message ?

usr/lib/pkcs11/cca_stdll/cca_stdll.h
```diff
@@ -94,10 +94,10 @@ enum cca_key_type {
 
 #ifdef DEBUG
 #define CCADBG(fn, rc, reason) ock_logit("CCA_TOK DEBUG %s:%d  %s failed. " \
-        "return: %ld, reason: %ld\n", __FUNCTION__, __LINE__, fn, rc, reason)
+        "return: %ld, reason: %ld\n", __func__, __LINE__, fn, rc, reason)
 
 #define DBG(fmt, ...) ock_logit("CCA_TOK DEBUG %s:%d %s " fmt "\n", \
-    __FILE__, __LINE__, __FUNCTION__, ##__VA_ARGS__)
+    __FILE__, __LINE__, __func__, ##__VA_ARGS__)
 #else
```
hfreude: same here: the function name should already be part of the debug trace message

usr/lib/pkcs11/common/loadsave.c
```diff
@@ -616,7 +616,7 @@ CK_RV save_private_token_object(STDLL_TokData_t *tokdata, OBJECT *obj)
     CK_BYTE *clear = NULL;
     CK_BYTE *cipher = NULL;
     CK_BYTE *ptr = NULL;
-    CK_BYTE fname[266];
+    char fname[266];
```
hfreude: 266 -> PATH_MAX

usr/lib/pkcs11/ep11_stdll/ep11_func.h
p-steuer: The function prototypes in ep11.h have return type int. I assumed the header to be a copy of the original ep11 header file and thus decisive.

Moreover, m_init return value is checked for < 0 to indicate init failure, meaning m_init would always succeed if it returned an unsigned value.

usr/lib/pkcs11/opencryptoki_tok.map
p-steuer: yes

usr/lib/pkcs11/common/utility.c
```diff
@@ -282,8 +282,8 @@ CK_RV _UnlockMutex(MUTEX *mutex)
 
 CK_RV CreateXProcLock(char *tokname, STDLL_TokData_t *tokdata)
 {
-    CK_BYTE lockfile[2 * PATH_MAX + sizeof(LOCKDIR_PATH) + 6];
-    CK_BYTE lockdir[PATH_MAX + sizeof(LOCKDIR_PATH)];
+    char lockfile[2 * PATH_MAX + sizeof(LOCKDIR_PATH) + 6];
```
hfreude: 2 * PATH_MAX + sizeof ... + 6 ???
what about simple use PATH_MAX and change the code some lines below to use snprintf instead of sprintf

configure.ac
p-steuer: No. autotools complains about missing AC_PROC_CXX because some Makefile.am defines CXX flags. I l look into that, dont think its needed.

usr/lib/pkcs11/cca_stdll/cca_specific.c
p-steuer: Its a workaround to assign a function pointer to a pointer in a warnig-free way. See man 3 dlopen. I added a reference to that man page in the commit message.

usr/lib/pkcs11/icsf_stdll/icsf_specific.c
```diff
@@ -3415,7 +3417,7 @@ CK_RV icsftok_sign(SESSION * session, CK_BYTE * in_data, CK_ULONG in_data_len,
     struct session_state *session_state;
     SIGN_VERIFY_CONTEXT *ctx = &session->sign_ctx;
     struct icsf_object_mapping *mapping = NULL;
-    CK_BYTE chain_data[ICSF_CHAINING_DATA_LEN] = { 0, };
+    char chain_data[ICSF_CHAINING_DATA_LEN] = { 0, };
```
hfreude: why use char here instead of CK_BYTE ?

usr/lib/pkcs11/tpm_stdll/tpm_specific.c
```diff
@@ -3398,11 +3398,11 @@ CK_RV token_specific_get_mechanism_info(STDLL_TokData_t * tokdata,
 
 int token_specific_creatlock(void)
 {
-    CK_BYTE lockfile[PATH_MAX + (sizeof(LOCKDIR_PATH) - 1)
-                     + 2 * (sizeof(SUB_DIR) - 1)
-                     + (sizeof("///LCK..") - 1) + 1];
-    CK_BYTE lockdir[(sizeof(LOCKDIR_PATH) - 1) + (sizeof(SUB_DIR) - 1)
-                    + (sizeof("/") - 1) + 1];
+    char lockfile[PATH_MAX + (sizeof(LOCKDIR_PATH) - 1)
```
hfreude: PATH_MAX should be enough to store filenames. Maybe use snprintf some lines later in the code and check for truncation.

usr/lib/pkcs11/common/trace.c
p-steuer: has been working forever like that. will not be fixed with this warning-related PR.

usr/lib/pkcs11/cca_stdll/cca_specific.c
p-steuer: yes it is. i wont modify the trace output with this PR.

usr/lib/pkcs11/common/ock_syslog.h
p-steuer: fixed.

usr/lib/pkcs11/common/trace.h
p-steuer: Fixed (reintroduced the attribute).

usr/sbin/pkcsep11_migrate/pkcsep11_migrate.c
p-steuer: The tools under sbin/ have a lot of non-static global vars. Im not going to fix it whith this PR.

usr/lib/pkcs11/tpm_stdll/tpm_specific.c
p-steuer: LOCKDIR_PATH is provided by the build environment, technically you cant tell if it fits in PATH_MAX.

usr/lib/pkcs11/api/apiutil.c
p-steuer: see above.

usr/lib/pkcs11/cca_stdll/cca_stdll.h
p-steuer: see above.

p-steuer:
Thnaks for the feedback. I addressed everything.

PR #185, opened 2018-09-28
EP11: Call error translation routine after every EP11 call

ifranzki:


usr/lib/pkcs11/ep11_stdll/ep11_specific.c
```diff
@@ -1460,13 +1460,16 @@ static int read_cp_filter_config_file(const char *conf_name,
 #define CKR_IBM_STATIC_POLICY       CKR_VENDOR_DEFINED + 0x0001000d
 #define CKR_IBM_TRANSPORT_LIMIT     CKR_VENDOR_DEFINED + 0x00010010
 
-static CK_RV ep11_error_to_pkcs11_error(CK_RV rc)
+static CK_RV ep11_error_to_pkcs11_error(CK_RV rc, SESSION *session)
```
p-steuer: Are ep11 return codes always unsigned types such taht CK_RV aka unsigned long is appropriate? I recall library int function has int return value, negative values indiacting error.

ifranzki: This error translation routine is only meant for those dll_m_* functions that corelate to the PKCS#11 functions and return a CK_RV. It can't be used for such EP11 host library functions that return negative values. 

Unfortunately the typedefs in ep11_func.h are outdated or wrong. The header shipped with the EP11 library defines those functions with CK_RV, but the typedefs use int. 

While we are at it, do you want me to correct the typedefs according to the EP11 header of EP11 host library 1.3.x ?

p-steuer: yes that would be helpful. The typefdefs should match the function prototypes.

ifranzki: Just pushed a 2nd patch to correct the typedefs.

PR #187, opened 2018-10-01
Misc updates

ifranzki:


p-steuer:
> When ulSourceDataLen in CK_RSA_PKCS_OAEP_PARAMS is 0, then
also pSourceData must be NULL.

Quoted from pkcs11 ?

ifranzki:
> Quoted from pkcs11 ?

See 2.1.7 PKCS #1 RSA OAEP mechanism parameters, Table 6, PKCS #1 RSA OAEP: Encoding parameter sources, in description of CKZ_DATA_SPECIFIED:
> Array of CK_BYTE containing the value of the encoding parameter. If the parameter is empty, pSourceData must be NULL and ulSourceDataLen must be zero.

PR #188, opened 2018-10-01
EP11: add support for querying the EP11 lib and firmware version

ifranzki:
Sometimes the EP11 token must adjust the behavior of the EP11 host library when its behavior is not exactly as defined by the PKCS#11 standard. This adjustments are mostly EP11 host library or even firmware version dependent. Also some new features are only available/working with a certain host library or firmware version.

This patch adds code to query the versions of the host library, the firmware API level and the firmware version, so that this version information can be used later on.

### DIFF
--- a/usr/lib/pkcs11/ep11_stdll/ep11_func.h
+++ b/usr/lib/pkcs11/ep11_stdll/ep11_func.h
@@ -331,6 +331,99 @@ typedef long (*xcpa_queryblock_t) (unsigned char *blk, size_t blen,
 typedef long (*xcpa_internal_rv_t) (const unsigned char *rsp, size_t rlen,
                                     struct XCPadmresp * rspblk, CK_RV * rv);
 
+typedef CK_RV (*m_get_xcp_info_t)(CK_VOID_PTR pinfo, CK_ULONG_PTR infbytes,
+                                unsigned int query, unsigned int subquery,
+                                uint64_t target);
+
+#ifndef CK_IBM_XCP_HOSTQ_IDX
+
+typedef enum {
+    CK_IBM_XCPQ_API         =  0,  /* API and build identifier     */
+    CK_IBM_XCPQ_MODULE      =  1,  /* module-level information     */
+    CK_IBM_XCPQ_DOMAINS     =  2,  /* list active domains & WK IDs */
+    CK_IBM_XCPQ_DOMAIN      =  3,
+    CK_IBM_XCPQ_SELFTEST    =  4,  /* integrity & algorithm tests  */
+    CK_IBM_XCPQ_EXT_CAPS    =  5,  /* extended capabilities, count */
+    CK_IBM_XCPQ_EXT_CAPLIST =  6,  /* extended capabilities, list  */
+    CK_IBM_XCPQ_AUDITLOG    =  8,  /* audit record or records      */
+    CK_IBM_XCPQ_DESCRTEXT   =  9,  /* human-readable text/tokens   */
+    CK_IBM_XCPQ_EC_CURVES   = 10,  /* supported elliptic curves    */
+    CK_IBM_XCPQ_COMPAT      = 11,  /* domains' compatibility modes */
+    CK_IBM_XCPQ_MAX         = CK_IBM_XCPQ_COMPAT
+} CK_IBM_XCPQUERY_t;
+
+#define CK_IBM_XCP_HOSTQ_IDX  0xff000000  /* host-only queries index, min. */
+typedef enum {
+    CK_IBM_XCPHQ_COUNT    = 0xff000000, /* number of host-query indexes   */
+                                        /* including this type itself     */
+    CK_IBM_XCPHQ_VERSION  = 0xff000001, /* host-specific package version  */
+                                        /* such as packaging library ID   */
+    CK_IBM_XCPHQ_VERSION_HASH = 0xff000002,
+                                        /* assumed-unique identifier of   */
+                                        /* host code, such as version-    */
+                                        /* identifying cryptographic hash */
+                                        /* (library signature field...)   */
+    CK_IBM_XCPHQ_DIAGS    = 0xff000003, /* host code diagnostic level     */
+                                        /* 0 if non-diagnostics host code */
+    CK_IBM_XCPHQ_HVERSION = 0xff000004, /* human-readable host version    */
+                                        /* identification (recommended:   */
+                                        /* UTF-8 string)                  */
+    CK_IBM_XCPHQ_TGT_MODE = 0xff000005, /* host targeting modes           */
+                                        /* returns supported target modes */
+                                        /* as bitmask                     */
+                                        /* if not available only compat   */
+                                        /* target mode is in use          */
+                                        /* See CK_IBM_XCPHQ_TGT_MODES_t   */
+                                        /**/
+    CK__IBM_XCPHQ_MAX = CK_IBM_XCPHQ_TGT_MODE
+} CK_IBM_XCPHQUERY_t;
+
+typedef struct CK_IBM_XCPAPI_INFO {
+    CK_ULONG firmwareApi;
+    CK_ULONG firmwareConfig;          /* truncated firmware hash */
+} CK_IBM_XCPAPI_INFO;
+
+typedef CK_IBM_XCPAPI_INFO    CK_PTR   CK_IBM_XCPAPI_INFO_PTR;
+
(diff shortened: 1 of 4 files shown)


usr/lib/pkcs11/ep11_stdll/ep11_specific.c
```diff
@@ -7692,3 +7706,188 @@ CK_BBOOL ep11tok_optimize_single_ops(STDLL_TokData_t *tokdata)
 
     return ep11_data->optimize_single_ops ? CK_TRUE : CK_FALSE;
 }
+
+/* return -1 if v1 < v2, = iof v1 == v2, 1 if v1 > v2 */
```
p-steuer: did you mean "0 if v1 == v2" ?

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
```diff
@@ -1723,6 +1731,8 @@ static CK_RV ep11_resolve_lib_sym(void *hdl)
     *(void **)(&dll_xcpa_queryblock) = dlsym(hdl, "xcpa_queryblock");
     *(void **)(&dll_xcpa_internal_rv) = dlsym(hdl, "xcpa_internal_rv");
 
+    *(void **)(&dll_m_get_xcp_info) = dlsym(hdl, "m_get_xcp_info");
```
p-steuer: Is the m_get_xcp_info symbol exported by all ep11 lib versions? dll_m_get_xcp_info is dereferenced later w/o NULL checking.

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
ifranzki: Oups, yes of course. I have corrected it.

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
ifranzki: Good point. I added a NULL check. It should be available on all recent versions, but I don't know exactly when it was added.

p-steuer:
very helpful, thanks. looks good to me. @hfreude can you do a second review ?

usr/lib/pkcs11/ep11_stdll/ep11_func.h
```diff
@@ -331,6 +331,99 @@ typedef long (*xcpa_queryblock_t) (unsigned char *blk, size_t blen,
...
+} CK_IBM_XCPHQUERY_t;
+
+typedef struct CK_IBM_XCPAPI_INFO {
+    CK_ULONG firmwareApi;
+    CK_ULONG firmwareConfig;          /* truncated firmware hash */
+} CK_IBM_XCPAPI_INFO;
+
+typedef CK_IBM_XCPAPI_INFO    CK_PTR   CK_IBM_XCPAPI_INFO_PTR;
+
+typedef struct CK_IBM_XCP_INFO {
```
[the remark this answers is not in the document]
ifranzki: Yea, good idea ! See my second patch.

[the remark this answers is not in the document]
p-steuer: There are more fields that could be filled with real token data instead of the generic ock "dummy"data: flags, utcTime,  serialNumber, ? compare to CK_TOKEN_INFO definition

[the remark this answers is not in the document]
ifranzki: Regarding Flags, I don't know if this adds anything useful. The wire spec does not say much about the flags they supply.

utcTime is filled when C_GetTokenInfo is called with the current Linux time. The version query is only executed during token initialization, so supplying that UTC time would be wrong.

The serialnumber might be useful, however when multiple adapters are available, I assume that each adapter has its own serial number. So we could simply supply the serial number of the first adapter. Nevertheless, due to hot-plug, it might be that it will display a different serial number dependent on which adapters are online at the time of init.

What do you think?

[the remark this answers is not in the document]
p-steuer: The CK_TOKEN_INFO is returned by C_GetTokenInfo which should return up-to-date token info. We currently supply dummy values in most cases which can be cached. But if there is a clock on a HSM, its time should prabably be displayed in CK_TOKEN_INFO's utcTime such that it can be checked against the system time for example for tampering detection.

A token which doesnt have its own clock should prabably not set CKF_CLOCK_ON_TOKEN in CK_TOKEN_INFO's flags. Only for soft token it probably makes sense to supply system time in utcTime (and maybe ica, since its some kind of soft-token).

In the best case (we guessed the right token flags), the native ep11 flags dont add anything useful. But if we get the flag information directly from the token (as intendet) why not use it?

In case of the serial number im not sure how we model the token(=one ep11lib) - adapter(possible multiple) relation.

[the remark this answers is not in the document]
ifranzki: Regarding utcTime: OK, that would mean that we need another call to get_xcp_info to obtain the time value during C_GetTokenInfo. Here again the problem is that we can have multiple adapters. Theoretically every adapter could have a different time.... Again, I could just use the first one for simplicity.

Regarding flags: I checked and it only reports CKF_RNG, CKF_CLOCK_ON_TOKEN and CKF_TOKEN_INITIALIZED. We set those already in the code. So OR'ing the flags from the get_xcp_info call won't add anything useful. The token flags contain additional flags that the EP11 lib can not provide, such as PIN states, etc.

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
```diff
@@ -7692,3 +7706,194 @@ CK_BBOOL ep11tok_optimize_single_ops(STDLL_TokData_t *tokdata)
...
+    if (v1->minor < v2->minor)
+        return -1;
+    if (v1->minor > v2->minor)
+        return 1;
+    return 0;
+}
+
+static CK_RV get_card_type(uint_32 adapter, CK_ULONG *type)
+{
+    char fname[290];
```
hfreude: 290 -> PATH_MAX

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
```diff
@@ -7692,3 +7706,194 @@ CK_BBOOL ep11tok_optimize_single_ops(STDLL_TokData_t *tokdata)
...
+       OCK_SYSLOG(LOG_ERR,
+                  "Warning: Adapter %02X.%04X has a different firmware version"
+                  " than the previous adapters: %d.%d\n",
+                  adapter, domain,  xcp_info.firmwareVersion.major,
+                  xcp_info.firmwareVersion.minor);
+       qv->error = TRUE;
+       return CKR_OK;
+   }
+
+   if (!qv->first && qv->card_type != card_type) {
```
hfreude: Would't it make sense to move the card type check before the firmware version check? I assume that for different card types the firmware is anyway different.

ifranzki: Yes good point. Will move the card-type check as first check.

usr/lib/pkcs11/ep11_stdll/ep11_specific.c
ifranzki: Will be fixed. I will update the patch when you are through with your review.

ifranzki:
> But as there can be more than one HSM backends (APQNs on the same or different cards) it's not so easy. E.g. time or firmware level …

Even worse since we have hot-plug, so adapters can come and go during runtime. 
It is currently not handled when an adapter comes online AFTER token initialization with a different firmware version or card type, which can cause various problems.... Currently it is the responsibility of the admin to prevent such things.

We might want to think about a notification mechanism when adapters come online or go offline, e.g. by using udev. But this needs a good design.

p-steuer:
Slot events e.g.Token insertion/removal can be queried by C_WaitForSlotEvent. C_GetSlotInfo would return a CK_SLOT_INFO in that case with the CKF_TOKEN_PRESENT flag off.

ifranzki:
C_WaitForSlotEvent is implemented only very roughly in OCK, see api_interface.c
It also might not cover dynamic plugging of crypto adapters anyway, since the user of a token does not necessarily know (and need to know) that it is backed by multiple adapters.

I will open an issue with this topic (dynamic adapter plugging), we can then use this issue for further discussion.

p-steuer:
Okay, please copy my note regarding slot event handling.

This PR will be merged and closed now.

PR #194, opened 2018-10-09
CCA: Don't print debug messages to stdout

ifranzki:
Looks like a left over debug vehicle....
At lest it should go into the trace file and not to stdout.

If you think that this can be removed totally, then let me know.


p-steuer:
we dont print the rule array in other places as far as i can see. So i think this can be removed.

ifranzki:
That's what I thought as well. I just wanted to give Harald a chance to explain why it should be left in.
@hfreude Do you agree to remove?

ifranzki:
I changed the commit to remove the debug code.

Issue #186, opened 2018-09-28
Investigate return codes when a token does not support a specific EC curve


ifranzki:
Currently the tokens either return CKR_MECHANISM_PARAM_INVALID or CKR_ATTRIBUTE_VALUE_INVALID when they encounter a EC curve that they don't support. Mostly the same return code is returned regardless if C_CreateObject, C_GenerateKey, C_GenerateKeyPair, C_UnwrapKey, or C_DeriveKey is used. CKR_MECHANISM_PARAM_INVALID makes only sense for those cases where the mechanism parameter specifies the  EC curve, while  CKR_ATTRIBUTE_VALUE_INVALID makes sense for those cases where the EC curve is passed as attribute in a template.

PKCS#11 v2.40 added a new return code CKR_CURVE_NOT_SUPPORTED that fits much better.  According to v2.40 it can be returned by all above mention C_xxx functions.

The tokens also may need to be adapted to determine if a curve is supported or not.  There is a global array of EC curves supported by common code defined in mech_ec.c: `der_ec_supporte`d together with define `NUMEC`.

Currently the CCA token uses this array to determine if it supports a curve or not. However, the array of curves in common code may be (and was already) adjusted by adding additional curves, that CCA does not support. This will lead to various errors when curves that are supported by common code, but not supported by CCA are used with the CCA token. Either the CCA token needs its own array of supported curves, or the error handling must be enhanced to return the correct return code (i.e. CKR_CURVE_NOT_SUPPORTED) derived from the CCA return codes when a CCA verb is called with an unsupported curve.

The EP11 token uses the global curve array to determine if the common code supports that curve. So that usage is OK. Not sure if the EP11 library returns CKR_CURVE_NOT_SUPPORTED to indicate that a curve is not supported. 

The ICA token has its own list of supported curves already. Nevertheless it uses various return codes to indicate that a curve is not supported. This needs correction as well.

ifranzki:
This should be covered now.

PR #195, opened 2018-10-09
ICA/CCA: Return CKR_CURVE_NOT_SUPPORTED when EC curve is not supported

ifranzki:
This fixes issue https://github.com/opencryptoki/opencryptoki/issues/186 for ICA and CCA. 
For EP11 the EP11 host library is providing the return codes, so we can't change this in OCK.

### DIFF
--- a/usr/lib/cca_stdll/cca_specific.c
+++ b/usr/lib/cca_stdll/cca_specific.c
@@ -1556,6 +1556,54 @@ CK_RV build_update_attribute(TEMPLATE * tmpl,
     return CKR_OK;
 }
 
+CK_BBOOL is_curve_error(long return_code, long reason_code)
+{
+    if (return_code == 8) {
+        /*
+         * The following reason codes denote that the curve is not supported
+         *  8 874 (36A)    Error in Cert processing. Elliptic Curve is not
+         *                 supported.
+         *  8 2158 (86E)   There is a mismatch between ECC key tokens of curve
+         *                 types, key lengths, or both. Curve types and key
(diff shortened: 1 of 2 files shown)


usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -3697,27 +3697,27 @@ static CK_RV is_equal(unsigned char *a,
 static int nid_from_oid(CK_BYTE *oid, CK_ULONG oid_length)
 {
     /* Supported Elliptic Curves */
-    CK_BYTE brainpoolP160r1[] =
+    static CK_BYTE brainpoolP160r1[] =
```
p-steuer: static const ?

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -1744,6 +1778,8 @@ CK_RV token_specific_ec_generate_keypair(STDLL_TokData_t * tokdata,
     if (return_code != CCA_SUCCESS) {
         TRACE_ERROR("CSNDPKB (EC KEY TOKEN BUILD) failed. return:%ld,"
                     " reason:%ld\n", return_code, reason_code);
+        if (is_curve_error(return_code, reason_code))
```
p-steuer: Curve support is checked before the operation. How can the operation fail with curve not supported ?

p-steuer:
CKR_CURVE_NOT_SUPPORTED is new in PKCS 2.40. We should clarify which version is the target for the next release.

PR #196, opened 2018-10-09
Testcase fixes

ifranzki:


testcases/include/regress.h
```diff
@@ -296,11 +296,11 @@ int get_user_pin(CK_BYTE_PTR);
 
 #define testcase_user_logout()                                          \
     do {                                                                \
-        rc = funcs->C_Logout(session);                                  \
-        if (rc != CKR_OK) {                                             \
-            testcase_error("C_Logout() rc = %s", p11_get_ckr(rc));      \
-            if (rc != CKR_USER_NOT_LOGGED_IN)                           \
-                goto testcase_cleanup;                                  \
+        if (session != CK_INVALID_HANDLE) {                             \
+            rc = funcs->C_Logout(session);                              \
```
p-steuer: So this macro requires a session and rc variable to be defined. Any reason this is not a static inline function with a session parameter and a local rc variable ?

ifranzki: Don't ask me why... I did not invent this :-)
All the macros in regress.h seem to be that way.

PR #198, opened 2018-10-11
EP11: Fix memory leak when token initialization fails

ifranzki:


usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -232,7 +233,7 @@ typedef struct {
 #define CKF_EP11_HELPER_SESSION      0x80000000
 
 static CK_BOOL ep11_is_session_object(CK_ATTRIBUTE_PTR attrs, CK_ULONG attrs_len);
-static CK_RV ep11tok_relogin_session(STDLL_TokData_t * tokdata, SESSION * session);
+CK_RV ep11tok_relogin_session(STDLL_TokData_t * tokdata, SESSION * session);
```
p-steuer: why remove static if its local to ep11_specific.c ?

ifranzki: Well, that function is defined in header ep11_specific.h (which I needed to include to avoid another prototype). In the header it is defined without static. So I got a compile error.

But you are correct, it really is local to that file only, and would not need to be in the header file at all, and it could be static. 

So I fixed this the correct way: make it static, and remove the prototype in the header.

Issue #171, opened 2018-09-10
NUMBER_SLOTS_MANAGED vs MAX_SLOT_ID


p-steuer:
MAX_SLOT_ID and NUMBER_SLOTS_MANAGED are defined as 1024. However, slot numbers are checked for <= MAX_SLOT_ID in some places and <  NUMBER_SLOTS_MANAGED in others.

ifranzki:
MAX_SLOT_ID should be 1023. It is only used in the tokens (new_host.c) to validate the slot ID passed.  The rest of the common code (mainly api_interface.c) uses NUMBER_SLOTS_MANAGED and compares >= plus allocates arrays of 1024 elements (index 0 - 1023). Thus I would assume that the token never gets a slot ID larger than 1023 anyway, but lets correct it.

p-steuer:
Fixed by dba6d50c752d27d8e08792d37d2dc879528d50ad .

PR #200, opened 2018-10-16
Remove definition of MAX_SLOT_ID and use NUMBER_SLOTS_MANAGED instead

ifranzki:
This fixes issue https://github.com/opencryptoki/opencryptoki/issues/171

### DIFF
--- a/usr/lib/common/defs.h
+++ b/usr/lib/common/defs.h
@@ -21,8 +21,6 @@
 #define MAX_PIN_LEN           8
 #define MIN_PIN_LEN           4
 
-#define MAX_SLOT_ID           1024
-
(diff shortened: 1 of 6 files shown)


p-steuer:
Why do 2 defs exist at all ? Can MAX_SLOT_ID not be replaced by NUMBER_OF_SLOTS_MANAGED ?

ifranzki:
I guess the reason is that NUMBER_OF_SLOTS_MANAGED is defined in slotmgr.h, but this is not included in the files where MAX_SLOT_ID is used.

p-steuer:
If you need a def from slotmgr.h why not include it ?

PR #201, opened 2018-10-19
common: add CKA_LOCAL=TRUE attribute at rsa key generation

hfreude:
Local generated keys should have the CKA_LOCAL attribute
with a value of TRUE. However, for RSA keys this was not
always the case. Some token code adds this attribute, others
don't. Now the opencryptoki common code in mech_rsa.c
adds this attribute after token specific RSA key generation.

### DIFF
--- a/usr/lib/common/defs.h
+++ b/usr/lib/common/defs.h
@@ -21,8 +21,6 @@
 #define MAX_PIN_LEN           8
 #define MIN_PIN_LEN           4
 
-#define MAX_SLOT_ID           1024
-
 #ifndef MIN
 #define MIN(a, b)  ((a) < (b) ? (a) : (b))
 #endif
--- a/usr/lib/common/new_host.c
+++ b/usr/lib/common/new_host.c
@@ -43,6 +43,7 @@
 #include "tok_spec_struct.h"
 #include "pkcs32.h"
 #include "trace.h"
+#include "slotmgr.h"
 
 #include "../api/apiproto.h"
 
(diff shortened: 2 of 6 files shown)


ifranzki:
Hmm, with this change the LOCAL attribute is only set/added for RSA key pairs, but not for other key types such as EC, DSA, etc. Isn't it required for those key types similarly?
If so, then place your added code into function key_mgr_generate_key_pair in key_mgr.c instead. There is already code there to handle CKA_SENSITIVE nd CKA_EXTRACTABLE.

ifranzki:
Hmm, with this change the LOCAL attribute is only set/added for RSA key pairs, but not for other key types such as EC, DSA, etc. Isn't it required for those key types similarly?
If so, then place your added code into function key_mgr_generate_key_pair in key_mgr.c instead. There is already code there to handle CKA_SENSITIVE nd CKA_EXTRACTABLE.

ifranzki:
Hmm, with this change the LOCAL attribute is only set/added for RSA key pairs, but not for other key types such as EC, DSA, etc. Isn't it required for those key types similarly?
If so, then place your added code into function key_mgr_generate_key_pair in key_mgr.c instead. There is already code there to handle CKA_SENSITIVE nd CKA_EXTRACTABLE.

hfreude:
On 22.10.18 09:56, Ingo Franzki wrote:
> Hmm, with this change the LOCAL attribute is only set/added for RSA key pairs, but not for other key types such as EC, DSA, etc. Isn't it required for those key types similarly?
> If so, then place your added code into function key_mgr_generate_key_pair in key_mgr.c instead. There is already code there to handle CKA_SENSITIVE nd CKA_EXTRACTABLE.
>
>
>
You are right. This would make more sense... I'll rework and check ...

PR #204, opened 2018-10-24
Coverty fixes

ifranzki:
Lots of small and mostly trivial fixes for the issues found by the coverty scan.

Take you time to review....

### DIFF
--- a/testcases/pkcs11/sess_perf.c
+++ b/testcases/pkcs11/sess_perf.c
@@ -183,7 +183,7 @@ int do_SessionPerformance(unsigned int count)
         rc = create_des_encrypt_context(&(t[i].hsess), &(t[i].hkey));
         if (rc == FALSE) {
             show_error("create_aes_encrypt_context", (CK_RV) 0);
-            return FALSE;
+            goto ret;
         }
     }
 
@@ -192,15 +192,13 @@ int do_SessionPerformance(unsigned int count)
     rc = encrypt_DATA(t[0].hsess, t[0].hkey, 16);
     if (rc == FALSE) {
         show_error("encrypt_DATA #1", (CK_RV) 0);
-        return FALSE;
-
+        goto ret;
     }
 
     rc = encrypt_DATA(t[count - 1].hsess, t[count - 1].hkey, 16);
     if (rc == FALSE) {
         show_error("encrypt_DATA #2", (CK_RV) 0);
-        return FALSE;
-
+        goto ret;
     }
     GetSystemTime(&t2);
     process_time(t1, t2);
@@ -209,11 +207,15 @@ int do_SessionPerformance(unsigned int count)
         rc = finalize_des_encrypt_context(t[i].hsess);
         if (rc == FALSE) {
             show_error("finalize_aes_encrypt_context", (CK_RV) 0);
-            return FALSE;
+            goto ret;
         }
     }
 
-    return TRUE;
+    rc = TRUE;
+ret:
+    if (t != NULL)
+        free(t);
+    return rc;
 }
 
 int main(int argc, char **argv)
--- a/usr/lib/cca_stdll/cca_specific.c
+++ b/usr/lib/cca_stdll/cca_specific.c
@@ -2026,7 +2026,6 @@ CK_RV token_specific_sha(STDLL_TokData_t * tokdata, DIGEST_CONTEXT * ctx,
     if (return_code != CCA_SUCCESS) {
         TRACE_ERROR("CSNBOWH failed. return:%ld, reason:%ld\n",
                     return_code, reason_code);
-        free(cca_ctx->tail);
         return CKR_FUNCTION_FAILED;
     }
 
@@ -2273,7 +2272,6 @@ CK_RV token_specific_sha_final(STDLL_TokData_t * tokdata, DIGEST_CONTEXT * ctx,
     if (return_code != CCA_SUCCESS) {
         TRACE_ERROR("CSNBOWH (SHA FINAL) failed. return:%ld,"
                     " reason:%ld\n", return_code, reason_code);
-        free(cca_ctx->tail);
         return CKR_FUNCTION_FAILED;
     }
 
--- a/usr/lib/cca_stdll/cca_specific.c
+++ b/usr/lib/cca_stdll/cca_specific.c
@@ -2357,7 +2357,7 @@ CK_RV ccatok_hmac(STDLL_TokData_t * tokdata, SIGN_VERIFY_CONTEXT * ctx,
                   CK_BYTE * in_data, CK_ULONG in_data_len, CK_BYTE * signature,
                   CK_ULONG * sig_len, CK_BBOOL sign)
 {
-    struct cca_sha_ctx *cca_ctx = (struct cca_sha_ctx *) ctx->context;
+    struct cca_sha_ctx *cca_ctx;
     long keylen, return_code = 0, reason_code = 0, rule_array_count = 3;
     unsigned char rule_array[CCA_RULE_ARRAY_SIZE];
     OBJECT *key = NULL;
@@ -2368,6 +2368,7 @@ CK_RV ccatok_hmac(STDLL_TokData_t * tokdata, SIGN_VERIFY_CONTEXT * ctx,
         TRACE_ERROR("%s\n", ock_err(ERR_OPERATION_NOT_INITIALIZED));
         return CKR_OPERATION_NOT_INITIALIZED;
     }
+    cca_ctx = (struct cca_sha_ctx *) ctx->context;
 
     if (sign && !sig_len) {
         TRACE_ERROR("%s received bad argument(s)\n", __func__);
--- a/usr/lib/cca_stdll/cca_specific.c
+++ b/usr/lib/cca_stdll/cca_specific.c
@@ -1460,6 +1460,8 @@ CK_RV token_specific_aes_cbc(STDLL_TokData_t * tokdata,
                         " reason:%ld\n", return_code, reason_code);
         }
         (*out_data_len) = 0;
+        if (local_out != out_data)
+            free(local_out);
         return CKR_FUNCTION_FAILED;
     } else if (reason_code != 0) {
         if (encrypt) {
--- a/usr/lib/cca_stdll/cca_specific.c
+++ b/usr/lib/cca_stdll/cca_specific.c
@@ -1424,7 +1424,7 @@ CK_RV token_specific_aes_cbc(STDLL_TokData_t * tokdata,
                     &length,
                     in_data,
                     (long int *)out_data_len,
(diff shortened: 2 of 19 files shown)


usr/lib/common/mech_ssl3.c
```diff
@@ -920,12 +919,12 @@ static CK_RV ssl3_md5_only(STDLL_TokData_t *tokdata,
...
-    len = sizeof(hash);
-    rc = digest_mgr_digest_final(tokdata, sess, FALSE, &digest_ctx, hash, &len);
+    len = MD5_HASH_SIZE;
+    rc = digest_mgr_digest_final(tokdata, sess, FALSE, &digest_ctx, outBuff,
+                                 &len);
 
-    if (rc == CKR_OK) {
+    if (rc != CKR_OK) {
         TRACE_DEVEL("Digest Final failed.\n");
-        memcpy(outBuff, hash, len);
```
hfreude: Ingo, please check the memcpy line you removed here. For the good case (rc == CKR_OK) it should be called.

usr/sbin/pkcscca/pkcscca.c
```diff
@@ -1537,11 +1538,15 @@ int find_wrapped_keys(CK_FUNCTION_LIST *funcs, CK_SESSION_HANDLE sess,
...
-        handles = realloc(handles, sizeof(CK_OBJECT_HANDLE) * (++ulTotalCount));
-        if (!handles) {
-            print_error("Malloc of %lu bytes failed!", ulTotalCount);
-            break;
+        ptr = realloc(handles, sizeof(CK_OBJECT_HANDLE) * (++ulTotalCount));
+        if (!ptr) {
+            print_error("Malloc of %lu bytes failed!",
+                        sizeof(CK_OBJECT_HANDLE) * ulTotalCount);
+            funcs->C_FindObjectsFinal(sess);
+            free(handles);
```
hfreude: hm, if realloc() fails, the original block is left untouched; it is not freed or moved.
Why do you then free handle ? the old code did the right thing here.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -2368,6 +2368,7 @@ CK_RV ccatok_hmac(STDLL_TokData_t * tokdata, SIGN_VERIFY_CONTEXT * ctx,
         TRACE_ERROR("%s\n", ock_err(ERR_OPERATION_NOT_INITIALIZED));
         return CKR_OPERATION_NOT_INITIALIZED;
     }
+    cca_ctx = (struct cca_sha_ctx *) ctx->context;
```
hfreude: this change does nothing, even more it splits one clear line of code into two.
What do you want to fix here ?

usr/lib/common/mech_ssl3.c
ifranzki: The old code did first receive the hash into variable hash, and then copy it as is to outBuff. The new code receives it directly into outBuff, so no memcpy needed anymore, regardless of good or bad case.

usr/sbin/pkcscca/pkcscca.c
ifranzki: If realloc fails it returns NULL and thus sets handles to NULL. The original block is not freed, but you don't have the address anymore, because handles was set to NULL. So the original block is leaked. 
The new code frees the original block when realloc fails, and cleas up and returns. The original code continued with handles being NULL and thus causes a null pointer dereference as well in later code.

usr/lib/cca_stdll/cca_specific.c
ifranzki: The problem was that ctx->context was done in local variable declaration, before ctx is checked for being not NULL.  So in case ctx is NULL, then you have a null pointer dereference.  The new code avoids this by moving the dereferencing after the NULL checking.

usr/lib/common/mech_ssl3.c
hfreude: ok - fine with me

usr/sbin/pkcscca/pkcscca.c
hfreude: Would it be possible, to leave the handle memory untouched if realloc() fails and just
return with some failed. I agree, the old code didn't make it correct, but the fix now
could do better - and not free the handle memory but just return with failed.

usr/lib/cca_stdll/cca_specific.c
hfreude: hm, let's assume ctx = NULL then the line
struct cca_sha_ctx *cca_ctx = (struct cca_sha_ctx *) ctx->context;
does introduce a new variable cca_ctx which is a pointer to ctx + some offset.
So if ctx is NULL, cca_ctx points to NULL + some offset.
As soon as you access something via this pointer you have a problem, but the pointer declaration and assignment is not a dereferencing by itself. In my opinion the old code here was and is correct and clearer to understand. The only point Coverty has in my mind is that when you try to check our derived pointer e.g.
if (!cca_ctx)  or  if (cca_ctx != NULL) or so, you will not get the right result.

usr/sbin/pkcscca/pkcscca.c
ifranzki: If you leave handles untouched when realloc fails and you break the loop (as it was before), then you would process only a subset of all the objects found. The user would not notice that you processed only a subset of the objects. Also ulTotalCount would have been incremented already, so it would actually not match the elements in handles, which causes another problem....

In my opinion: If realloc fails because we run out of memory (very unlikely anyway), it is very likely that any subsequent processing will also fail sooner or later, because it will  also allocate memory. So lets stop processing right now with a proper error message, and do proper cleanup.

ifranzki:
The proposed fix in mech_ec.c from issue 205 is already included in my commit a759cfbeca38bc1c2ee4f80188b30a405b5b5782 .
The proposed fix in mech_ssl3.c from issue 205 is already included in my commit 86a1cda5842c08b2aff045dfc090221403a3bdbe .

usr/lib/common/utility.c
```diff
@@ -581,9 +581,10 @@ CK_RV compute_next_token_obj_name(CK_BYTE *current, CK_BYTE *next)
     for (i = 0; i < 8; i++) {
         if (current[i] >= '0' && current[i] <= '9')
             val[i] = current[i] - '0';
-
-        if (current[i] >= 'A' && current[i] <= 'Z')
+        else if (current[i] >= 'A' && current[i] <= 'Z')
             val[i] = current[i] - 'A' + 10;
+        else
+            val[i] = 36;
```
p-steuer: The function should fail here with e.g. EINVAL because a byte in current is not in range [0-9A-Z].

ifranzki: Yes I agree, but not all callers check the return code of this function (e.g.  object_mgr_copy() and object_mgr_create_final() so I thought it would be better to do something that produces a valid name. With value 36 the code following this will set the invalid character to '0' and increment the following characters by one to make in unique. 

Let me know if you still prefer to return an error here, with the risk that the calling function will ignore the return code and work with a non-unique or invalid object name.

testcases/pkcs11/sess_perf.c
```diff
@@ -209,10 +212,12 @@ int do_SessionPerformance(unsigned int count)
         rc = finalize_des_encrypt_context(t[i].hsess);
         if (rc == FALSE) {
             show_error("finalize_aes_encrypt_context", (CK_RV) 0);
+            free(t);
             return FALSE;
         }
     }
 
+    free(t);
```
p-steuer: Please use the following function exit:
```
    rv = TRUE;
ret:
    if (t != NULL)
        free(t);
    return rv;
```

Initialize 'rv = FALSE' and replace all
```
free(t);
return FALSE;
```
by
```
goto ret;
```

usr/lib/common/mech_ec.c
```diff
@@ -852,6 +852,7 @@ CK_RV ckm_kdf_X9_63(STDLL_TokData_t *tokdata, SESSION *sess, CK_ULONG kdf,
         counter++;
     }
 
+    free(ctx);
```
p-steuer: also free ctx in above
```
if (rc != 0) {
    return rc;
}
```

usr/sbin/pkcscca/pkcscca.c
```diff
@@ -1537,11 +1538,15 @@ int find_wrapped_keys(CK_FUNCTION_LIST *funcs, CK_SESSION_HANDLE sess,
         if (ulObjectCount == 0)
             break;
 
-        handles = realloc(handles, sizeof(CK_OBJECT_HANDLE) * (++ulTotalCount));
```
p-steuer: The above free(handles) needs a NULL check.

testcases/pkcs11/sess_perf.c
ifranzki: OK, I am now reusing variable rc and use the ret goto lable.

usr/lib/common/mech_ec.c
ifranzki: good catch. Fixed.

usr/sbin/pkcscca/pkcscca.c
ifranzki: Good catch, fixed.

usr/lib/common/utility.c
p-steuer: Okay so leave it like that but extend the above comment stating that function will return the next obj name even if given name is invalid.

testcases/pkcs11/sess_perf.c
p-steuer: okay for me if it works but usually its better to have two local vars, one for caller return code (pre-set to eg false) and one to catch and check the callee return codes.

usr/lib/common/utility.c
ifranzki: done.

testcases/pkcs11/sess_perf.c
```diff
@@ -183,7 +183,7 @@ int do_SessionPerformance(unsigned int count)
         rc = create_des_encrypt_context(&(t[i].hsess), &(t[i].hkey));
         if (rc == FALSE) {
             show_error("create_aes_encrypt_context", (CK_RV) 0);
-            return FALSE;
+            goto ret;
```
p-steuer: problem is function returned boolean before and now rcs of callees. As noted in my other comment i think it would make sense to have a seperate return code var for  the function return, initialized to false.

ifranzki: Well, in all cases there is a check for 

> if (rc == FALSE)

 just before the goto, so rc is FALSE in all those cases.

p-steuer: somehow missed that

p-steuer:
@hfreude : all your comments seem to be addressed. Please approve if you are happy with it now.

PR #207, opened 2018-10-29
EP11: Firmware version checking for HMAC min/max key sizes

ifranzki:
Add firmware version dependent behavior for HMAC mechanisms. When a certain fix is available the min/max key sizes must be adjusted to report bits in C_GetMechanismInfo. When the fix is not available, only the min key size must be adjusted. For mixed version configurations, exclude HMAC mechanisms
from the list of supported mechanisms (C_GetMechanismList), to avoid inconsistent behavior.


ifranzki:
made both functions static.

p-steuer:
lgtm, please also check @hfreude .

ifranzki:
LOL yes... But I will reuse all the version handling code for other place as well in the future.

PR #208, opened 2018-10-29
Fix -Wpedantic and a typo

p-steuer:
Enums are int-valued (c99).

Signed-off-by: Patrick Steuer <patrick.steuer@de.ibm.com>


p-steuer:
well, this is an advantage of shipping the header: being able to address its problems. conforming programs including the original header will see the warnings because the value is implicitely negative.

PR #210, opened 2018-10-31
Ensure the XProcLock is held when token object files are saved, read, or deleted.

ifranzki:
This fixes a problem with message `usr/lib/common/loadsave.c Cannot read size.` on syslog when running multi threaded.

Note: This patch is on top of the "Check XProcLock return values" patch.
Since that other patch is not yet merged, it is also included in this PR. 
I may need to rebase my patch once the other one is merged.

### DIFF
--- a/usr/lib/common/loadsave.c
+++ b/usr/lib/common/loadsave.c
@@ -532,6 +532,7 @@ out_nolock:
 }
 
 //
+// Note: The token lock (XProcLock) must be held when calling this function.
 //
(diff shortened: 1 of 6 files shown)


p-steuer:
Looks good to me. Maybe it would be a good idea to add a comment to functions like load_public_token_objects which require the caller to hold the lock.

ifranzki:
Added a note to the functions like load_public_token_objects in loadsave.c that they require the caller to hold the lock.
Rebased to the current master branch.

Issue #166, opened 2018-09-07
btrees memleak on C_Finalize


p-steuer:
opencryptoki stores state information in five btrees: sess_btree, sess_obj_btree, priv_token_obj_btree, publ_token_obj_btree and object_map_btree.

When a stdll is unloaded by C_Finalize, only  priv_token_obj_btree and publ_token_obj_btree seem to be deallocated and pointers to the memory allocated for the remaining btrees are lost. To be investigated.

ifranzki:
sess_obj_btree is freed when the session terminates, which should happen during or before C_Finalize. C_Finalize calls session_mgr_close_all_sessions() which then closes all left over sessions and also frees its session objects from sess_obj_btree.

This should also cleanup  sess_btree, since when all sessions are closed, the sess_btree is empty, thus no need to further cleanup sess_btree.

Similar with object_map_btree. When all objects are freed (token objs via object_mgr_purge_token_objects(), as well as session objs via session_mgr_close_all_sessions()) then this tree is empty, thus no need to further cleanup.

p-steuer:
As far as i investigated the problem it is supposed to work exactly as you said. However, valgrind still reports memory loss in case of said three btrees.

p-steuer:
Problem with the freeing procedures in session_mgr_close_all_sessions, object_mgr_purge_token_objects etc was that nodes on the free list(*) were ignored.

(*) Free listed nodes are currently not used but were not deallocated such that the can be reused later to save alloc/free calls. Question is if the number of such nodes should be restricted to prevent an excessive amount of them.

PR #214, opened 2018-11-06
Fixes

p-steuer:
First patch:
Fix conditional jump or move depending on uninitialised value(s)

Valgrind complaints about some uninitialized local variables.

Second patch:
Fix C_Finalize memleak of binary trees' nodes with BT_FLAG_FREE set.

C_Finalize uses bt_for_each_node to free the btrees' nodes which
in turn calls bt_get_node for each node in the tree. However,
bt_get_node returns NULL if a node has the BT_FLAG_FREE set
such that bt_for_each_node ignores those nodes, so they are
never actually freed.
Call bt_destroy in C_Finalize to also cleanup those nodes.

Fixes https://github.com/opencryptoki/opencryptoki/issues/166 .

### DIFF
--- a/usr/lib/icsf_stdll/new_host.c
+++ b/usr/lib/icsf_stdll/new_host.c
@@ -269,6 +269,14 @@ CK_RV SC_Finalize(STDLL_TokData_t * tokdata, CK_SLOT_ID sid, SLOT_INFO * sinfp)
 
     session_mgr_close_all_sessions();
     object_mgr_purge_token_objects(tokdata);
+
+    /* Finally free the nodes on free list. */
+    bt_destroy(&sess_btree, NULL);
+    bt_destroy(&sess_obj_btree, NULL);
+    bt_destroy(&object_map_btree, NULL);
+    bt_destroy(&priv_token_obj_btree, NULL);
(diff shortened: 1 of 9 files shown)


usr/lib/icsf_stdll/new_host.c
```diff
@@ -269,7 +269,15 @@ CK_RV SC_Finalize(STDLL_TokData_t * tokdata, CK_SLOT_ID sid, SLOT_INFO * sinfp)
...
     object_mgr_purge_token_objects(tokdata);
+
+    /* Finally free the nodes on free list. */
+    bt_destroy(&sess_btree, NULL);
+    bt_destroy(&sess_obj_btree, NULL);
+    bt_destroy(&object_map_btree, NULL);
+    bt_destroy(&priv_token_obj_btree, NULL);
+    bt_destroy(&publ_token_obj_btree, NULL);
     detach_shm(tokdata);
+
```
ifranzki: I guess you intended to add that blank line above detach_shm(tokdata) ….

p-steuer: fixed

Issue #230, opened 2019-01-31
opencryptoki tools (pkcsconf) should inform the user that he is not in pkcs11 group


ngothan:
Description of problem:
In the current version the user needs to be added to the pkcs11 group so the tools work for him. It would be really user friendly if the opencryptoki tools would inform the user about this need.

Version-Release number of selected component (if applicable):
opencryptoki-3.11.0-1

How reproducible:
100%

Steps to Reproduce:
1. install opencryptoki{,-libs,-swtok}
2. as non-root run "pkcsconf -t"

Actual results:
Error initializing the PKCS11 library: 0x6 (CKR_FUNCTION_FAILED)

Expected results:
Warning that user is not in pkcs11 group

ngothan:
i attach a proposed patch for this issue: 

~~~
diff -up opencryptoki-3.11.0/usr/sbin/pkcsconf/pkcsconf.c.me opencryptoki-3.11.0/usr/sbin/pkcsconf/pkcsconf.c
--- opencryptoki-3.11.0/usr/sbin/pkcsconf/pkcsconf.c.me 2019-01-31 13:27:05.720647942 +0100
+++ opencryptoki-3.11.0/usr/sbin/pkcsconf/pkcsconf.c    2019-01-31 13:36:05.763624633 +0100
@@ -1072,6 +1072,9 @@ CK_RV init(void)
     if (rc != CKR_OK) {
         printf("Error initializing the PKCS11 library: 0x%lX (%s)\n", rc,
                p11_get_ckr(rc));
+        printf("Note: all non-root users that require access to PKCS#11 tokens "
+               "using opencryptoki must be assigned to the pkcs11 group to be "
+               "able to communicate with the pkcsslotd daemon.\n");
         fflush(stdout);
         cleanup();
     }
~~~

p-steuer:
I turned your patch proposal into a PR: https://github.com/opencryptoki/opencryptoki/pull/231/commits/b29f7c3011c057acae6d0cf12de7d9d7e912fb10 .

p-steuer:
Corresponding PR has been merged. Closing .

PR #231, opened 2019-02-01
Inform the user about required pkcs11 group membership

p-steuer:
See issue https://github.com/opencryptoki/opencryptoki/issues/230 .

### DIFF
--- a/usr/sbin/pkcsconf/pkcsconf.c
+++ b/usr/sbin/pkcsconf/pkcsconf.c
@@ -22,6 +22,8 @@
 #include <memory.h>
 #include <string.h>
 #include <strings.h>
+#include <pwd.h>
+#include <grp.h>
 #include <openssl/crypto.h>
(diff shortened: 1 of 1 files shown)


p-steuer:
updated.

usr/sbin/pkcsconf/pkcsconf.c
```diff
@@ -1072,6 +1120,14 @@ CK_RV init(void)
     if (rc != CKR_OK) {
         printf("Error initializing the PKCS11 library: 0x%lX (%s)\n", rc,
                p11_get_ckr(rc));
+
+        if (check_user_and_group() != CKR_OK) {
+            printf("Note: all non-root users that require access to PKCS#11 "
```
hfreude: consider using fprintf(stderr, ...
however, it looks like the other errors are also printed to stdout instead of stderr.
So if this is a bigger change, let's go with stdout then.

PR #233, opened 2019-02-15
EP11: Support key usage restrictions for public keys

ifranzki:
EP11 does not allow to restrict a public RSA, EC, or DSA keys using CKA_VERIFY=FALSE or CKA_ENCRYPT=FALSE, because it can not be technically enforced since the public key material is available in clear (refer to chapter 2.2.6 of the EP11 structure document). Thus, generating a key pair using C_GenerateKeyPair fails with CKR_TEMPLATE_INCONSISTENT when the public key template contains CKA_VERIFY=FALSE or CKA_ENCRYPT=FALSE.

With this change, the EP11 token allows to restrict public keys with CKA_VERIFY=FALSE or CKA_ENCRYPT=FALSE. When specified, these attributes are overridden with TRUE when passing the key generate request to the EP11 library, but the original attribute values are stored in the object. The key usage restriction in regards of SIGN and ENCRYPT are enforced by the EP11 token, before the sign or encrypt request is passed to the EP11 library.

### DIFF
--- a/usr/lib/ep11_stdll/ep11_specific.c
+++ b/usr/lib/ep11_stdll/ep11_specific.c
@@ -3908,7 +3908,8 @@ static CK_RV dh_generate_keypair(STDLL_TokData_t * tokdata,
     dh_pgs.pg = malloc(p_len * 2);
     if (!dh_pgs.pg) {
         TRACE_ERROR("%s Memory allocation failed\n", __func__);
-        return CKR_HOST_MEMORY;
+        rc = CKR_HOST_MEMORY;
+        goto dh_generate_keypair_end;
     }
     memset(dh_pgs.pg, 0, p_len * 2);
     memcpy(dh_pgs.pg, prime_attr->pValue, p_len);     /* copy CKA_PRIME value */
@@ -4168,8 +4169,10 @@ static CK_RV dsa_generate_keypair(STDLL_TokData_t * tokdata,
         }
     }
 
-    if (prime_attr == NULL || sub_prime_attr == NULL || base_attr == NULL)
-        return CKR_TEMPLATE_INCOMPLETE;
+    if (prime_attr == NULL || sub_prime_attr == NULL || base_attr == NULL) {
+        rc = CKR_TEMPLATE_INCOMPLETE;
+        goto dsa_generate_keypair_end;
+    }
 
     /* copy CKA_PRIME/CKA_BASE/CKA_SUBPRIME to private template */
     rc = build_attribute(CKA_PRIME, prime_attr->pValue,
@@ -4221,7 +4224,8 @@ static CK_RV dsa_generate_keypair(STDLL_TokData_t * tokdata,
     dsa_pqgs.pqg = malloc(p_len * 3);
     if (!dsa_pqgs.pqg) {
         TRACE_ERROR("%s Memory allocation failed\n", __func__);
-        return CKR_HOST_MEMORY;
+        rc = CKR_HOST_MEMORY;
+        goto dsa_generate_keypair_end;
     }
     memset(dsa_pqgs.pqg, 0, p_len * 3);
     memcpy(dsa_pqgs.pqg, prime_attr->pValue, p_len);
@@ -4255,7 +4259,7 @@ static CK_RV dsa_generate_keypair(STDLL_TokData_t * tokdata,
     if (rc != CKR_OK) {
         TRACE_ERROR("%s RSA/EC check public key attributes failed with "
                     "rc=0x%lx\n", __func__, rc);
-        return rc;
+        goto dsa_generate_keypair_end;
     }
 
     rc = check_key_attributes(tokdata, CKK_DSA, CKO_PRIVATE_KEY,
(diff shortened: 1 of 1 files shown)


usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -560,6 +560,72 @@ static CK_RV check_key_attributes(STDLL_TokData_t * tokdata,
...
+        /*
+         * EP11 does not allow to restrict public RSA/DSA/EC keys with
+         * CKA_VERIFY=FALSE and/or CKA_ENCRYPT=FALSE since it can not
+         * technically enforce the restrictions. Therefore override these
+         * attributes for the EP11 library, but keep the original attribute
+         * values in the object.
+         */
+        if ((kt == CKK_EC) || (kt == CKK_RSA) || (kt == CKK_DSA)) {
+            override_types = &override_types_public_key[0];
+            attr_cnt = sizeof(override_types_public_key) / sizeof(CK_ULONG);
```
p-steuer: / sizeof(override_types_public_key[0])

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -560,6 +560,72 @@ static CK_RV check_key_attributes(STDLL_TokData_t * tokdata,
...
+        }
+        break;
+
+    default:
+        return CKR_OK;
+    }
```
p-steuer: i dont get the above switch statement: i mean in case CKO_PUBLIC_KEY, but kt is not rsa, ec, dsa .. in that case loop below does nothing and the case is no different from the default case..

So why not check for all cases when theres nothing to do at the top and return directly ..
```
if (kc != CKO_PUBLIC_KEY || (kt != CKK_ECC && kt != CKK_RSA && kt != CKK_DSA))
    return CKR_OK;
```
.. such that the rest of the code is free of branching.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -3836,7 +3906,8 @@ static CK_RV dh_generate_keypair(STDLL_TokData_t * tokdata,
     dh_pgs.pg = malloc(p_len * 2);
     if (!dh_pgs.pg) {
         TRACE_ERROR("%s Memory allocation failed\n", __func__);
-        return CKR_HOST_MEMORY;
+        rc = CKR_HOST_MEMORY;
+        goto dh_generate_keypair_end;
```
p-steuer: detected bugfix sneaking in ; -)
please make it a seperate commit even if its just 2 locs

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -4051,8 +4167,10 @@ static CK_RV dsa_generate_keypair(STDLL_TokData_t * tokdata,
         }
     }
 
-    if (prime_attr == NULL || sub_prime_attr == NULL || base_attr == NULL)
-        return CKR_TEMPLATE_INCOMPLETE;
+    if (prime_attr == NULL || sub_prime_attr == NULL || base_attr == NULL) {
+        rc = CKR_TEMPLATE_INCOMPLETE;
+        goto dsa_generate_keypair_end;
```
p-steuer: same as above

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -4104,7 +4222,8 @@ static CK_RV dsa_generate_keypair(STDLL_TokData_t * tokdata,
     dsa_pqgs.pqg = malloc(p_len * 3);
     if (!dsa_pqgs.pqg) {
         TRACE_ERROR("%s Memory allocation failed\n", __func__);
-        return CKR_HOST_MEMORY;
+        rc = CKR_HOST_MEMORY;
+        goto dsa_generate_keypair_end;
```
p-steuer: same as above

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -4138,7 +4257,7 @@ static CK_RV dsa_generate_keypair(STDLL_TokData_t * tokdata,
     if (rc != CKR_OK) {
         TRACE_ERROR("%s RSA/EC check public key attributes failed with "
                     "rc=0x%lx\n", __func__, rc);
-        return rc;
+        goto dsa_generate_keypair_end;
```
p-steuer: same as above

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -4148,7 +4267,25 @@ static CK_RV dsa_generate_keypair(STDLL_TokData_t * tokdata,
     if (rc != CKR_OK) {
         TRACE_ERROR("%s RSA/EC check private key attributes failed with "
                     "rc=0x%lx\n", __func__, rc);
-        return rc;
+        goto dsa_generate_keypair_end;
```
p-steuer: same as above

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: fixed

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: Well, yes, at the moment there is only one switch case, so the whole thing could be done as you suggested. Nevertheless, I would like to keep the switch so that other key classes can be added at a later time easily. 

However, I will change the key type check and return CKR_OK if its not one of the types to handle.

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: will do

PR #235, opened 2019-02-21
Fix coverty scan findings for stat/open

ifranzki:
A stat() before open() is not required and can cause timing ralted errors. The code can directly react on the result of open(). Flag O_CREAT will create the file only if it is not existent, so we don't need to know upfront if the file exists already or not.

**For discussion:** 
The only difference in usr/sbin/pkcsslotd/mutex.c is that now the code always uses fchmod() and fchown() on the lock file when the lock file was successfully opened, so also when it was already existent, not only when it was newly created. This happens once when pkcsslotd starts, so IMHO this does not hurt.

### DIFF
--- a/usr/lib/api/apiutil.c
+++ b/usr/lib/api/apiutil.c
@@ -51,15 +51,12 @@ extern API_Proc_Struct_t *Anchor;
 
 CK_RV CreateProcLock(void)
 {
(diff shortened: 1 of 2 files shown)


p-steuer:
lgtm @hfreude can you have a look ?

PR #236, opened 2019-03-04
EP11: Created MACed-SPKIs when importing public keys

ifranzki:
The EP11 host library does not allow to use plain SPKIs as public key blobs for encrypt operations, this requires MACed-SPKIs. Create MACed-SPKIs whenever public keys are imported, and store it in CKA_IBM_OPAQUE instead of the plain SPKI.


p-steuer:
lgtm. i didnt review the test code though. @hfreude can you have a look, too ?

PR #237, opened 2019-03-13
Update soft_specific.c

Hugo0208:
multipart signing calling C_signfinal the first time with a null signature will cause a segfault

### DIFF
--- a/usr/lib/soft_stdll/soft_specific.c
+++ b/usr/lib/soft_stdll/soft_specific.c
@@ -2756,7 +2756,7 @@ static CK_RV softtok_hmac_final(SIGN_VERIFY_CONTEXT *ctx, CK_BYTE *signature,
             if (general)
                 *sig_len = *(CK_ULONG *) ctx->mech.pParameter;
             else
-                *sig_len = (CK_ULONG)mac_len;
+                *sig_len = (CK_ULONG) mac_len;
             return CKR_OK;
         }
     }
--- a/usr/lib/soft_stdll/soft_specific.c
+++ b/usr/lib/soft_stdll/soft_specific.c
@@ -2753,12 +2753,12 @@ static CK_RV softtok_hmac_final(SIGN_VERIFY_CONTEXT *ctx, CK_BYTE *signature,
 
     if (signature == NULL) {
         if (sign) {
-      	    if (general)
+            if (general)
                 *sig_len = *(CK_ULONG *) ctx->mech.pParameter;
             else
-                *sig_len = (CK_ULONG)mac_len;	
-	    return CKR_OK;
-	}
+                *sig_len = (CK_ULONG)mac_len;
+            return CKR_OK;
+        }
     }
 	
     mdctx = (EVP_MD_CTX *) ctx->context;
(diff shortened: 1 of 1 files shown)


ifranzki:
Also when parameter sign is false, then you should not update the `*sig_len`. It won't hurt if you do in the current implementation, but I would tend to only update `*sig_len` when sign is true.

Hugo0208:
Yes you are right.

ifranzki:
Please also sign-off you commit, like mentioned in CONTRIBUTING.md, i.e. like `Signed-off-by: YOUR_NAME <youremail@something.com>`

ifranzki:
You can use git's interactive rebase to change you commit and then push it again to you branch. This will then automatically update this pull request.

usr/lib/soft_stdll/soft_specific.c
```diff
@@ -2751,14 +2751,17 @@ static CK_RV softtok_hmac_final(SIGN_VERIFY_CONTEXT *ctx, CK_BYTE *signature,
...
-	/* 
-	   Check if signature is NULL to allow multipart signing without segfault 
-	   C_signFinal must be called twice, one time to know the size and allocate the memory, one another to perform the signature
-	*/
-	if (signature == NULL) 
-	{
-		*sig_len = (CK_ULONG)mac_len;
-		return CKR_OK;
+
+if (signature == NULL)
```
ifranzki: Much better now.
You still have a mixture of tabs and spaces here. Please use 4 spaces to indent, not tabs.  See https://github.com/opencryptoki/opencryptoki/blob/master/doc/coding_style.md for details.

usr/lib/soft_stdll/soft_specific.c
```diff
@@ -2751,18 +2751,15 @@ static CK_RV softtok_hmac_final(SIGN_VERIFY_CONTEXT *ctx, CK_BYTE *signature,
...
-      	  if (general)
-            *sig_len = *(CK_ULONG *) ctx->mech.pParameter;
-     	  else
-            *sig_len = (CK_ULONG)mac_len;
-	
-	 return CKR_OK;
-	}
+    if (signature == NULL) {
+        if (sign) {
+      	    if (general)
```
ifranzki: Better, but still some tabs...
In line 'if (general)' is a tab, and in line '*sig_len = (CK_ULONG)mac_len;' is a tab at the end of the line. Also in line 'return CKR_OK' and the following line with just a closing bracket.  There is also a tab in the empty line after the code block you added. I would suggest to use an editor that shows whitespace characters. That way you can quickly see the tabs....

Hugo0208:
Sorry, i hope this is good now ^^

usr/lib/soft_stdll/soft_specific.c
```diff
@@ -2756,7 +2756,7 @@ static CK_RV softtok_hmac_final(SIGN_VERIFY_CONTEXT *ctx, CK_BYTE *signature,
             if (general)
                 *sig_len = *(CK_ULONG *) ctx->mech.pParameter;
             else
-                *sig_len = (CK_ULONG)mac_len;
+                *sig_len = (CK_ULONG) mac_len;
             return CKR_OK;
```
ifranzki: Please move the return CKR_OK outside of this block, so that it also returns when sign is false.

Also, please squash you commits together so that it is only one commit. Currently you have 5 commits on top of each other.

ifranzki:
I have squashed and adapted your patches and submitted a new pull request: https://github.com/opencryptoki/opencryptoki/pull/238

PR #241, opened 2019-05-17
SEC1714: Rng changes

ifranzki:


usr/lib/cca_stdll/cca_specific.c
```diff
@@ -235,33 +236,35 @@ CK_RV token_specific_rng(STDLL_TokData_t * tokdata, CK_BYTE * output,
...
 
     UNUSED(tokdata);
 
-    memcpy(form, "RANDOM  ", (size_t) CCA_KEYWORD_SIZE);
+    memcpy(rule_array, "RANDOM  ", (size_t) CCA_KEYWORD_SIZE);
 
     while (bytes_so_far < bytes) {
-        dll_CSNBRNG(&return_code,
-                    &reason_code, NULL, NULL, form, random_number);
+        num_bytes = bytes - bytes_so_far;
```
p-steuer: loop condition feat. 3 vars (num_bytes, bytes, bytes_so_far) hard to read. why not
```
while (bytes) {
    num_bytes = bytes > 8192 ? 8192 : bytes;
    [....]
    bytes -= num_bytes;
}
```

ifranzki: Well, there are many ways to do this...
I guess I just wanted to keep it the way it was before as much as possible.

Issue #229, opened 2018-12-06
make ica token open libica.so.3 instead of libica.so


hfreude:
The ica token should dlopen libica.so.3 instead of libica.so.

p-steuer:
fixed by https://github.com/opencryptoki/opencryptoki/commit/73f05eb53f12197f081fd7ec75619c6ea3a39b2c

PR #245, opened 2019-08-12
Use versioned so name for libica: libica.so.3 

ifranzki:


sharkcz:
LGTM as I've initiated the changes

ifranzki:
This one also fixes https://github.com/opencryptoki/opencryptoki/issues/229

p-steuer:
merged, thanks.

PR #246, opened 2019-08-13
EP11: Use new target handling API and load versioned EP11 host library

ifranzki:


usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -7302,62 +7327,67 @@ static CK_RV get_control_points_for_adapter(uint_32 adapter, uint_32 domain,
...
+        goto out;
     }
     clen = rc;
 
     memset(rsp, 0, sizeof(rsp));
     rlen = sizeof(rsp);
-    rc = dll_m_admin(rsp, &rlen, NULL, NULL, cmd, clen, NULL, 0,
-                     (uint64_t) & target);
+    rc = dll_m_admin(rsp, &rlen, NULL, NULL, cmd, clen, NULL, 0, target);
     if (rc < 0) {
```
hfreude: &target ???
A nice example why this target_t = uin64_t is in fact the wrong type.
The code now has to cast a pointer to the target list (uint64_t*) to uint64_t which is the target parameter type
and the compiler can't help any more.
It should be target_t = uint64_t * or even better const void *.

hfreude: O sorry - I got it. target is here in fact a target_t and so your line is fine.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -1781,6 +1784,60 @@ static CK_RV make_wrapblob(STDLL_TokData_t * tokdata, CK_ATTRIBUTE * tmpl_in,
...
+    char *errstr;
+
+    ep11_lib_name = getenv(EP11SHAREDLIB_NAME);
+    if (ep11_lib_name != NULL) {
+        lib_ep11 = dlopen(ep11_lib_name, RTLD_GLOBAL | RTLD_NOW);
+
+        if (lib_ep11 == NULL) {
+            errstr = dlerror();
+            OCK_SYSLOG(LOG_ERR,
+                       "%s: Error loading shared library '%s' [%s]\n",
```
hfreude: warning instead of error ? Or better just a debug trace and if all attempts to load an ep11 lib fail one error syslog entry? You know what I mean: an admin just seeing the ERROR with no knowledge about the way how the ep11 token comes up will be concerned without a reason.

hfreude: oh sorry -i've been too quick. This first load is for a set environment variable, and should of course be an error if faileed and you did already the other logs as debug.

PR #247, opened 2019-08-16
RFC: Provide an external header file with the supported EC curves

ifranzki:
To make it easier for applications to specify the DER encoded OIDs of the desired EC curve, provide an external, but non-PKCS11-standard header file that contains defines for the supported EC curves. These defines can be used as array initializers. The DER encoded OIDs are specified as attribute CKA_ECDSA_PARAMS or CKA_EC_PARAMS with C_GenerateKeyPair to generate an EC key pair for the desired curve.

While at it, also rename some of the static variables to match the exact EC curve names.

I marked this PR as RFC to first start a discussion about it, but if you are OK with it, then feel free to merge it.


p-steuer:
looks good. If its a public header, id consider using less generic names to avoid name space conflicts, eg PRIME256V1 could easily be defined by some other crypto lib header. I think we already install other headers besides pkcs11.h. Dont we use some OCK specific prefix there ?

ifranzki:
good point. The other external headers mostly define symbols used in the PKCS#11 standard, and those are prefixed with CK_ or CKx_. We certainly can also prefix the curves with CK_ as well, but then it sounds like as if they would be part of the PKCS#11 standard....

Opinions?

PR #248, opened 2019-09-10
Replace local md5 implementation by openssl

jschmidb:


usr/lib/common/mech_md5.c
```diff
@@ -67,81 +128,73 @@ CK_RV md5_hash(STDLL_TokData_t *tokdata,
...
-        TRACE_ERROR("%s received bad argument(s)\n", __func__);
-        return CKR_FUNCTION_FAILED;
-    }
+    UNUSED(tokdata);
+    UNUSED(sess);
 
-    return ckm_md5_update(tokdata, (MD5_CONTEXT *) ctx->context, in_data,
-                          in_data_len);
-}
+    /* if no data to hash, just return */
```
p-steuer: is this the right thing to do ?

usr/lib/common/mech_md5.c
```diff
@@ -11,47 +11,108 @@
 #include <pthread.h>
 #include <string.h>             // for memcmp() et al
 #include <stdlib.h>
+#include <memory.h>
```
p-steuer: what depends on meory.h ?

usr/lib/common/mech_md5.c
```diff
@@ -67,81 +128,73 @@ CK_RV md5_hash(STDLL_TokData_t *tokdata,
...
-                     CK_BYTE length_only,
-                     DIGEST_CONTEXT *ctx,
+CK_RV md5_hash_final(STDLL_TokData_t *tokdata, SESSION *sess,
+                     CK_BYTE length_only, DIGEST_CONTEXT *ctx,
                      CK_BYTE *out_data, CK_ULONG *out_data_len)
 {
-    CK_RV rc;
-
+    UNUSED(tokdata);
+    UNUSED(sess);
```
p-steuer: looks something is wrong with tabs/spaces

usr/lib/common/mech_md5.c
jschmidb: I think yes:
if we don't return here, sw_MD5_Update() gets called, which in turn calls MD5_Update(). This function is already part of OpenSSL and implemented in md32_common.h as HASH_UPDATE. First statement here is: 

>     if (len == 0)
>         return 1;

So we can shorten this path and return in md5_hash_update(). BTW it's the same construction for SHA1.

usr/lib/common/mech_md5.c
jschmidb: Nothing, I will remove this #include. (The same #include is contained in mech_sha.c, where it can be removed as well.)

usr/lib/common/mech_md5.c
jschmidb: The code looks ok in Eclipse: 4 spaces in front of the UNUSED() statements. Maybe a formatting problem here.

usr/lib/common/mech_md5.c
p-steuer: Im asking because md5 of the empty string is d41d8cd98f00b204e9800998ecf8427e . So this is comouted at final ? and just update does not accept empty strings?

jschmidb: Yes, update does not accept/handle empty strings. However, this is true only for md5 and sha1. For example, sha256 in openssl is coded differently and the update routine also processes empty strings. 
I made a quick test with the if-statement removed and, as expected, the opencryptoki digest tests all pass ok. So we may remove this if-stmt, or leave it as it is. 
Background: I adjusted the code in mech_md5.c as much as possible with mech_sha.c for easier maintenance. So this if-stmt came in, because mech_sha.c was coded this way. And it makes sense, because the openssl sw implementations of md5 and sha1 are both given by md32_common.h.
If we decide to remove the if-stmt in mech_md5.c, I would also remove it in mech_sha.c.

Issue #250, opened 2019-09-16
Replace memcmp by CRYPTO_memcmp when comparing tags


p-steuer:
for constant-time-ness.

PR #251, opened 2019-09-16
Replace memcmp by CRYPTO_memcmp when comparing tags

ifranzki:
Fixes: https://github.com/opencryptoki/opencryptoki/issues/250


p-steuer:
looks good to me. i didnt verify that you caught all place tough.

PR #253, opened 2019-10-02
EP11 z15/CEX7 support part 1

ifranzki:


ifranzki:
> Funny, the digest tests have 9 byte on each line …
Yes, all other testcase data also have 9 bytes per line. Don't ask me why.... So I decided to keep it that way.

PR #254, opened 2019-10-08
RFC: Make OCK thread save

ifranzki:
A series of fixes/changes to make Opencryptoki thread save.


p-steuer:
looks good to me. can be included in next release imo. wouldnt expect a hughe perf. hit for single threaded apps.

ifranzki:
rebased to current master

PR #255, opened 2019-10-15
EP11 z15/CEX7 support part 2

ifranzki:


p-steuer:
travis fail looks like openssl <=1.0.2 does not have BN_bn2binpad or is it just a missing include.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -1869,9 +1870,17 @@ static void *ep11_load_host_lib()
         return lib_ep11;
     }
 
-    ep11_lib_name = EP11SHAREDLIB_V2;
+    ep11_lib_name = EP11SHAREDLIB_V3;
     lib_ep11 = dlopen(ep11_lib_name, RTLD_GLOBAL | RTLD_NOW);
 
+    if (lib_ep11 == NULL) {
```
p-steuer: we have this if-statement now 3 times. for future releases it would make sense to have the libnames in an array and probe for lib availability from rightmost element downwards, so adding a new lib version would only mean adding it to the array and not copy the if-statement again.

ifranzki: Yes I agree.

usr/lib/common/mech_ec.c
```diff
@@ -26,6 +26,57 @@
...
 #include "ec_defs.h"
 
+#include "openssl/obj_mac.h"
+#include <openssl/ec.h>
+
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
+/*
+ * Older OpenSLL versions do not have BN_bn2binpad, so implement it here
+ */
+int BN_bn2binpad(const BIGNUM *a, unsigned char *to, int tolen)
```
p-steuer: static

usr/lib/common/mech_ec.c
```diff
@@ -26,6 +26,57 @@
 #include "tok_specific.h"
 #include "ec_defs.h"
 
+#include "openssl/obj_mac.h"
+#include <openssl/ec.h>
+
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
```
p-steuer: check for < 1.0.1 is pkay for the brainpool defines, but for BN_bn2binpad it should be < 1.1.0 afaik.

usr/lib/common/mech_ec.c
```diff
@@ -1113,3 +1180,132 @@ CK_RV ecdh_pkcs_derive(STDLL_TokData_t *tokdata, SESSION *sess,
...
+                               CK_ULONG privkey_len,
+                               CK_BYTE *out_pubkey, CK_ULONG *out_len)
+{
+    EC_GROUP *group = NULL;
+    EC_POINT *point = NULL;
+    CK_ULONG pad_len = 0;
+    BIGNUM *bn_x = NULL;
+    BIGNUM *bn_y = NULL;
+    BN_CTX *ctx = NULL;
+    CK_RV rc = CKR_OK;
```
p-steuer: i generally prefer to initialize rc to some some default general error and set rc=CKR_OK just above the function's exit label (at that point its sure the function really was succesful)

p-steuer:
looks good. going to approve it when last issues are resolved.

usr/lib/common/mech_ec.c
ifranzki: fixed

usr/lib/common/mech_ec.c
ifranzki: fixed

usr/lib/common/mech_ec.c
ifranzki: fixed

ifranzki:
hm travis fails again, let me look at it...

p-steuer:
now its the missing brainpools again. looks like travis openssl does not have them, even its >= 1.0.1. Maybe distro package removed them for licensing/patent/compliance or whatever reason. I suggest to add an #fndef guard for every curve.

ifranzki:
Added an ifdef guard for one curve for now. They were all added in one single commit.

PR #256, opened 2019-10-18
EP11: Extract the public key attributes when unwrapping a private key

ifranzki:
When unwrapping a private key using C_UnwrapKey, the public key attributes (key type dependent) should also be recreated. This was only done for EC keys, but not for RSA, DSA, and DH keys.

Please note that this series of commits is based on top of PR 255 "EP11 z15/CEX7 support part 2", that's why it also contains the commits from the other PR. Once that other PR is merged, I can rebase this one if needed.


p-steuer:
looks good. i didnt go into details of ber encoding. hope we have good test coverage for this. this could be considered a bug fix , no? so id think about structuring it in less commits in case we have to backport it to some older version.

ifranzki:
Well, hard to tell of it is a feature or a fix.... Not sure if we ever need to backport this.
In general I prefer multiple smaller commits, but for  backporting one larger commit is of course easier.

ifranzki:
rebased to current master

PR #257, opened 2019-10-24
CKA_EC_POINT needs to contain the DER-encoded EC-point value Q

ifranzki:
As per PKCS#11 standard, the CKA_EC_POINT attribute needs to contain the DER-encoding of ANSI X9.62 ECPoint value Q. 

This was not the case for the ICA, CCA and EP11 tokens. For CCA and EP11 the raw ECPoint value was used, that is the format byte (indicating uncompressed, compressed or hybrid) followed by X and Y as raw bytes. The ICA token did not even include the format byte, but just X and Y as raw bytes.

With this PR, this is changed to supply the ECPoint value (including the format byte) as DER encoded OCTET STRING as required by the PKCS#11 standard.

Also the EC testcases are changed to supply DER encoded values for the CKA_EC_POINT attribute.


usr/lib/common/asn1.c
```diff
@@ -2892,13 +2929,20 @@ CK_RV der_decode_ECPublicKey(CK_BYTE *data,
...
+        TRACE_DEVEL("ber_encode_OCTET_STRING failed\n");
+        goto cleanup;
+    }
+    rc = build_attribute(CKA_EC_POINT, ecpoint, ecpoint_len, &point_attr);
     if (rc != CKR_OK) {
         TRACE_DEVEL("build_attribute failed\n");
         goto cleanup;
     }
 
+    free(ecpoint);
```
jschmidb: Check "if (ecpoint)" like in line 2754 and 2768 before free()?

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -1681,10 +1683,19 @@ CK_RV token_create_ec_keypair(TEMPLATE * publ_tmpl,
...
+        return rv;
+    }
+
+    if ((rv = build_update_attribute(publ_tmpl, CKA_EC_POINT,
+                                     ecpoint, ecpoint_len))) {
         TRACE_DEVEL("build_update_attribute for q failed rv=0x%lx\n", rv);
+        free(ecpoint);
         return rv;
     }
+    free(ecpoint);
```
jschmidb: if (ecpoint)
    free(ecpoint);   ?
Same in line 1695 above.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -1650,6 +1650,8 @@ CK_RV token_create_ec_keypair(TEMPLATE * publ_tmpl,
     CK_BYTE q[CCATOK_EC_MAX_Q_LEN];
```
jschmidb: In ica_specific.c, line 3888 you increased the q_array by one:
 CK_BYTE q_array[1 + ICATOK_EC_MAX_Q_LEN];

usr/lib/common/asn1.c
ifranzki: Well, ecpoint is always non NULL here. ber_encode_OCTET_STRING sets it. Either ber_encode_OCTET_STRING retutrns an error, then we don't reach here, otherwise it is set, and we can free it unconditionally.

BTW:  the NULL check before free() is superfluous anyway. free() can be called with a NULL pointer, it is just a NOP then, see man page of free.

usr/lib/cca_stdll/cca_specific.c
ifranzki: Right, but the CCA key token already contains the q value including the format byte. CCATOK_EC_MAX_Q_LEN is defined as 133, i.e. 1 + 2 * 66.

PR #258, opened 2019-10-30
EP11: Fix possible memory leak

ifranzki:


usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -5401,6 +5402,7 @@ static CK_RV ep11_ende_crypt_init(STDLL_TokData_t * tokdata, SESSION * session,
     rc = h_opaque_2_blob(tokdata, key, &blob, &blob_len, &key_obj);
     if (rc != CKR_OK) {
         TRACE_ERROR("%s no blob rc=0x%lx\n", __func__, rc);
+        free(ep11_state);
```
p-steuer: would you mind using an error exit label instead which frees ep11_state if its != NULL, so adding future exit points would be less error-prone ?

ifranzki: Changed to use an error exit. Please take a look again.

PR #259, opened 2019-11-06
update pin and data store encryption

p-steuer:


ifranzki:
The fact that you have changed the SHM block size may cause problems for existing installations that upgrade to this version. Even if one stays with the old token version, the SHM size has been changed anyway. As you have seen this may result in SHM size mismatch errors.

So at least a hint that the SHM files under/dev/shm must be removed should be added. Maybe you can even add code into shared_memory.c to automatically update existing SHM files if they exists but are in the old size...

usr/sbin/pkcscca/pkcscca.h
```diff
@@ -118,13 +118,35 @@ typedef struct _TWEAK_VEC {
...
     CK_TOKEN_INFO_32 token_info;
 
     CK_BYTE user_pin_sha[3 * DES_BLOCK_SIZE];
     CK_BYTE so_pin_sha[3 * DES_BLOCK_SIZE];
     CK_BYTE next_token_object_name[8];
     TWEAK_VEC tweak_vector;
+
+    /* new for tokversion >= 3.12 */
+    TOKEN_DATA_VERSION dat;
 } TOKEN_DATA;
```
ifranzki: You will also need the TOKEN_DATA_OLD struct here, and also use it in the copied code in pkcscca.c. There are copies of some functions from loadsave.c with light modifications (TRACE macros are different, etc).

PR #263, opened 2019-11-19
Fix the pkcsep11_migrate tool

ifranzki:
Since commit 2668e8f347ca2514b4274daccf71defce0e2313f the contents of attribute CKA_IBM_OPAQUE has been changed to contain the raw EP11 blob directly, no longer wrapped into struct ep11_opaque.

Since then, the pkcsep11_migrate tool could not have worked to re-encrypt EP11 key blobs stored in token objects.

Correct the tool and while at it also replace some older, now deprecated EP11 host library functions with ist newer equivalents.

### DIFF
--- a/usr/sbin/pkcsep11_migrate/pkcsep11_migrate.c
+++ b/usr/sbin/pkcsep11_migrate/pkcsep11_migrate.c
@@ -79,18 +79,15 @@ typedef struct {
 } __attribute__ ((packed)) ep11_target_t;
 
 
-#define blobsize 2048*4
+#define BLOBSIZE         2048*4
 
-typedef struct {
-    size_t blob_size;
-    size_t blob_id;
-    unsigned char blob[blobsize];
-} ep11_opaque;
 
-static int reencrypt(CK_SESSION_HANDLE session, CK_ULONG obj, CK_BYTE *old)
+
+static int reencrypt(CK_SESSION_HANDLE session, CK_ULONG obj, CK_BYTE *old,
+                     CK_ULONG old_len)
 {
-    CK_BYTE req[blobsize];
-    CK_BYTE resp[blobsize];
+    CK_BYTE req[BLOBSIZE];
+    CK_BYTE resp[BLOBSIZE];
     CK_LONG req_len;
     size_t resp_len;
     struct ep11_admresp rb;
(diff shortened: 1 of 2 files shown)


usr/sbin/pkcsep11_migrate/pkcsep11_migrate.c
```diff
@@ -100,13 +97,11 @@ static int reencrypt(CK_SESSION_HANDLE session, CK_ULONG obj, CK_BYTE *old)
     target_t target = XCP_TGT_INIT;
     CK_RV rc;
     CK_BYTE name[256];
-
-    ep11_opaque *op_old = (ep11_opaque *) old;
-    ep11_opaque op_new;
-
+    unsigned char blob[blobsize];
```
hfreude: Please rework the code and rename blobsize to BLOBSIZE.

usr/sbin/pkcsep11_migrate/pkcsep11_migrate.c
```diff
@@ -100,13 +97,11 @@ static int reencrypt(CK_SESSION_HANDLE session, CK_ULONG obj, CK_BYTE *old)
...
+    CK_ULONG blob_len;
 
     CK_ATTRIBUTE opaque_template[] = {
-        {CKA_IBM_OPAQUE, &op_new, sizeof(op_new)}
+        {CKA_IBM_OPAQUE, blob, sizeof(blob)}
```
hfreude: I struggled over: blob is a pointer but sizeof(blub) is the size of blob[]. Can we make this clear with using BLOBSIZE instead ?

usr/sbin/pkcsep11_migrate/pkcsep11_migrate.c
```diff
@@ -168,40 +162,46 @@ static int reencrypt(CK_SESSION_HANDLE session, CK_ULONG obj, CK_BYTE *old)
...
+                old_len, lrb.pllen, resp_len, req_len);
         rc = -5;
         goto out;
     }
 
-    memset(&op_new, 0, sizeof(op_new));
-    op_new.blob_id = op_old->blob_id;
-    op_new.blob_size = op_old->blob_size;
-    memcpy(op_new.blob, lrb.payload, op_new.blob_size);
+    memset(&blob, 0, sizeof(blob));
```
hfreude: what is &blob now ? the address of the pointer ?

usr/sbin/pkcsep11_migrate/pkcsep11_migrate.c
ifranzki: Fixed

usr/sbin/pkcsep11_migrate/pkcsep11_migrate.c
ifranzki: fixed

usr/sbin/pkcsep11_migrate/pkcsep11_migrate.c
ifranzki: fixed. Actually &blob and just blob is the same in this context.

ifranzki:
> Ingo, as far as I have in mind is 'XCP' a protected term - copyright or something.
> We should not use it.

The term XCP is used over and over in the header files that are publicly available in the EP11 Host library devel packages, as well as in the Structure Document also part of the devel packages. So we are just using what is publicly available.

Also, many functions, structures, defines, etc provided by the EP11 library are named 'xcp....' so there is no way of not using this term.

PR #270, opened 2020-01-27
Fix return code checking when using template_attribute_find

jschmidb:
template_attribute_find returns true or false, which was
not handled correctly here.


usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -533,9 +533,8 @@ static CK_RV cleanse_attribute(TEMPLATE *template,
     CK_RV rc;
```
p-steuer: rc looks unused now.

jschmidb: Yes, you're right ... I will remove this line.

PR #273, opened 2020-02-06
Dilithium stage1

jschmidb:
This is Dilithium support stage1: key generation and sign/verify.


usr/lib/common/h_extern.h
```diff
@@ -2478,6 +2478,17 @@ CK_RV ecdsa_priv_unwrap_get_data(TEMPLATE *tmpl, CK_BYTE *data,
 CK_RV ec_priv_unwrap(TEMPLATE *tmpl,
                      CK_BYTE *data, CK_ULONG data_len, CK_BBOOL isOpaque);
 
+// Dilithium routines
+//
+CK_RV dilithium_publ_check_required_attributes(TEMPLATE *tmpl, CK_ULONG mode);
```
p-steuer: Its a little weird to have the headers in a separate commit from the function bodies. Id rather squash the first to commits (dilithium support) and keep the test cases in an extra commit.

p-steuer: .. or at least just move the headers to the second commit if you want to keep the define stuff separate.

p-steuer:
hmm travis seems to have stop working. have to look into that

usr/lib/common/h_extern.h
jschmidb: No that's fine. I almost decided to put the first two commits into one myself.

usr/lib/common/key.c
```diff
@@ -3185,6 +3185,335 @@ CK_BBOOL kea_priv_check_exportability(CK_ATTRIBUTE_TYPE type)
...
+        }
+    }
+
+    found = template_attribute_find(tmpl, CKA_IBM_DILITHIUM_KEYFORM, &attr);
+    if (!found) {
+        if (mode == MODE_CREATE) {
+            TRACE_ERROR("%s\n", ock_err(ERR_TEMPLATE_INCOMPLETE));
+            return CKR_TEMPLATE_INCOMPLETE;
+        }
+    }
```
p-steuer: I think the above functions would be more concise if they just implemented a loop like
```
CK_ULONG req_attrs[] = {....}, i;
for (i = 0; i < sizeof(req_attrs) / sizeof(req_attrs[0]; i++) {
  if (template_attribute_find(tmpl, CKA_IBM_DILITHIUM_KEYFORM, &attr)! && mode == MODE_CREATE) {
    TRACE_ERROR("%s\n", ock_err(ERR_TEMPLATE_INCOMPLETE));
    return CKR_TEMPLATE_INCOMPLETE;
  }
}
```

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -329,6 +329,13 @@ static const version_req_t edwards_req_versions[] = {
 };
 #define NUM_EDWARDS_REQ sizeof(edwards_req_versions)/sizeof(version_req_t)
 
+static const CK_VERSION cex7p_dilithium_support = { .major = 7, .minor = 15 };
+
+static const version_req_t dilithium_req_versions[] = {
+        { .card_type = 7, .min_firmware_version = &cex7p_dilithium_support }
+};
+#define NUM_DILITHIUM_REQ sizeof(dilithium_req_versions)/sizeof(version_req_t)
```
p-steuer: add parantheses here just to be sure

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -4601,6 +4612,245 @@ static CK_RV rsa_ec_generate_keypair(STDLL_TokData_t * tokdata,
...
+    rc = override_key_attributes(tokdata, ktype, CKO_PRIVATE_KEY,
+                                 new_pPrivateKeyTemplate,
+                                 new_ulPrivateKeyAttributeCount);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("%s Dilithium override private key attributes failed with "
+                    "rc=0x%lx\n", __func__, rc);
+        goto error;
+    }
+
+    /* debug */
```
p-steuer: did you intend to put it inside DEBUG ifdef ?

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -4601,6 +4612,245 @@ static CK_RV rsa_ec_generate_keypair(STDLL_TokData_t * tokdata,
...
+        goto error;
+    }
+
+    memset(publblob, 0, sizeof(publblob));
+    memset(privblob, 0, sizeof(privblob));
+
+    memcpy(publblob, spki, spki_len);
+    publblobsize = spki_len;
+
+    memcpy(privblob, privkey_blob, privkey_blob_len);
```
p-steuer: imo privkey_blob can (and should) be erased afterwards.

testcases/crypto/dilithium_func.c
```diff
@@ -0,0 +1,297 @@
...
+        PRINT_ERR("ERROR do_GetFunctionList() Failed , rc = 0x%0x\n", rc);
+        return rc;
+    }
+
+    memset(&cinit_args, 0x0, sizeof(cinit_args));
+    cinit_args.flags = CKF_OS_LOCKING_OK;
+
+    funcs->C_Initialize(&cinit_args);
+
+    {
```
p-steuer: Whats the idea behind checking for the correct GetFunctionState resp CancelFunction return values in the dilithium test case ?

usr/lib/common/key.c
```diff
@@ -3185,6 +3185,335 @@ CK_BBOOL kea_priv_check_exportability(CK_ATTRIBUTE_TYPE type)
...
+{
+    CK_ATTRIBUTE *attr = NULL;
+    CK_BBOOL found;
+
+    found = template_attribute_find(tmpl, CKA_IBM_DILITHIUM_RHO, &attr);
+    if (!found) {
+        if (mode == MODE_CREATE) {
+            TRACE_ERROR("%s\n", ock_err(ERR_TEMPLATE_INCOMPLETE));
+            return CKR_TEMPLATE_INCOMPLETE;
+        }
```
p-steuer: so in case mode != MODE_CREATE, its okay if the required attribute is not found ?

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: Whats the reason for copying privkey_blob to privblob at all? Can't you just use the key blob from privkey_blob ? 
Same for the public blob. Why copy it?
Other key pair generating functions (e.g. RSA, EC, DH, DSA) in EP11 do not copy it, so why does this function?

usr/include/pkcs11types.h
```diff
@@ -542,6 +543,18 @@ typedef CK_ULONG CK_ATTRIBUTE_TYPE;
 #define CKA_IBM_STRUCT_PARAMS     (CKA_VENDOR_DEFINED +0x10009)
 #define CKA_IBM_STD_COMPLIANCE1   (CKA_VENDOR_DEFINED +0x1000a)
 
+/* For Dilithium, oid = 1.3.6.1.4.1.2.267.1.6.5 */
+#define IBM_DILITHIUM_KEYFORM_ROUND2        1
+
+#define CKA_IBM_DILITHIUM_KEYFORM (CKA_VENDOR_DEFINED +0xd0001)
```
ifranzki: To be pedantic ;-) Please add a blank beween the '+'and '0x....'.
I see that others above also do it wrong, but lets do it right for new code.

usr/lib/common/key.c
ifranzki: Good point. I see that all the other already existing functions for other key types do it the same way. This does not necessarrily mean that the others do it right ;-)
These functions are called when creating a new object, which is either Import (MODE_CREATE) or Key generation (MODE_KEYGEN). In addition, they are called when copying an object (MODE_COPY).
For KEYGEN these attributes are possibly not required, because they are built and added by the key generation anyway. For COPY it could be assumed that the copied object already is valid, otherwise it would not exist. For IMPORT (CREATE), the attributes need to be checked, which is performed by the code. 
So the code 'might' be OK as is.....

usr/lib/common/key.c
```diff
@@ -3185,6 +3185,335 @@ CK_BBOOL kea_priv_check_exportability(CK_ATTRIBUTE_TYPE type)
     return TRUE;
 }
 
+//  dilithium_publ_set_default_attributes()
+//
+CK_RV dilithium_publ_set_default_attributes(TEMPLATE *tmpl, CK_ULONG mode)
```
ifranzki: Would it make sense to change the function name of those dilithium functions to include 'ibm' as part of name? When official PKCS11 support for dilithium comes at some point in time, we might want to distinguish between the IBM specific key type and the PKCS11 key type. The PKCS11 key type will most likely have different attributes.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -4601,6 +4612,245 @@ static CK_RV rsa_ec_generate_keypair(STDLL_TokData_t * tokdata,
...
+        goto error;
+    }
+    rc = template_update_attribute(priv_tmpl, attr);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("%s template_update_attribute failed with rc=0x%lx\n",
+                    __func__, rc);
+        goto error;
+    }
+
+    /* Decode SPKI */
```
ifranzki: Starting from here, the code is specific to the IBM version of the dilithium mechanism and key type.  Should we prepare now for an potentially upcoming PKCS11 dlilithium mechanism and move the extraction of the IBM specific attributes into a separate function, to have it easier in the future? Or should we only care later when this happens?

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -4683,6 +4933,15 @@ CK_RV ep11tok_generate_key_pair(STDLL_TokData_t * tokdata, SESSION * sess,
                                   pPrivateKeyTemplate,
                                   ulPrivateKeyAttributeCount, sess->handle);
         break;
+    case CKM_IBM_DILITHIUM:
+        rc = dilithium_generate_keypair(tokdata, sess, pMechanism,
```
ifranzki: Maybe also name the function dilithium**_ibm**_generate_keypair ?

testcases/crypto/dilithium_func.c
```diff
@@ -0,0 +1,297 @@
...
+    };
+
+    /* Generate Dilithium key pair */
+    rc = funcs->C_GenerateKeyPair(session, &mech,
+                   dilithium_attr_public, 1,
+                   dilithium_attr_private, 1,
+                   &publ_key, &priv_key);
+    testcase_new_assertion();
+    if (rc != CKR_OK) {
+        if (rc == CKR_MECHANISM_PARAM_INVALID ||
```
ifranzki: Why this check? Shouldnt the above C_getMechanismInfo have checked if that mech is supported?

usr/lib/common/key.c
p-steuer: okay, so why not check the mode directly at the top and return in case != CREATE . I mean right now the code does everything to hide its doing essentially nothing in that case.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: ok, done.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: This trace is indeed intended to be displayed with trace level 4, but we could also decide to move it to trace level 5 and put a #ifdef DEBUG around it. The same piece of code shows up in rsa_ec_generate_keypair(). So whatever we decide, we should at least have the same logic for these two code pieces.

usr/include/pkcs11types.h
jschmidb: ok, done.

testcases/crypto/dilithium_func.c
jschmidb: Hmm ... this check does in fact more than necessary with our current Dilithium support: Dilithium does not support a mechanism parameter, so CKR_MECHANISM_PARAM_INVALID cannot be returned. But we also don't have special attributes, like EC with different curves, where some curves are not supported. I therefore removed this check completely. Let me know if you think we should be more defensive here ...

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: Good point ... in fact this code is not necessary. I removed it, retested, and saw no difference. The same piece of code also appears in rsa_ec_generate_keypair(), where it is also superfluous. I also removed the code there and retested with same result. So I'm now pushing the code change for Dilithium, but let's keep in mind to change rsa_ec_generate_keypair() accordingly in a separate pull request.

testcases/crypto/dilithium_func.c
jschmidb: Good point. All of our opencryptoki testcases, without any exception, show this same piece of code. Dilithium is just the last one adapting this. When looking at the implementations of C_GetFunctionStatus and C_CancelFunction, it appears that they check if opencryptoki is initialized and would return CKR_CRYPTOKI_NOT_INITIALIZED if it's not. So this check just makes sure that it's ok to start with the testcase, probably to avoid strange errors later.

jschmidb: To make this more clear, we could introduce one common macro, like CHECK_CRYPTOKI_INITIALIZED and use this in all testcases.

usr/lib/common/h_extern.h
jschmidb: Just squashed the first two commits into one.

usr/lib/common/key.c
jschmidb: ok, I renamed the 6 attribute-related functions, the dilithium_generate_keypair function, and also the version-checking related constants by pre-pending an "ibm_". I re-tested after these changes to make sure I didn't forget a code piece.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: Hmm ... here I would leave the code as is for now. We already renamed the function to ibm_dilithium_generate_keypair(), so we will most likely have a separate routine for an "official" Dilithium standard. But we may definitely put some rework on this routine if IBM would introduce new attributes.

ifranzki: agreed.

usr/lib/common/key.c
jschmidb: Yes, that's true. But as all other *_priv_check_required_attributes and *_publ_check_required_attributes routines are coded in the same way I would leave it as is for now. Let's then re-code all affected routines in a separate "housekeeping" pull-request.

usr/lib/ep11_stdll/ep11_specific.c
p-steuer: okay, just the "debug" comment made me think you intended to change something here

usr/lib/common/key.c
p-steuer: Id prefer to have it done now instead postponing it until the day the cows come home (I mean IF the additional effort is not too high).

jschmidb: ok, just pushed the changes. I also changed ibm_dilithium_publ_check_required_attributes() to have at least these two routines with the same logic. The cows may come home then ;-)

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: ok, done.

PR #274, opened 2020-02-13
Add thread id to the trace line output.

hfreude:
For multithread debugging the opencryptoki traces now have an
additional field 'tread id' which is simple the gettid() value.

### DIFF
--- a/usr/lib/common/trace.c
+++ b/usr/lib/common/trace.c
@@ -21,6 +21,7 @@
 #include <unistd.h>
 #include <sys/file.h>
 #include <sys/types.h>
+#include <sys/syscall.h>
 
 #include "pkcs11types.h"
 #include "defs.h"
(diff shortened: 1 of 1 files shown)


usr/lib/common/trace.c
```diff
@@ -217,6 +222,7 @@ void ock_traceit(trace_level_t level, const char *file, int line,
     char buf[1024];
```
ifranzki: I assume buf is still large enough so that the longest trace message plus the tid value still fits?

hfreude: Well, I think the longest trace message from ock I saw was in the range of about 100 characters.
As long as snprintf is used to avoid writing beyond the buffer space I don't really care.

usr/lib/common/trace.c
```diff
@@ -217,6 +222,7 @@ void ock_traceit(trace_level_t level, const char *file, int line,
...
     char *pbuf;
     int buflen, len;
+    pid_t tid;
```
p-steuer: you want to put that inside ifdef, too

usr/lib/common/trace.c
```diff
@@ -29,6 +30,10 @@
 #include "trace.h"
 #include "ock_syslog.h"
 
+#ifdef SYS_gettid
+#define gettid() syscall(SYS_gettid)
```
p-steuer: rename gettid macro to avoid collision with gettid function

PR #283, opened 2020-03-16
Multi-threading fixes, multi-instance fixes, locking fixes

ifranzki:
This patch series reworks the opencryptoki code to support or correct
- multi-threading.
- multiple instances of the same token.
- Locking of btrees, objects, sessions, etc.

Short summary of what the patch series changes:
- Remove unused global variables.
- make global constants const.
- move various global variables into token specific data structures.
- correct locking of session state.
- correct fork, initialization, and termination handling.
- move global btrees into token specific data structure to support multi-instance tokens.
- rework btrees to unify callback function usage.
- use a mutex per btree to lock concurrent access to the btree.
- fix race conditions with btrees.
- add reference counts for anything that is contained in btrees (objects, sessions, etc).
- Correct synchronization of token objects via shared memory.
- Add a rw-lock into each (pkcs11-) object to protect from concurrent attribute changes.
- remove the global API lock


p-steuer:
just .. wow. need some time to go through all of it.

ifranzki:
Yep, take you time.....

Still need to test with TPM and ICSF token.

ifranzki:
fixed some compiler warnings brought up by travis...

Did you notice that the travis builds for x86 fails due to some trousers errors? I don't think its related to my changes, but more a general problem, since it happens even before the build starts.

p-steuer:
yes i noticed it, but dont know how to fix it yet.

you could try to install trousers in a seperate command in .travis.yaml and ignore the error:
```
- sudo apt-get install -y trousers || true
```

ifranzki:
fixed some ICSF related issues. ...

p-steuer:
note to myself: stopped at db4aade514e8363297643ba1620b414e5ffef17e

usr/lib/api/api_interface.c
```diff
@@ -56,22 +56,37 @@ void api_init();
...
+    /*
+     * Reinitialize trace so that the trace output appears under the new
+     * process's trace file, and not in the parent process's once.
+     * C_Finalize will pass the new trace handle to the tokens, so that their
+     * finalization code will also trace into the new trace file.
+     */
+    trace_finalize();
+    trace_initialize();
+    /*
+     * Terminate all slots. This will also free the Anchor.
```
p-steuer: Anchor is also set to NULL after free ? If so id probably state that explicitely here as well, since Anchor is checked for NULL.

usr/lib/api/api_interface.c
```diff
@@ -56,22 +56,37 @@ void api_init();
...
+     * A forked child needs to call C_Initialize again to use Opencryptoki.
+     * Indicate that we are in the fork initializer.
+     * A forked child process has the same mapped memory as the parent, but the
+     * reference count as part of the shared memory was not increased during
+     * the fork, so we should not decrement it when terminating.
```
p-steuer: I personally found the explaination of what we are not doing confusing. Id rather say that the child is no PKCS11 application after fork (so no refcount needed). If it becomes a PKCS11 application by calling Initialize, refcount will be handled just like in the parent, right ?

p-steuer:
stopped at 03795d2714934b69dd3b01b0e1622a933dbe9cc3

usr/lib/api/api_interface.c
ifranzki: added this to the comment.

usr/lib/api/api_interface.c
ifranzki: rephrased the comment, please take a look

ifranzki:
Also fixed the C_Login race condition and rebased to current master.

usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -63,7 +65,10 @@ const char label[] = "IBM ICA  PKCS #11";
 static pthread_mutex_t rngmtx = PTHREAD_MUTEX_INITIALIZER;
 
 #define LIBICA_SHARED_LIB "libica.so.3"
-#define BIND(dso, sym)  (*(void **)(&p_##sym) = dlsym(dso, #sym))
+#define BIND(dso, sym)  do {                                             \
+                            if (p_##sym == NULL)                         \
```
p-steuer: wrt multi intance token, i think all global token data should either be moved to the token data structure or locked. Those kind of checks still leave a small window for race.

testcases/misc_tests/fork.c
```diff
@@ -0,0 +1,357 @@
...
+        testcase_fail("C_Finalize (client) rc = %s", p11_get_ckr(rv));
+        goto out;
+    }
+out:
+
+    testcase_print_result();
+    exit(rv);
+}
+
+int main(int argc, char **argv)
```
p-steuer: awsome testcase. adding a test case for forking behvaior was really overdue.

testcases/misc_tests/multi_instance.c
```diff
@@ -0,0 +1,410 @@
...
+                                  hPubKey, hPrivKey);
+    if (rv != CKR_OK) {
+        testcase_fail("C_GenerateKeyPair rc = %s", p11_get_ckr(rv));
+        return rv;
+    }
+    return CKR_OK;
+}
+
+
+int main(int argc, char **argv)
```
p-steuer: Great test case as well. Could you add a RWEADME(.md) file in the test case subdir, which lists all tests which have to invoked by hand i.e., do not run with the test suite ? Im afraid w/o one, those special test cases are forgotten or run seldomly.

usr/lib/common/obj_mgr.c
```diff
@@ -749,16 +749,25 @@ CK_RV object_mgr_create_final(STDLL_TokData_t *tokdata,
...
-void destroy_object_cb(STDLL_TokData_t *tokdata, void *node)
+CK_RV object_mgr_destroy_object(STDLL_TokData_t *tokdata,
+                                SESSION *sess, CK_OBJECT_HANDLE handle)
 {
-    OBJECT_MAP *map = (OBJECT_MAP *) node;
+    CK_RV rc = CKR_OK;
+    OBJECT_MAP *map;
     OBJECT *o;
 
+    if (!sess) {
```
p-steuer: looks like sess param has been and still is unused except for checking != NULL. Why not remove it ?

usr/lib/common/lock_btree.c
```diff
@@ -379,5 +375,28 @@ void bt_destroy(struct btree *t, void (*delete_func)(void *))
...
+    t->top = NULL;
+    t->size = 0;
+    t->free_nodes = 0;
+
+    /*
+     * Need a recursive mutex, because btree callback functions may call
+     * btree functions again
+     */
+    pthread_mutexattr_init(&attr);
+    pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_RECURSIVE);
```
p-steuer: Using a lock per btree is certainly the right way to go. The problem with recursive locking is a) kernel has to keep track of a lot of things so its usually slow b) pthreads doesnt know recursive rwlocks so you have to make it a mutex which blocks read ops. And id say once an application has set up everything (sessions, created keys ..), most of th the ops on btrees would be r/o. So if there is a chance we could get rid of btree's recursive calls id consider going with rw locks.

usr/lib/common/btree.c
```diff
@@ -36,37 +69,24 @@
...
-                temp = temp->right;
-            } else {
-                /* If the bit is 0, traverse left */
-                temp = temp->left;
-            }
-            i >>= 1;
-        }
-done:
-        return ((temp->flags & BT_FLAG_FREE) ? NULL : temp);
+        return __bt_get_node(t, node_num);
```
p-steuer: well, im sure you tested it, but does this actually work: calling a function from atomic transaction w/o marking it transaction save ?

p-steuer:
stopped at 772b6e85b68f320559473724881a95f1fdd81474

usr/lib/ica_s390_stdll/ica_specific.c
ifranzki: Well, I thought about this, but its not really needed.
Token initialization is done during C_Initialize. The tokens are initialized one after the other (for loop that calls DL_Load_and_Init() ). The whole initialization process is locked using the GlobMutex in api_interface.c, so it can't run concurrently at the same time.
Thus it can not happen that two tokens get the symbols from the libraries at the same time. If 2 instances of the same token are initialized one after the other, the second one will either see that the symbol has already been loaded (ICA token), or will load the same symbol again (EP11), which does not hurt either.
So I would tend to leave it as is.

testcases/misc_tests/multi_instance.c
ifranzki: Good idea. Will do.

usr/lib/common/obj_mgr.c
ifranzki: Good catch. 
I will leave the parameter there for consistency, since most other obj_mgr_xxx functions also get a session as parameter, but I remove the NULL check and mark it as UNUSED instead.

usr/lib/common/lock_btree.c
ifranzki: The callbacks for freeing a node are not critical, none of these callbacks use btree functions currently. 

The critical callbacks are those for bt_for_each_node(). There are several different callbacks in the code, and some of them do in fact use btree functions inside the callback. 

If they would only do read-type of locks, then a rw lock could still be OK (since rw locks allow multiple readers, even for the same thread), however, I am not sure if the callbacks are only doing read operations. I fear they also do write operations (e.g. bt_node_free(), bt_put_node_value()).

So as of today, I don't see a chance to change this to non-recursive rw locks.

usr/lib/common/btree.c
ifranzki: As far as I remember a function that does not use anything that makes it transaction-unsafe, is implicitly transaction-save, and can thus be called within a transaction block.

ifranzki:
Maked sess as unused in object_mgr_destroy_object() and added info to README in testcases/misc_test/. Also rebased to current master.

usr/lib/common/btree.c
p-steuer: i also cannot reproduce the old clobbered warnings with the tm build, so it seems you fixed that one as well ;-)

usr/lib/ica_s390_stdll/ica_specific.c
p-steuer: ah okay. i assumed the token_specific_init function was actually called at CTokenInit, but your right. C_Initialize should be locked anyway.

usr/lib/common/lock_btree.c
p-steuer: im okay with leaving it as is for now. first make it work. we can think about perf improvements later on.

usr/lib/api/api_interface.c
```diff
@@ -803,7 +803,7 @@ CK_RV C_DestroyObject(CK_SESSION_HANDLE hSession, CK_OBJECT_HANDLE hObject)
     STDLL_FcnList_t *fcn;
     ST_SESSION_T rSession;
 
-    TRACE_INFO("C_DestrypObject\n");
+    TRACE_INFO("C_DestryObject\n");
```
p-steuer: C_DestroyObject

usr/sbin/pkcsconf/pkcsconf.c
```diff
@@ -862,8 +862,12 @@ void print_token_info(int slot_id, CK_TOKEN_INFO *TokenInfo)
 
     printf("\tSessions: %lu/%lu\n", TokenInfo->ulSessionCount,
            TokenInfo->ulMaxSessionCount);
-    printf("\tR/W Sessions: %lu/%lu\n", TokenInfo->ulRwSessionCount,
-           TokenInfo->ulMaxRwSessionCount);
+    if (TokenInfo->ulRwSessionCount == CK_UNAVAILABLE_INFORMATION)
```
p-steuer: Why not also check for CK_UNAVAILABLE_INFORMATION and CK_EFFECTIVELY_INFINITE for other fields ?
```
CK_TOKEN_INFO Note: The fields ulMaxSessionCount, ulSessionCount, ulMaxRwSessionCount, 
ulRwSessionCount, ulTotalPublicMemory, ulFreePublicMemory, ulTotalPrivateMemory, 
and ulFreePrivateMemory can have the special value CK_UNAVAILABLE_INFORMATION, 
which means that the token and/or library is unable or unwilling to provide that information. 
In addition, the fields ulMaxSessionCount and ulMaxRwSessionCount can have the special value 
CK_EFFECTIVELY_INFINITE, which means that there is no practical limit on the number of sessions
(resp. R/W sessions) an application can have open with the token.

It is important to check these fields for these special values.  This is particularly true for 
CK_EFFECTIVELY_INFINITE, since an application seeing this value in the ulMaxSessionCount or 
ulMaxRwSessionCount field would otherwise conclude that it can’t open any sessions with the token, 
which is far from being the case.
```

usr/include/stdll.h
```diff
@@ -88,7 +88,8 @@ typedef CK_RV (CK_PTR ST_C_OpenSession) (STDLL_TokData_t *tokdata,
                                          CK_SESSION_HANDLE_PTR phSession);
 
 typedef CK_RV (CK_PTR ST_C_CloseSession) (STDLL_TokData_t *tokdata,
-                                          ST_SESSION_T *hSession);
+                                          ST_SESSION_T *hSession,
+                                          CK_BBOOL in_fork_initializer);
```
p-steuer: i possibly misunderstood this patch but is the additional function param only used to propagate "in_child_fork_initializer" which is a global variable anyway ?

usr/lib/api/api_interface.c
ifranzki: LOL. Fixed one typo with another.

usr/sbin/pkcsconf/pkcsconf.c
ifranzki: Good point. Will change this for the other fields, and also handle CK_EFFECTIVELY_INFINITE.

usr/include/stdll.h
ifranzki: Yes it is to propagate in_child_fork_initializer. 
Variable in_child_fork_initializer is a global variable in the API shared library, but it is not available in the token shared libraries, so it has to be passed over.

usr/lib/common/lock_sess_mgr.c
```diff
@@ -88,6 +88,12 @@ CK_RV session_mgr_new(CK_ULONG flags, CK_SLOT_ID slot_id,
     so_session = session_mgr_so_session_exists();
     user_session = session_mgr_user_session_exists();
 
+    if (pthread_rwlock_wrlock(&sess_list_rwlock)) {
+        TRACE_ERROR("Read Lock failed.\n");
```
hfreude, line 98: Traces message is wrong - this is an rwlock not a read lock.
This is false for all the traces following in this patch.

usr/lib/common/sess_mgr.c
```diff
@@ -90,21 +90,23 @@ CK_RV session_mgr_new(CK_ULONG flags, CK_SLOT_ID slot_id,
     // we don't have to worry about having a user and SO session at the same
     // time. that is prevented in the login routine
     //
-    if (user_session) {
```
hfreude, line 93: I see you changed the transactional memory version and the locking version.
Makes sence as it is nearly impossible to just rework the locking code.
However, we should talk about if we want to keep the transactional memory code anyway - as this still means to
have two versions to maintain, test and verify.

usr/lib/common/lock_sess_mgr.c
ifranzki: good catch. Fixing this....

PR #284, opened 2020-03-25
Fix build with openssl master

p-steuer:
The U64 macro seems to be removed from public header file.

edit: There are still warnings regarding the next openssl release's function deprecations but for now i only want a working build.


hfreude:
Patrick, what did the U64 macro do ?
If this was some hton for u64 then your fix would produce the wrong values on a little endian plattform. Please check this.

p-steuer:
I checked it before doing the change. As Ingo said did the same thing.

PR #286, opened 2020-04-01
Add pkcs11test as a git submodule

p-steuer:
.. for convenience. To fetch the pkcs11test suite, you can now either
```
 git clone --recurse-submodules <opencryptoki>
```
If you already cloned opencryptoki you can invoke
```
git submodule update --init
```
from the opencryptoki source directory

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "pkcs11test"]
+	path = pkcs11test
+	url = https://github.com/google/pkcs11test


ifranzki:
Sounds OK for me.
Would this mean anything for the distributors? Would they need to handle building/packaging different compared to how it was before?

p-steuer:
I dont think so. submodules are not included in the tarball per default.

p-steuer:
Does not make sense without our changes wich are not ipstream yet:
https://github.com/p-steuer/pkcs11test/tree/changes

Closing this one for now.

PR #290, opened 2020-04-03
configure.ac: add locks status message, and improve configure --help

xnox:
Tested all combinations of:

```
./configure
./configure --enable-locks
./configure --enable-locks=no
./configure --disable-locks
```

Normally, the opposite of default is listed in the `./configure --help` output (i.e. `--disable` for things that are enabled by default), and this variable seems important enough to show, given that default has recently changed.

Also note that prior to v3.13 the locks flag behavior was buggy, and `--disable-locks` actually enabled using locks, instead of transactional memory.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -181,9 +181,7 @@ AC_ARG_ENABLE([pkcsep11_session],
 
 dnl --- locking support
 AC_ARG_ENABLE([locks],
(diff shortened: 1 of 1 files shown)


p-steuer:
Ah now i got it. It was already working in 3.13 but was broke previously. Pre 3.13 the only option to build with TM was to not specify any lock option with configure.

p-steuer:
merged, thanks.

PR #291, opened 2020-04-03
Dilithium stage2

jschmidb:


p-steuer:
travis:
```
./testcases/crypto/dilithium.h:154:12: error: string literal of length 6732 exceeds maximum length 4095 that ISO C99 compilers are required to support [-Werror,-Woverlength-strings]
```

usr/lib/common/asn1.c
```diff
@@ -3473,3 +3574,806 @@ CK_RV ber_decode_DHPublicKey(CK_BYTE *data,
...
+        offset += len;
+        rc |= ber_encode_BIT_STRING(TRUE, NULL, &len, NULL, s1->ulValueLen, 0);
+        offset += len;
+        rc |= ber_encode_BIT_STRING(TRUE, NULL, &len, NULL, s2->ulValueLen, 0);
+        offset += len;
+        rc |= ber_encode_BIT_STRING(TRUE, NULL, &len, NULL, t0->ulValueLen, 0);
+        offset += len;
+        if (t1) {
+            ber = ber_alloc_t(LBER_USE_DER);
+            rc = ber_put_bitstring(ber, t1->pValue, t1->ulValueLen * 8, 0x03);
```
ifranzki, line 3874: Why don't you use ber_encode_BIT_STRING here as well?

usr/lib/common/asn1.c
```diff
@@ -3473,3 +3574,806 @@ CK_RV ber_decode_DHPublicKey(CK_BYTE *data,
...
+            ber_encode_CHOICE(TRUE, 0, NULL, &len, (CK_BYTE *)val->bv_val,
+                              val->bv_len);
+            ber_free(ber, 1);
+
+            offset += len;
+        }
+    }
+
+    if (rc != CKR_OK) {
+        TRACE_DEVEL("ber_encode_INTEGER failed\n");
```
ifranzki, line 3886: The trace text is not entirely true....

usr/lib/common/asn1.c
```diff
@@ -3473,3 +3574,806 @@ CK_RV ber_decode_DHPublicKey(CK_BYTE *data,
...
+        rc |= ber_encode_BIT_STRING(TRUE, NULL, &len, NULL, s2->ulValueLen, 0);
+        offset += len;
+        rc |= ber_encode_BIT_STRING(TRUE, NULL, &len, NULL, t0->ulValueLen, 0);
+        offset += len;
+        if (t1) {
+            ber = ber_alloc_t(LBER_USE_DER);
+            rc = ber_put_bitstring(ber, t1->pValue, t1->ulValueLen * 8, 0x03);
+            rc = ber_flatten(ber, &val);
+
+            ber_encode_CHOICE(TRUE, 0, NULL, &len, (CK_BYTE *)val->bv_val,
```
ifranzki, line 3877: error checking ?

usr/lib/common/asn1.c
```diff
@@ -3473,3 +3574,806 @@ CK_RV ber_decode_DHPublicKey(CK_BYTE *data,
...
+    BerValue *val;
+
+    /* Calculate storage for sequence */
+    offset = 0;
+    rc = 0;
+
+    rc |= ber_encode_INTEGER(TRUE, NULL, &len, NULL, sizeof(version));
+    offset += len;
+    if (opaque != NULL) {
+        rc |= ber_encode_BIT_STRING(TRUE, NULL, &len, NULL, opaque->ulValueLen, 0);
```
ifranzki, line 3857: Shouldn't this be ber_encode_OCTET_STRING here? Below you encode it as an octet string

usr/lib/common/asn1.c
```diff
@@ -3473,3 +3574,806 @@ CK_RV ber_decode_DHPublicKey(CK_BYTE *data,
...
+        offset += len;
+        free(buf2);
+        buf2 = NULL;
+    }
+
+    /* Check if key given as opaque blob */
+    if (opaque != NULL) {
+        // the CKA_IBM_OPAQUE attrib
+        rc = ber_encode_OCTET_STRING(FALSE, &buf2, &len,
+                                     (CK_BYTE *) opaque +
```
ifranzki, line 3933: you do "attr + sizeof(CK_ATTRIBUTE)" at several places in this function.  Shouldn't this be "attr->pValue" instead? "attr + sizeof(CK_ATTRIBUTE)" only works if the attribute was allocated that way, that the value is directly following the CK_ATTRIBUET structure. I would not assume that this is always the case, you never know where you get the inout attributes from.

usr/lib/common/asn1.c
```diff
@@ -3473,3 +3574,806 @@ CK_RV ber_decode_DHPublicKey(CK_BYTE *data,
...
+            memcpy(buf + offset, buf2, len);
+            offset += len;
+            free(buf2);
+            buf2 = NULL;
+        }
+
+        /* (t1) Optional bit-string of public key */
+        if (t1 && t1->pValue) {
+            ber = ber_alloc_t(LBER_USE_DER);
+            rc = ber_put_bitstring(ber, t1->pValue, t1->ulValueLen * 8, 0x03);
```
ifranzki, line 4037: Why don't you use ber_encode_BIT_STRING here as well?

usr/lib/common/asn1.c
```diff
@@ -3473,3 +3574,806 @@ CK_RV ber_decode_DHPublicKey(CK_BYTE *data,
...
+            rc = ber_flatten(ber, &val);
+
+            ber_encode_CHOICE(FALSE, 0, &buf2, &len, (CK_BYTE *)val->bv_val,
```
ifranzki, line 4040: error checking ?

usr/lib/common/asn1.c
```diff
@@ -3473,3 +3574,806 @@ CK_RV ber_decode_DHPublicKey(CK_BYTE *data,
...
+
+        /* t1 */
+        rc = ber_decode_BIT_STRING(buf + offset, &tmp, &len, &field_len);
+        if (rc != CKR_OK) {
+            TRACE_DEVEL("ber_decode_BIT_STRING of (t1) failed\n");
+            goto cleanup;
+        }
+        offset += field_len;
+
+        /* Check if buffer big enough */
```
ifranzki, line 4207: So you do all this just to check if the individual fields fit into the sequence? 
Wouldn't it be easier to create the attributes right away, and add the check at the end? You free all the attributes already in case of an error. That way you would not need to decode the fields twice.

usr/lib/common/key.c
```diff
@@ -1660,6 +1663,133 @@ CK_RV rsa_priv_unwrap_get_data(TEMPLATE *tmpl,
...
+                    *(CK_ULONG *) keyform->pValue);
+        return CKR_TEMPLATE_INCONSISTENT;
+    }
+
+    /* Check if key given as opaque blob */
+    if (template_attribute_find(tmpl, CKA_IBM_OPAQUE, &opaque) == FALSE) {
+
+        if (template_attribute_find(tmpl, CKA_IBM_DILITHIUM_RHO, &rho) == FALSE) {
+            TRACE_ERROR("Could not find CKA_IBM_DILITHIUM_RHO for the key.\n");
+            return CKR_FUNCTION_FAILED;
```
ifranzki, line 1694: Shouldn't this be CKR_TEMPLATE_INCONSISTENT (also for the following attributes)?

usr/lib/common/key.c
```diff
@@ -1660,6 +1663,133 @@ CK_RV rsa_priv_unwrap_get_data(TEMPLATE *tmpl,
...
+
+    rc = ber_decode_IBM_DilithiumPublicKey(data, total_length, &rho, &t1);
+    if (rc != CKR_OK) {
+        TRACE_DEVEL("ber_decode_DilithiumPublicKey failed\n");
+        return rc;
+    }
+
+    rc = template_update_attribute(tmpl, rho);
+    if (rc != CKR_OK)
+        TRACE_DEVEL("template_update_attribute(CKA_IBM_DILITHIUM_RHO) failed\n");
```
ifranzki, line 1753: Shouldn't it return the return code in case of an error?

usr/lib/common/key.c
```diff
@@ -1660,6 +1663,133 @@ CK_RV rsa_priv_unwrap_get_data(TEMPLATE *tmpl,
...
+    rc = template_update_attribute(tmpl, t1);
+    if (rc != CKR_OK)
+        TRACE_DEVEL("template_update_attribute(CKA_IBM_DILITHIUM_T1) failed\n");
```
ifranzki, line 1756: same here

usr/lib/common/key.c
```diff
@@ -1660,6 +1663,133 @@ CK_RV rsa_priv_unwrap_get_data(TEMPLATE *tmpl,
...
+                                 &opaque, isOpaque);
+    if (rc != CKR_OK) {
+        TRACE_DEVEL("der_decode_IBM_DilithiumPrivateKey failed\n");
+        return rc;
+    }
+
+    if (isOpaque)
+        template_update_attribute(tmpl, opaque);
+
+    template_update_attribute(tmpl, rho);
```
ifranzki, line 1782: Error checking ?  Maybe like rc |= template_update_attribute ... and check rc at the end?

usr/lib/common/key.c
```diff
@@ -3388,13 +3518,14 @@ CK_RV ibm_dilithium_priv_check_required_attributes(TEMPLATE *tmpl, CK_ULONG mode
...
     /* MODE_KEYGEN: attrs are added during keygen */
-    if (mode == MODE_KEYGEN)
+    if (mode == MODE_KEYGEN || mode == MODE_UNWRAP)
         return priv_key_check_required_attributes(tmpl, mode);
 
     /* MODE_CREATE (key import) or MODE_COPY: check if all attrs present */
     for (i = 0; i < sizeof(req_attrs) / sizeof(req_attrs[0]); i++) {
         if (!(template_attribute_find(tmpl, req_attrs[i], &attr))) {
-            TRACE_ERROR("%s\n", ock_err(ERR_TEMPLATE_INCOMPLETE));
+            TRACE_ERROR("%s, attribute at index %ld missing.\n",
```
ifranzki, line 3527: Better trace req_attr[i].type instead of the index.

testcases/crypto/dilithium_func.c
```diff
@@ -48,21 +48,18 @@ CK_RV run_SignVerifyDilithium(CK_SESSION_HANDLE session,
...
                               CK_OBJECT_HANDLE publ_key)
 {
-    CK_MECHANISM mech2;
+    CK_MECHANISM mech;
     CK_BYTE_PTR data = NULL, signature = NULL;
     CK_ULONG i, signaturelen;
     CK_MECHANISM_INFO mech_info;
     CK_RV rc;
 
-    testcase_begin("Starting with mechtype='%s', inputlen=%lu",
```
ifranzki, line 57: Why is the testcase_begin removed?

testcases/crypto/dilithium_func.c
```diff
@@ -240,6 +321,394 @@ CK_RV run_GenerateDilithiumKeyPairSignVerify()
...
+    testcase_user_logout();
+    testcase_close_session();
+
+    return rc;
+}
+
+/**
+ * Creates an AES-256 wrapping key.
+ */
+CK_RV createWrappingKey(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE *secret_key)
```
ifranzki, line 432: Why not use generate_AESKey from testcases/common/common.c ?

testcases/crypto/dilithium_func.c
```diff
@@ -240,6 +321,394 @@ CK_RV run_GenerateDilithiumKeyPairSignVerify()
...
+    if (rc != CKR_OK)
+        goto done;
+
+    *wrapped_key = tmp_key;
+    *wrapped_keylen = tmp_len;
+
+    rc = CKR_OK;
+
+done:
+
```
ifranzki, line 496: free tmp_key if rc != CKR_OK ?

usr/lib/common/asn1.c
```diff
@@ -3473,3 +3574,806 @@ CK_RV ber_decode_DHPublicKey(CK_BYTE *data,
...
+    /**
+     *    SEQUENCE (2 elem)
+     *       BIT STRING -> rho
+     *       BIT STRING -> t
+     */
+    offset = 0;
+    rc = ber_encode_BIT_STRING(FALSE, &buf2, &len,
+                            rho->pValue, rho->ulValueLen, 0);
+    if (rc != CKR_OK) {
+        TRACE_DEVEL("%s ber_encode_Int failed with rc=0x%lx\n", __func__, rc);
```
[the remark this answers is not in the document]
jschmidb: Right. 
buf2 is the buffer allocated in ber_encode_BIT_STRING and this one is in fact freed.
buf is the buffer where the key parts are assembled and must be freed at the end of the routine.
Same with buf3 below. 
By the way: this is also incorrect in ber_encode_RSAPublicKey.

usr/lib/common/asn1.c
```diff
@@ -3473,3 +3574,806 @@ CK_RV ber_decode_DHPublicKey(CK_BYTE *data,
...
+        return rc;
+    }
+    memcpy(buf + offset, buf2, len);
+    offset += len;
+    free(buf2);
+
+    rc = ber_encode_BIT_STRING(FALSE, &buf2, &len,
+                            t1->pValue, t1->ulValueLen, 0);
+    if (rc != CKR_OK) {
+        TRACE_DEVEL("%s ber_encode_Int failed with rc=0x%lx\n", __func__, rc);
```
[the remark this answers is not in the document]
jschmidb: see above.

usr/lib/common/asn1.c
```diff
@@ -3473,3 +3574,806 @@ CK_RV ber_decode_DHPublicKey(CK_BYTE *data,
...
+        return rc;
+    }
+    memcpy(buf + offset, buf2, len);
+    offset += len;
+    free(buf2);
+
+    rc = ber_encode_SEQUENCE(FALSE, &buf2, &len, buf, offset);
+    if (rc != CKR_OK) {
+        TRACE_DEVEL("%s ber_encode_Seq failed with rc=0x%lx\n", __func__, rc);
```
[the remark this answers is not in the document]
jschmidb: see above.

usr/lib/common/asn1.c
```diff
@@ -3473,3 +3574,806 @@ CK_RV ber_decode_DHPublicKey(CK_BYTE *data,
...
+        return rc;
+    }
+
+    /* Calculate length of outer sequence */
+    rc = ber_encode_OCTET_STRING(TRUE, NULL, &total, buf2, len);
+    if (rc != CKR_OK) {
+        TRACE_DEVEL("%s ber_encode_Oct_Str failed with rc=0x%lx\n", __func__,
```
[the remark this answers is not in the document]
jschmidb: for buf, see above.
buf2 needs not to be freed here, because ber_encode_OCTET_STRING is called to retrieve the length of the sequence. In this case no buf is allocated.

usr/lib/common/asn1.c
```diff
@@ -3473,3 +3574,806 @@ CK_RV ber_decode_DHPublicKey(CK_BYTE *data,
...
+                    rc);
+        return rc;
+    } else {
+        total_len += total + 1;
+    }
+
+    /* Allocate storage for outer sequence and bit string */
+    buf3 = (CK_BYTE *) malloc(total_len);
+    if (!buf3) {
+        TRACE_ERROR("%s Memory allocation failed\n", __func__);
```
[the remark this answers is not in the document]
jschmidb: Yes, buf3 must also be freed at the end of the routine.

usr/lib/common/asn1.c
```diff
@@ -3473,3 +3574,806 @@ CK_RV ber_decode_DHPublicKey(CK_BYTE *data,
...
+     *       OBJECT IDENTIFIER 1.3.6.1.4.1.2.267.1.6.5
+     *       NULL -> no parms for this oid
+     *    BIT STRING (1 elem)
+     *       SEQUENCE (2 elem)
+     *          BIT STRING  -> rho
+     *          BIT STRING  -> t1
+     */
+    rc = ber_encode_SEQUENCE(FALSE, data, data_len, buf3, total_len);
+    if (rc != CKR_OK)
+        TRACE_DEVEL("%s ber_encode_Seq failed with rc=0x%lx\n", __func__, rc);
```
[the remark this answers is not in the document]
jschmidb: see above

usr/lib/common/asn1.c
```diff
@@ -3473,3 +3574,806 @@ CK_RV ber_decode_DHPublicKey(CK_BYTE *data,
...
+    memcpy(buf + offset, buf2, len);
+    offset += len;
+    free(buf2);
+
+    rc = ber_encode_SEQUENCE(FALSE, &buf2, &len, buf, offset);
+    if (rc != CKR_OK) {
+        TRACE_DEVEL("%s ber_encode_Seq failed with rc=0x%lx\n", __func__, rc);
+        return rc;
+    }
+
```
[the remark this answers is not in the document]
jschmidb: That's it.

usr/lib/common/asn1.c
```diff
@@ -3473,3 +3574,806 @@ CK_RV ber_decode_DHPublicKey(CK_BYTE *data,
...
+    offset = 0;
+    rc = 0;
+
+    /* Version */
+    rc = ber_encode_INTEGER(FALSE, &buf2, &len, version, sizeof(version));
+    if (rc != CKR_OK) {
+        TRACE_DEVEL("ber_encode_INTEGER of version failed\n");
+        goto error;
+    }
+    if (buf2 != NULL) {
```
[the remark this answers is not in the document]
jschmidb: It can't! Fixed.

usr/lib/common/asn1.c
jschmidb: ok, changed the text.

usr/lib/common/asn1.c
jschmidb: Yes, to get a rc = 0 only if all operations succeeded, we must use rc |= as above.

usr/lib/common/asn1.c
jschmidb: Good point ... and there are a couple of other code pieces with the same logic.

usr/lib/common/key.c
jschmidb: yes.

usr/lib/common/key.c
jschmidb: yes, agreed. 
As all other ..._priv_unwrap_get_data routines are coded in the same way, we should also fix these. I wonder if someone in the past intentionally introduced this logic ....

usr/lib/common/key.c
jschmidb: yep

usr/lib/common/key.c
jschmidb: Yes, agreed.
Also here we have many routines coded this way, so we should do some overall cleanup too.

usr/lib/common/key.c
jschmidb: Fixed.

testcases/crypto/dilithium_func.c
jschmidb: The testcase_begin macro prints some "----" on the console and I want to group the individual outputs of the different tests (generate, import, transport) including sign/verify within such output "groups".

testcases/crypto/dilithium_func.c
jschmidb: Fixed.

usr/lib/common/asn1.c
jschmidb: Interesting: ber_put_bitstring returns the number of put bits, so a check for rc = 0 fails. I guess this is the reason why it was always coded without rc checking. I will check if we can replace ber_put_bitstring by ber_encode_BIT_STRING, which would also remove this dependency to OpenLDAP.

usr/lib/common/asn1.c
```diff
@@ -411,8 +411,109 @@ CK_RV ber_decode_OCTET_STRING(CK_BYTE *str,
     return CKR_FUNCTION_FAILED;
 }
 
-//
-//
+/**
+ * Here we assume that the indication about unused bits is NOT
+ * part of the data. It will be added by this function.
+ */
+CK_ULONG ber_encode_BIT_STRING(CK_BBOOL length_only,
```
p-steuer, line 418: dont we link some library for normal ber endcoding ?

usr/lib/common/asn1.c
```diff
@@ -411,8 +411,109 @@ CK_RV ber_decode_OCTET_STRING(CK_BYTE *str,
...
+    } else if (data_len + 1 < (1 << 16)) {
+        len = 1 + (1 + 2) + 1 + data_len;
+    } else if (data_len + 1 < (1 << 24)) {
+        len = 1 + (1 + 3) + 1 + data_len;
+    } else {
+        TRACE_ERROR("%s\n", ock_err(ERR_FUNCTION_FAILED));
+        return CKR_FUNCTION_FAILED;
+    }
+    if (length_only == TRUE) {
+        *ber_str_len = len;
```
p-steuer, line 445: I know we are actually not doing this right now, but it think we sould use assertions at function entry to document calling conventions. E.g., this function must not be called with a NULL ber_str_len arg but you have to read the source code up to the first dereference to actually see it.

usr/lib/common/asn1.c
jschmidb: OK, I can replace ber_put_bitstring / ber_flatten by ber_encode_BIT_STRING. So I will do this now for Dilithium, but let's not forget to fix the other places also to get rid of OpenLdap.

usr/lib/common/asn1.c
```diff
@@ -411,8 +411,109 @@ CK_RV ber_decode_OCTET_STRING(CK_BYTE *str,
...
 }
 
-//
-//
+/**
+ * Here we assume that the indication about unused bits is NOT
+ * part of the data. It will be added by this function.
+ */
+CK_ULONG ber_encode_BIT_STRING(CK_BBOOL length_only,
+                            CK_BYTE **ber_str,
```
p-steuer, line 419: do we actually need length_only param ? can we not use the standard "if outbuf == NULL then return outbuflen" ?

usr/lib/common/asn1.c
p-steuer: ber = ber_alloc_t(LBER_USE_DER); above also needs  a check for NULL i guess.

p-steuer: Instead of looking to replace -lber i would rather see if we can replace our own ber stuff by lber or some other ber encoding library.

usr/lib/common/asn1.c
jschmidb: Yes, I think we could. Let's put this on our list of housekeeping todo's. I counted 16 routines in asn1.c that are coded this way.

jschmidb:
I remember a discussion with Reinhard about replacing our own asn.1 encoding/decoding routines by e.g. using OpenSSL. And there might even be more own implementations of any standard functionality in Opencryptoki. We already replaced the local MD5 implementation by OpenSSL. Let's also put this to our todo list.

usr/lib/common/asn1.c
jschmidb: see below ..

usr/lib/common/asn1.c
jschmidb: We have a dependency on OpenLDAP libs lber and ldap, see configure.ac:
> OPENLDAP_LIBS=
> AC_CHECK_HEADERS([lber.h ldap.h],
>                 [OPENLDAP_LIBS="-llber -lldap"],
>                 [AC_MSG_ERROR([lber.h and ldap.h are missing. Please install
>                               'openldap-devel'.])])
> LIBS="$LIBS $OPENLDAP_LIBS"
> AC_SUBST([OPENLDAP_LIBS])

testcases/crypto/dilithium_func.c
jschmidb: Yes, this works. Will replace it. Let's add to our todo list: also clean up this for rsa_func.c and ec_func.c.

usr/lib/common/asn1.c
p-steuer: My point was that this lib probably has some interface  for encoding bit strings so we dont have to write code for that ?

usr/lib/common/key.c
jschmidb: While changing this code, I realized that we use TRACE_DEVEL if any ber_decode_xxx or template_update_attribute routine failed. Shouldn't we better use TRACE_ERROR here? TRACE_DEVEL may be additional output of key parts, but here we really have error situations. What do you think?

usr/lib/common/asn1.c
jschmidb: You mean ber_put_bitstring and ber_flatten that are used at many places in Opencryptoki? In fact they do the same as ber_encode_BIT_STRING, but their usage is more complicated than ber_encode_BIT_STRING. Per Ingos comment I verified that I can replace ber_put_bitstring / ber_flatten by ber_encode_BIT_STRING. After all we can do it either way, but should end up with only one implementation. I think it would be a good idea to look into OpenSSL for this.

p-steuer: okay

usr/lib/common/asn1.c
jschmidb: Good point. I changed this code to using attr->pValue, but also put on our todo list that a couple of other functions (ber_encode_RSAPrivateKey, ber_encode_RSAPublicKey, ber_encode_DSAPrivateKey, ber_encode_ECPrivateKey ...) are coded in the same way.

usr/lib/common/asn1.c
jschmidb: Fixed.

usr/lib/common/asn1.c
```diff
@@ -3473,3 +3574,806 @@ CK_RV ber_decode_DHPublicKey(CK_BYTE *data,
...
+    }
+
+    /*
+     * SEQUENCE (2 elem)
+     *      OBJECT IDENTIFIER 1.3.6.1.4.1.2.267.1.6.5
+     *      NULL  <- no parms for this oid
+     */
+    total_len = 0;
+    memcpy(buf3 + total_len, ber_AlgIdDilithium, ber_AlgIdDilithiumLen);
+    total_len += ber_AlgIdDilithiumLen;
```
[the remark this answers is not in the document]
jschmidb: Yes, because total_len is initialized with
> total_len = ber_AlgIdDilithiumLen;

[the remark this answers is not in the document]
jschmidb: ... I mean, at the beginning of the function before the malloc takes place. 
total_len is set to 0 after the malloc.

usr/lib/common/asn1.c
jschmidb: Fixed. 
Let's remember to also fix ber_decode_RSAPrivateKey, ber_decode_DSAPrivateKey, and ber_decode_DHPrivateKey later.

usr/lib/common/asn1.c
```diff
@@ -3473,3 +3574,806 @@ CK_RV ber_decode_DHPublicKey(CK_BYTE *data,
...
+    free(buf2);
+
+    rc = ber_encode_SEQUENCE(FALSE, &buf2, &len, buf, offset);
+    if (rc != CKR_OK) {
+        TRACE_DEVEL("%s ber_encode_Seq failed with rc=0x%lx\n", __func__, rc);
+        return rc;
+    }
+
+    /* Calculate length of outer sequence */
+    rc = ber_encode_OCTET_STRING(TRUE, NULL, &total, buf2, len);
```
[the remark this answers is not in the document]
jschmidb: Fixed. Should also be corrected in ber_encode_RSAPublicKey.

usr/lib/common/asn1.c
```diff
@@ -3473,3 +3574,806 @@ CK_RV ber_decode_DHPublicKey(CK_BYTE *data,
...
+    if (rc != CKR_OK) {
+        TRACE_DEVEL("%s ber_encode_Oct_Str failed with rc=0x%lx\n", __func__,
+                    rc);
+        return rc;
+    } else {
+        total_len += total + 1;
```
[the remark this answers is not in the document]
jschmidb: Fixed together with replacing ber_encode_OCTET_STRING by ber_encode_BIT_STRING. Should be corrected in ber_encode_RSAPublicKey also.

usr/lib/common/asn1.c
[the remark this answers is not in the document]
jschmidb: Fixed.

usr/lib/common/asn1.c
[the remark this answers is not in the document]
jschmidb: Fixed.

usr/lib/common/asn1.c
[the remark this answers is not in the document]
jschmidb: Fixed.

PR #294, opened 2020-04-08
Avoid mutex lock at bt_put_node_value()

ifranzki:
Function bt_put_node_value() does not really modify the btree, but it decrements the node values reference count.

Do not use the btree's mutex to lock the reference count from being modified concurrently, but use the built-in functions for atomic memory access instead. This allows bt_put_node_value() to run lock-free.

That way, we can save one mutex-lock/unlock cycle per typical BTree access (bt_get_node_value()/bt_put_node_value()).

### DIFF
--- a/usr/include/local_types.h
+++ b/usr/include/local_types.h
@@ -22,7 +22,11 @@ typedef unsigned int uint32;
 
 /* Each node value must start with struct bt_ref_hdr */
 struct bt_ref_hdr {
+#ifdef ENABLE_LOCKS
+    volatile unsigned long ref;
+#else
     unsigned long ref;
+#endif
 };
 
 #define BT_FLAG_FREE 1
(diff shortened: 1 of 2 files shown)


p-steuer:
Is the gcc builtin also supported by clang ?

ifranzki:
> Is the gcc builtin also supported by clang ?
> edit: Ah yes -- i forgot we have clang build in travis ..

At least travis builds successfully with it ....

It might also be available only since a certain gcc (and clang) version. Not sure how to find that out. We might then want to add a check to configure?

p-steuer:
you can ask our toolchain folks how portable this is wrt comiler versions, platforms etc. yes, we should also add a configure.ac check and use the locking behavior as a last resort if the atomics are not available. im afraid if we dont, the distributors might fail to build the opencryptoki package for platforms we dont test.

PR #297, opened 2020-05-12
Added new p11sak tool in opencryptoki/usr/sbin to list and generate keys

reumann-ibm:


ifranzki:
Please check the Travis builds, you have build errors there (click on Details link at continuous-integration/travis-ci/pr):
```
In file included from usr/sbin/p11sak/p11sak.c:15:
./usr/include/pkcs11.h:14:10: fatal error: 'opencryptoki/pkcs11types.h' file not
      found
#include <opencryptoki/pkcs11types.h>
         ^
1 error generated.
```

Note that the first 2 Travis builds will always fail due to Trousers errors. Those are unrelated to your tool and can be ignored. But the remaining builds should all succeed.

jschmidb:
I'm getting a segfault when invoking the tool without parms:

```
# p11sak
Segmentation fault
```

jschmidb:
When I display the generate-key help, also the list option is showed:
```
# p11sak generate-key -h
...
 Options:
      -l, --long                              list output with long format
   ...
```
Is this intended?

jschmidb:
I can create two AES keys with the same label, is this intended?
```
# p11sak list-key secret --slot 4
Please enter user PIN:

 | M  R  L  S  E  D  G  V  W  U  X  A  N |    KEY TYPE | LABEL
 |---------------------------------------+-------------+-------------
 | 0  0  0  0  0  0  0  0  0  0  0  0  0 |     AES 128 | aes128key
 | 0  0  0  0  0  0  0  0  0  0  0  0  0 |     AES 128 | aes128key
```

jschmidb:
When listing keys, a "list-all" arg would be great!

jschmidb:
When using the --long option for listing keys, can you also show additional attributes like key length, EC curve name, value of an RSA exponent, etc.?

ifranzki:
I created an AES key and a 3DES key with the EP11 token. Worked fine. 
When listing the keys, I see that all boolean attributes are FALSE:
```
 | M  R  L  S  E  D  G  V  W  U  X  A  N |    KEY TYPE | LABEL
 |---------------------------------------+-------------+-------------
 | 0  0  0  0  0  0  0  0  0  0  0  0  0 |     AES 256 | TEST
 | 0  0  0  0  0  0  0  0  0  0  0  0  0 |        3DES | TEST
```
A secret key with all attributes FALSE is useless. If I cant encrypt/decrypt, nor sign/verify, nor wrap/unwrap with the key, then its not usable for anything, and thus useless. 

Asym keys seem to have better default attributes, allowing all 3 operation types:
```
 | M  R  L  S  E  D  G  V  W  U  X  A  N |    KEY TYPE | LABEL
 |---------------------------------------+-------------+-------------
 | 1  0  0  0  -  1  1  -  -  1  1  0  0 | private RSA | TEST:prv
 | 1  0  0  0  -  1  1  -  -  1  1  0  0 |  private EC | TEST:prv

 | M  R  L  S  E  D  G  V  W  U  X  A  N |    KEY TYPE | LABEL
 |---------------------------------------+-------------+-------------
 | 1  0  0  -  1  -  -  1  1  -  -  -  - |  public RSA | TEST:pub
 | 1  0  0  -  1  -  -  1  1  -  -  -  - |   public EC | TEST:pub
```

ifranzki:
Specifying an unknown option does not produce an error:
```
./p11sak list-key public --slot 4 --jxyyhv
./p11sak list-key public --slot 4 -x
```
I would expect that at least a warning message about an unrecognized option is printed, preferable it should even fail.

When specifying an invalid command, it shows the help output, but doe snot tell me what is wrong with the command:
`# ./p11sak  dgfsgf`
I would expect an error message that command dgfsgf is unknown or similar.

When I specify only the command, but without key type, I get an error message, but what does the [9] mean ?
```
# ./p11sak list-key
Cipher key type [9] is not set or not supported
```

Specifying an unparsable slot number shows an error message, that the slot number would be missing... Maybe print the real reason of the error?
```
# ./p11sak list-key aes --slot abc
Slot number must be specified.
```
If a parsable, but non-existing slot number is specified it fails later on when opening the session, that's OK.

ifranzki:
> I can create two AES keys with the same label, is this intended?

Yes, labels are not unique in PKCS#11, so thats OK and not the fault of p11sak. 
Actually there no other attribute that is unique either.

usr/sbin/p11sak/p11sak.1
```diff
@@ -0,0 +1,46 @@
+.\" Copyright IBM Corp. 2020
```
ifranzki, line 1: I think the man page belongs into <ock>/man/man1/, not into the tools directory.

usr/sbin/p11sak/p11sak.c
```diff
@@ -8,9 +8,9 @@
  * https://opensource.org/licenses/cpl1.0.php
  */
 
+#include <stdio.h>
 #include <stdlib.h>
```
ifranzki, line 12: Mixing changes in a single commit.....  I would separate this, and sqaush this change with the original commit where you introduce that code

usr/sbin/p11sak/p11sak.1
ifranzki: And also add it to man1.mk so that it gets installed during make install

usr/sbin/p11sak/p11sak.mk
```diff
@@ -1,13 +1,13 @@
-sbin_PROGRAMS += usr/sbin/p11sak_tool/p11sak
-noinst_HEADERS += usr/sbin/p11sak_tool/p11sak.h
+sbin_PROGRAMS += usr/sbin/p11sak/p11sak
```
ifranzki, line 1: I would squash this commit with your very first commit.
Why fixing your own commit later? Just do it right in the first commit ;-)

configure.ac
```diff
@@ -172,7 +172,7 @@ AC_ARG_ENABLE([pkcsep11_migrate],
 	AS_HELP_STRING([--enable-pkcsep11_migrate],[build pkcsep11_migrate (EP11 token key migration tool) @<:@default=enabled if EP11 library is present@:>@]),
 	[],
 	[enable_pkcsep11_migrate=check])
-	
+
```
ifranzki, line 175: Unrelated change, but OK

usr/sbin/p11sak/p11sak.mk
```diff
@@ -9,4 +9,4 @@ usr_sbin_p11sak_p11sak_CFLAGS  =			\
 	-I${srcdir}/usr/sbin/p11sak
 
 usr_sbin_p11sak_p11sak_SOURCES =			\
-	usr/sbin/p11sak_tool/p11sak.c
+	usr/sbin/p11sak/p11sak.c
```
ifranzki, line 12: Squash this change with your very first commit, too

usr/sbin/p11sak/p11sak.c
```diff
@@ -13,12 +13,61 @@
 #include <stdio.h>
 #include <string.h>
 #include <pkcs11.h>
+#include <dlfcn.h>
```
ifranzki, line 16: Squash this commit also with your very first commit. Again: Why fixing your own commit, if you can do it right in the first commit?

reumann-ibm:
> When listing keys, a "list-all" arg would be great!

This will be included in a future release.

reumann-ibm:
> Please check the Travis builds, you have build errors there (click on Details link at continuous-integration/travis-ci/pr):
> 
> ```
> In file included from usr/sbin/p11sak/p11sak.c:15:
> ./usr/include/pkcs11.h:14:10: fatal error: 'opencryptoki/pkcs11types.h' file not
>       found
> #include <opencryptoki/pkcs11types.h>
>          ^
> 1 error generated.
> ```
> 
> Note that the first 2 Travis builds will always fail due to Trousers errors. Those are unrelated to your tool and can be ignored. But the remaining builds should all succeed.

error resolved by replacing pkcs11.h with pkcs11types.h

reumann-ibm:
> Your gits look strange in 'git log':
> 
> ```
> Author: reumann-ibm <<E2><80><9C>matthias.reumann1@ibm.com<E2><80><9D>>
> Date:   Tue May 12 07:33:47 2020 +0200
> 
>     Added initial man pages for p11sak
> ```
> 
> Check you git config and fix your email address. Should be in the form of 'Joerg Schmidbauer [jschmidb@de.ibm.com](mailto:jschmidb@de.ibm.com)' - of course using your own name :-)
> 
> Usually a git commit not only has a title, but also a short description what the commit is doing. Sometimes the title is enough, but its quite useful to have a little longer description about your change. Check the git log to see how other commits are looking like.
> 
> Also: **You need to Sign-off your commits**. We can't take commits without a Sign-off.
> Add 'Signed-off-by: Joerg Schmidbauer [jschmidb@de.ibm.com](mailto:jschmidb@de.ibm.com)' (using your own name and email address) to the commit. 'git commit -s' will automatically add your sign-off to the commits (if your git config is correct).

This is resolved now as recommended.

reumann-ibm:
> I'm getting a segfault when invoking the tool without parms:
> 
> ```
> # p11sak
> Segmentation fault
> ```

This bug is fixed.

reumann-ibm:
> When I display the generate-key help, also the list option is showed:
> 
> ```
> # p11sak generate-key -h
> ...
>  Options:
>       -l, --long                              list output with long format
>    ...
> ```
> 
> Is this intended?

Not intended. Should be removed now.

reumann-ibm:
> I created an AES key and a 3DES key with the EP11 token. Worked fine.
> When listing the keys, I see that all boolean attributes are FALSE:
> 
> ```
>  | M  R  L  S  E  D  G  V  W  U  X  A  N |    KEY TYPE | LABEL
>  |---------------------------------------+-------------+-------------
>  | 0  0  0  0  0  0  0  0  0  0  0  0  0 |     AES 256 | TEST
>  | 0  0  0  0  0  0  0  0  0  0  0  0  0 |        3DES | TEST
> ```
> 
> A secret key with all attributes FALSE is useless. If I cant encrypt/decrypt, nor sign/verify, nor wrap/unwrap with the key, then its not usable for anything, and thus useless.
> 
> Asym keys seem to have better default attributes, allowing all 3 operation types:
> 
> ```
>  | M  R  L  S  E  D  G  V  W  U  X  A  N |    KEY TYPE | LABEL
>  |---------------------------------------+-------------+-------------
>  | 1  0  0  0  -  1  1  -  -  1  1  0  0 | private RSA | TEST:prv
>  | 1  0  0  0  -  1  1  -  -  1  1  0  0 |  private EC | TEST:prv
> 
>  | M  R  L  S  E  D  G  V  W  U  X  A  N |    KEY TYPE | LABEL
>  |---------------------------------------+-------------+-------------
>  | 1  0  0  -  1  -  -  1  1  -  -  -  - |  public RSA | TEST:pub
>  | 1  0  0  -  1  -  -  1  1  -  -  -  - |   public EC | TEST:pub
> ```

I think right now it required a longer discussion what the default values should be. I think as it is currently, the user can specify the attributes and needs to decide which attributes to set to TRUE.

reumann-ibm:
> Specifying an unknown option does not produce an error:
> 
> ```
> ./p11sak list-key public --slot 4 --jxyyhv
> ./p11sak list-key public --slot 4 -x
> ```
> 
> I would expect that at least a warning message about an unrecognized option is printed, preferable it should even fail.
> 
> When specifying an invalid command, it shows the help output, but doe snot tell me what is wrong with the command:
> `# ./p11sak dgfsgf`
> I would expect an error message that command dgfsgf is unknown or similar.
> 
> When I specify only the command, but without key type, I get an error message, but what does the [9] mean ?
> 
> ```
> # ./p11sak list-key
> Cipher key type [9] is not set or not supported
> ```
> 
> Specifying an unparsable slot number shows an error message, that the slot number would be missing... Maybe print the real reason of the error?
> 
> ```
> # ./p11sak list-key aes --slot abc
> Slot number must be specified.
> ```
> 
> If a parsable, but non-existing slot number is specified it fails later on when opening the session, that's OK.

To resolve this, the code to parse the command line arguments was reworked. Arguments not recoginised should return an error message that the argument was not recognized. Otherwise it should print help.

usr/sbin/p11sak/p11sak.1
reumann-ibm: this is resolved.

usr/sbin/p11sak/p11sak.c
reumann-ibm: Done.

usr/sbin/p11sak/p11sak.mk
reumann-ibm: done

usr/sbin/p11sak/p11sak.c
reumann-ibm: done

reumann-ibm:
> When using the --long option for listing keys, can you also show additional attributes like key length, EC curve name, value of an RSA exponent, etc.?

This will be implemented for the future release.

ifranzki:
> This is resolved now as recommended.

The commit Author still shows 
reumann-ibm <“matthias.reumann1@ibm.com”>

ifranzki:
> > This is resolved now as recommended.
> 
> The commit Author still shows
> reumann-ibm <“[matthias.reumann1@ibm.com](mailto:matthias.reumann1@ibm.com)”>

Probably because the commit was created when your git config was still wrong. When you rebase or amend it keeps the author. Not sure if this can be changed afterwads... @hfreude ?

ifranzki:
During make:

usr/sbin/p11sak/p11sak.c: In function 'get_pin':
usr/sbin/p11sak/p11sak.c:90:13: warning: implicit declaration of function 'getline'; did you mean 'get_pin'? [-Wimplicit-function-declaration]
     nread = getline(&buff, &buflen, stdin);
             ^~~~~~~
             get_pin

Possibly `#define _GNU_SOURCE` missing in before `#include <stdio.h>`

ifranzki:
there is an error with your man page after you do 'make clean', it does not find the man page afterwards.

Make clean runs cleanup.sh which does a `rm -rf `cat .gitignore`;` . So maybe we need to generate .gitignore freshly...

reumann-ibm:
> `#define _GNU_SOURCE`

#define _GNU_SOURCE resolves the warning

ifranzki:
Minor thing:
```
# ./p11sak  generate-key aes 256 --label TEST2 --attr ED --slot 4
slot 4Please enter user PIN:
Generate symmetric key AES with keylen=256 and label=[TEST2]
Symmetric key generation successful!
```
There is a "slot 4" printed in front of 'Please enter user PIN:'

reumann-ibm:
> Minor thing:
> 
> ```
> # ./p11sak  generate-key aes 256 --label TEST2 --attr ED --slot 4
> slot 4Please enter user PIN:
> Generate symmetric key AES with keylen=256 and label=[TEST2]
> Symmetric key generation successful!
> ```
> 
> There is a "slot 4" printed in front of 'Please enter user PIN:'

done

configure.ac
hfreude: thats mine ... I always press F3 (whitespace cleanup) when saving a file ...

ifranzki:
> there is an error with your man page after you do 'make clean', it does not find the man page afterwards.
> 
> Make clean runs cleanup.sh which does a rm -rf `cat .gitignore`; . So maybe we need to generate .gitignore freshly...

I guess `p11sak.1` should be named `p11sak.1.in` like the other man pages in that directory.
This should fix the make clean problem.

hfreude:
pr commited, this will be part of ock 3.14

PR #301, opened 2020-06-08
Fix segfault when non-existing token object is deleted

ifranzki:
When a C_DestroyObject tries to delete a token object, that does no longer exist in the token directory and the shm, because another process has already deleted the object, and there are no further
token objects present in the token directory and shm, then function object_mgr_search_shm_for_obj() is called with an incorrect upper boundary of 0xffffffff (-1), which is higher than MAX_TOK_OBJS and
thus the for loop runs out of the array bounds causing a segfault.

### DIFF
--- a/usr/lib/common/obj_mgr.c
+++ b/usr/lib/common/obj_mgr.c
@@ -1751,6 +1751,12 @@ CK_RV object_mgr_set_attribute_values(STDLL_TokData_t *tokdata,
         save_token_object(tokdata, obj);
 
         if (priv_obj) {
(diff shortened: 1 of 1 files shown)


p-steuer:
Id document the restriction in a comment above object_mgr_search_shm_for_obj.

Looks good otherwise.

ifranzki:
> Id document the restriction in a comment above object_mgr_search_shm_for_obj.

Which restriction ?

p-steuer:
misunderstanding on my side. its okay as it is.

PR #302, opened 2020-06-08
Fix DH problems with soft token

ifranzki:


ifranzki:
Yes, based on your comment in the BZ we then do not need the 2nd and 3rd commit.

Without the 3rd commit, the DH testcase does currently allow the attributes to be 1 byte shorter than 'expected'. This might work for most of the times, but it could also happen that a value is 2 bytes shorter, or even more. 

Should we fix this and basically remove any length checks n the DH testcase, or would you prefer to keep it as is?

p-steuer:
Yes, the 1st commit is still needed. Regarding the 3rd commit, one has to remove the lenght checks on the public key completely. The lenght checks for the public key can either be removed or changed to bit level at the same time setting the CKA_VALUE_BITS attrribute in the priv. key template.

Note that i asked for clarification of the general "may big ints be padded?"-thing on the pkcs11 mailing list and also, in case the answer is no, if implementors are encouraged to handle such values anyway for robustness reason or to reject them ("template inconsistent"). Si id wait for an anwer to the first question and opinions on the second before merging any changes.

ifranzki:
I have changed the 3rd commit to remove the key size checks completely. 

I have left the 2nd commit included for now until you got the answers from the pkcs11 mailing list.

ifranzki:
Any news about his? Do you still expect to get any feedback from the PKCS11 mailing list? If not, how should be proceed? Should we just go without the 2nd commit and maybe come back to this at a later time when it has been clarified?

p-steuer:
No.. i dont expect to get an official statement. pkcs 11 seems underpsecified here and any specification could cause problems for vendor implementations. I did not get any opinions in favor of requiring 0 padding. Thus my preferred solution would be that OCK does not output any zero padding itself but can handle zero padded inputs. I dont know how much work is required to implement that tough. If its too much to "just do it", id go with the 2nd commit for now ..

ifranzki:
> Thus my preferred solution would be that OCK does not output any zero padding itself but can handle zero padded inputs.

Well, any code that uses **BN_bn2bin** should already do it the preferred way as far as I can tell. BN_bn2bin only outputs as many bytes that are needed to represent the number. It does not add any padding bytes. 

Also for the opposite direction, **BN_bin2bn** should already handle it correctly, allowing padded input, but does not require it to be padded.  So we should be good here already. 

So for the SOFT token, this would mean that we can skip the 2nd commit, and thus have the preferred behavior. Not sure if there are other places that output zero padded numbers though. Will have to look for them....

p-steuer (approved):

PR #303, opened 2020-06-17
Fixed upper bound buffer length in tok_key_get_key_type() and add remove-key command

reumann-ibm:


usr/sbin/p11sak/p11sak.c
```diff
@@ -1356,6 +1365,8 @@ static CK_RV tok_key_get_key_type(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE hk
                 ret);
         return ret;
     }
+
+
```
ifranzki, line 1369: did you add these 2 blank lines by intention? 1 blank line should be sufficient.....

reumann-ibm: No intention. Removed one blank line.

usr/sbin/p11sak/p11sak.c
```diff
@@ -1356,6 +1368,7 @@ static CK_RV tok_key_get_key_type(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE hk
                 ret);
         return ret;
```
p-steuer: I think buffer also needs to be free'd here (probably at all error exits). Think about providing an error exit label for this function.

usr/sbin/p11sak/p11sak.c
```diff
@@ -844,13 +865,13 @@ static CK_RV tok_key_gen(CK_SESSION_HANDLE session, CK_ULONG keylength,
...
-                    CKA_VERIFY, &a_verify, bs }, { CKA_WRAP, &a_wrap, bs }, {
-                    CKA_UNWRAP, &a_unwrap, bs }, { CKA_EXTRACTABLE,
-                    &a_extractable, bs },
+            CKA_MODIFIABLE, &a_modifiable, bs }, { CKA_DERIVE, &a_derive, bs },
+            { CKA_SENSITIVE, &a_sensitive, bs }, {
+            CKA_ENCRYPT, &a_encrypt, bs }, { CKA_DECRYPT, &a_decrypt, bs }, {
+                    CKA_SIGN, &a_sign, bs }, {
+            CKA_VERIFY, &a_verify, bs }, { CKA_WRAP, &a_wrap, bs }, {
+            CKA_UNWRAP, &a_unwrap, bs },
+            { CKA_EXTRACTABLE, &a_extractable, bs },
```
p-steuer, line 874: There a few chunks in this commit for which its hard to tell if you actually changed something. E.g., this seems to be formatting only. If you think the code needs re-formatting, please make it a separate commit next time.

p-steuer: (The same applies to renaming variables or, more general, all changes that are unrelated to the commit message)

usr/sbin/p11sak/p11sak.c
```diff
@@ -1910,104 +2051,264 @@ static CK_RV list_ckey(CK_SESSION_HANDLE session, p11sak_kt kt, int long_print)
...
+
+    rc = tok_key_list_init(session, kt, label);
+    if (rc != CKR_OK) {
+        printf("Init token key list failed (error code 0x%lX: %s)\n", rc,
+                p11_get_ckr(rc));
+        return rc;
+    }
+
+    while (1) {
+        rc = funcs->C_FindObjects(session, &hkey, 1, &count);
```
ifranzki, line 2092: You only get 1 find result per loop iteration. It works that way, but it would perform much better if you get multiple (say 10 or 20) results with one call to C_FindObjects and the have a secondary loop to iterate over the results (0 .. count -1).

usr/sbin/p11sak/p11sak.c
```diff
@@ -1910,104 +2051,264 @@ static CK_RV list_ckey(CK_SESSION_HANDLE session, p11sak_kt kt, int long_print)
...
+                    printf("Are you sure to delete %s? (y/n)\n", label);
+
+                    nread = getline(&buff, &buflen, stdin);
+                    if (nread == -1) {
+                        rc = -1;
+                        goto done;
+                    }
+                    buff[nread - 1] = '\0';
+
+                    if (strcmp(buff, "y") == 0) {
```
ifranzki, line 2177: Some lines seem to be a duplicate from above.  Maybe this can be refactored to use a subroutine for the common parts?

usr/sbin/p11sak/p11sak.c
```diff
@@ -1910,104 +2051,264 @@ static CK_RV list_ckey(CK_SESSION_HANDLE session, p11sak_kt kt, int long_print)
...
+                            "Key with label %s could not be destroyed (error code 0x%lX: %s)\n",
+                            label, rc, p11_get_ckr(rc));
+                    return rc;
+                }
+                printf("DONE - Destroy Object with Label: %s\n", label);
+            } else {
+                while (1) {
+                    printf("Are you sure to delete %s? (y/n)\n", label);
+
+                    nread = getline(&buff, &buflen, stdin);
```
ifranzki, line 2170: I guess you are intentionally not freeing buff within the loop iteration, so that it is still allocated from the previous iteration here, and is reused/reallocted by getline ?

usr/sbin/p11sak/p11sak.c
reumann-ibm: I have introduced a couple of subroutines to refactore the code as suggested.

usr/sbin/p11sak/p11sak.c
reumann-ibm: Ok. Will do. Thanks for the suggestion.

usr/sbin/p11sak/p11sak.c
reumann-ibm: Fixed. buffer and label need to be freed in the while(1) loop.

usr/sbin/p11sak/p11sak.c
reumann-ibm: This will become an item for future implementation.

reumann-ibm:
The comments and suggestions are worked into the code and the commit has been amended.

usr/sbin/p11sak/p11sak.c
```diff
@@ -1908,106 +2053,277 @@ static CK_RV list_ckey(CK_SESSION_HANDLE session, p11sak_kt kt, int long_print)
...
+static CK_RV getKeyboardInput(char *buff, char *label)
+{
+
+    CK_RV rc = CKR_OK;
+    int nread;
+    size_t buflen;
+
+    printf("Are you sure to delete %s? (y/n)\n", label);
+
+    nread = getline(&buff, &buflen, stdin);
```
ifranzki, line 2083: I don't think that this works as intended. 
getline() allocates  buff, if buff is NULL. If buff is not NULL, it checks if buflen is large enough to hold the input. If it is not large enough then it will reallocate the buffer.
You do allocate buff in the calling routine, but only in 1 byte length:
`if (!(buff = malloc(strlen("y"))))`
Thats wrong in 3 places:
1. strlen("y") returns 1, but you also need space for the terminating zero.
2. buflen is not initialized, so getline will get garbage as buffer length.
3. What if the user input is larger than one character? Then getline will reallocate buff, but you don't pass it back to the caller.

usr/sbin/p11sak/p11sak.c
```diff
@@ -1908,106 +2053,277 @@ static CK_RV list_ckey(CK_SESSION_HANDLE session, p11sak_kt kt, int long_print)
...
+static CK_RV deleteKey(CK_SESSION_HANDLE session, p11sak_kt kt, char *rm_label,
+                       CK_BBOOL *forceAll)
+{
+    CK_ULONG keylength, count;
+    CK_OBJECT_CLASS keyclass;
+    CK_OBJECT_HANDLE hkey;
+    char *keytype = NULL;
+    char *label = NULL;
+    CK_RV rc;
+    char *buff = NULL;
```
ifranzki, line 2128: I would not call this 'buff', name it 'user_input', or anything that tells the reader what it is used for.

usr/sbin/p11sak/p11sak.c
```diff
@@ -1908,106 +2053,277 @@ static CK_RV list_ckey(CK_SESSION_HANDLE session, p11sak_kt kt, int long_print)
...
+                rc = funcs->C_DestroyObject(session, hkey);
+                if (rc != CKR_OK) {
+                    printf(
+                            "Key with label %s could not be destroyed (error code 0x%lX: %s)\n",
+                            label, rc, p11_get_ckr(rc));
+                    return rc;
+                }
+                printf("DONE - Destroy Object with Label: %s\n", label);
+            } else {
+                while (1) {
```
ifranzki, line 2199: what is this while loop for? You exit the loop either by return or by break in any case after one iteration.

usr/sbin/p11sak/p11sak.c
```diff
@@ -1908,106 +2053,277 @@ static CK_RV list_ckey(CK_SESSION_HANDLE session, p11sak_kt kt, int long_print)
...
+                    rc = getKeyboardInput(buff, label);
+                    if (rc != CKR_OK) {
+                        goto done;
+                    }
+
+                    rc = finalizeDestroyObject(buff, label, &session, &hkey);
+                    if (rc != CKR_OK) {
+                        return rc;
+                    }
+                    else {
```
ifranzki, line 2209: else on the same line as the closing bracket

usr/sbin/p11sak/p11sak.c
```diff
@@ -113,20 +114,57 @@ static CK_RV get_pin(char **pin, size_t *pinlen) {
...
+{
+    int nread;
+    char *keyboard_input = NULL;
+    size_t buflen;
+    CK_RV rc = 0;
+
+    /* read the keyboard_input
+     * Note: getline will allocate memory for keyboard_input. free it when done.
+     */
+    nread = getline(&keyboard_input, &buflen, stdin);
```
ifranzki, line 138: Why not directly pass user_input to getline? 

*user_input  = NULL;
nread = getline(user_input , &buflen, stdin);

Then you do not need to allocate a new buffer and copy the content.

usr/sbin/p11sak/p11sak.c
```diff
@@ -113,20 +114,57 @@ static CK_RV get_pin(char **pin, size_t *pinlen) {
...
+    if (nread == -1) {
+        rc = -1;
+        goto done;
+    }
+
+    /* Allocate keyboard_input.
```
ifranzki, line 144: See above.
If you really need to duplicate the buffer, why not simply use strdup() ?

The buffer allocated by getline is supposed to be zero terminated, so stdup should just work fine.

ifranzki:
Travis build errors in https://travis-ci.org/github/opencryptoki/opencryptoki/builds/721927834:

In file included from /usr/include/string.h:494:0,
                 from usr/sbin/p11sak/p11sak.c:16:
In function ‘strcpy’,
    **inlined from ‘deleteKey’ at usr/sbin/p11sak/p11sak.c:2252:21,
    inlined from ‘execute_cmd’ at usr/sbin/p11sak/p11sak.c:2332:12,
    inlined from ‘main’ at usr/sbin/p11sak/p11sak.c:2481:8:**
/usr/include/s390x-linux-gnu/bits/string_fortified.h:90:10: error: ‘__builtin___memcpy_chk’ writing 2 bytes into a region of size 1 overflows the destination [-Werror=stringop-overflow=]
   return __builtin___strcpy_chk (__dest, __src, __bos (__dest));
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

usr/sbin/p11sak/p11sak.c:1858:17: error: equality comparison with extraneous parentheses [-Werror,-Wparentheses-equality]
    if ((*label == NULL)) {
         ~~~~~~~^~~~~~~
usr/sbin/p11sak/p11sak.c:1858:17: note: remove extraneous parentheses around the comparison to silence this warning
    if ((*label == NULL)) {
        ~       ^      ~
usr/sbin/p11sak/p11sak.c:1858:17: note: use '=' to turn this equality comparison into an assignment
    if ((*label == NULL)) {
                ^~
                =

usr/sbin/p11sak/p11sak.c
reumann-ibm: Changed as commented.

ifranzki:
You now have 2 commits named "Added new command remove-key to the p11sak tool to remove key objects… " in the PR. Is this on intention? Shouldn't these 2 be squashed together?

usr/sbin/p11sak/p11sak.c
p-steuer: buffer still needs to be freed on below switch statement's error return.

usr/sbin/p11sak/p11sak.c
p-steuer: If your intention was to only modify the user_input arg in case of successful exit, you could just set user_input = keyboard_input in that case, instead of allocating and copying to a new buffer. Caller is responsible to free user_input any way.

usr/sbin/p11sak/p11sak.c
p-steuer: There is still a lot of formatting/renaming included in this commit. Thats okay for now, but please take care to put such changes in an extra commit (if its more than just a few lines) in future contributions.

usr/sbin/p11sak/p11sak.c
```diff
@@ -1652,146 +1783,231 @@ static CK_RV parse_gen_key_args(char *argv[], int argc, p11sak_kt *kt,
...
+        } else if (strcmp(argv[i], "RSA") == 0 || strcmp(argv[i], "rsa") == 0) {
+            *kt = kt_RSAPKCS;
+        } else if (strcmp(argv[i], "EC") == 0 || strcmp(argv[i], "ec") == 0) {
+            *kt = kt_EC;
+            /* Get options */
+        } else if (strcmp(argv[i], "--slot") == 0) {
+            *slot = (CK_ULONG) atol(argv[i + 1]);
+            i++;
+        } else if (strcmp(argv[i], "--pin") == 0) {
+            *pin = argv[i + 1];
```
p-steuer, line 1835: I think you have to check for an out of bounds read here: If "--poin" is the last string in argv, argv[i+1] becomes a NULL pointer deref.

usr/sbin/p11sak/p11sak.c
```diff
@@ -1652,146 +1783,231 @@ static CK_RV parse_gen_key_args(char *argv[], int argc, p11sak_kt *kt,
...
+            i++;
+        } else if (strcmp(argv[i], "--label") == 0) {
+            *label = argv[i + 1];
```
p-steuer, line 1838: Same here.

usr/sbin/p11sak/p11sak.c
```diff
@@ -1908,109 +2129,266 @@ static CK_RV list_ckey(CK_SESSION_HANDLE session, p11sak_kt kt, int long_print)
...
-    free(keytype);
-    return ret;
+    return rc;
 }
 
-static CK_RV execute_cmd(CK_SESSION_HANDLE session, CK_SLOT_ID slot, p11sak_cmd cmd,
-        p11sak_kt kt, CK_ULONG keylength, CK_ULONG exponent, char *ECcurve,
-        char *label, char *attr_string, int long_print) {
-    CK_RV ret;
+static CK_RV confirmDestroy(char **user_input, char* label)
```
p-steuer, line 2150: We usually dont use camel case for internal functions, do we ?

usr/sbin/p11sak/p11sak.c
```diff
@@ -1908,109 +2129,266 @@ static CK_RV list_ckey(CK_SESSION_HANDLE session, p11sak_kt kt, int long_print)
...
+                }
+                printf("DONE - Destroy Object with Label: %s\n", label);
+            }
+        } else {
+            if ((strcmp(rm_label, "") == 0) || (strcmp(rm_label, label) == 0)) {
+                rc = confirmDestroy(&user_input, label);
+                if (rc != CKR_OK) {
+                    printf("User input failed (error code 0x%lX: %s)\n",
+                            rc, p11_get_ckr(rc));
+                    return rc;
```
p-steuer, line 2263: free(user_input) ?

usr/sbin/p11sak/p11sak.c
```diff
@@ -1908,109 +2129,266 @@ static CK_RV list_ckey(CK_SESSION_HANDLE session, p11sak_kt kt, int long_print)
...
+                }
+
+                rc = finalizeDestroyObject(user_input, label, &session, &hkey);
+                if (rc != CKR_OK) {
+                    return rc;
```
p-steuer, line 2268: free(user_input) ?

usr/sbin/p11sak/p11sak.c
```diff
@@ -1908,109 +2129,266 @@ static CK_RV list_ckey(CK_SESSION_HANDLE session, p11sak_kt kt, int long_print)
...
+    rc = funcs->C_FindObjectsFinal(session);
+    if (rc != CKR_OK) {
+        printf("C_FindObjectsFinal failed (error code 0x%lX: %s)\n", rc,
+                p11_get_ckr(rc));
+        goto done;
+    }
+
+    rc = CKR_OK;
+
+done:
```
p-steuer, line 2287: Why not use this exit label for more error exits to free buffers etc ?

p-steuer (changes requested):
Done with the review.

usr/sbin/p11sak/p11sak.c
reumann-ibm: Made use of the exit label to free buffers.

usr/sbin/p11sak/p11sak.c
reumann-ibm: Reworked user_input into finialize_destroy_object()

usr/sbin/p11sak/p11sak.c
reumann-ibm: Reworked user_input into finialize_destroy_object()

usr/sbin/p11sak/p11sak.c
reumann-ibm: Changed internal function names.

usr/sbin/p11sak/p11sak.c
reumann-ibm: Included out of bounds check.

usr/sbin/p11sak/p11sak.c
reumann-ibm: Included out of bounds check.

usr/sbin/p11sak/p11sak.c
reumann-ibm: Simplified user input into confirm_destroy()

usr/sbin/p11sak/p11sak.c
reumann-ibm: Reworked user_input to remove allocating new buffers & content copy.

p-steuer (approved):
looks good

PR #308, opened 2020-08-13
EP11: Filter mechanisms using a positive list, instead of a negative list

ifranzki:
Change the code from using a negative list (banned mechanisms) to using a positive list (supported mechanisms) instead, to filter out unknown or unsupported mechanisms. Now, only those mechanisms are reported that are supported (i.e. known) by the EP11 token code, and by the configured set of APQNs.

This has the benefit that new mechanisms added by future EP11 crypto card enhancements are not automatically reported as supported by the EP11 token, although it is not clear if the EP11 token code really supports that mechanism. With the negative list, such new mechanisms that are not on the negative list because they have been unknown so far, would be reported as supported. With the new positive list approach, these mechanisms remain unsupported, until they are explicitly added to the list of supported mechanisms.

This also helps to solve problems with mixed card version environments, where cards of different versions are in the set of configured APQNs of the EP11 token. Here, a newer card may support a new, so far unknown mechanism, but older cards do not support it. For mechanisms known by the EP11 token, it is verified that the mechanism is supported by all APQNs, otherwise the
mechanism is not reported via C_GetMechanismList. For unknown mechanisms this can not be checked. So it can happen that a mechanism is reported that is only supported by one APQN, but not by the others. This may lead to errors, when that mechanism is used, or even when C_GetMechanismInfo is used to obtain information about the mechanism (as done by pkcsconf -m). Inconsistencies can also happen when C_GetMechanismList is called several times (e.g. once to get the number of mechanisms, and a second time to obtain the list). With the positive list approach, unknown mechanisms will never be reported, so these problems can not happen.


ifranzki:
> Can we figure out a way - an option or similar - to provide also mechanisms which are not backed up by all the APQNs within a configuration ?

To support this, I would need to remember those APQNs out of the configured set of APQNs that support those mechanism (i.e. kind of tag the mechanisms with the APQNs that support it), and when executing the mechanism I would need to select only those APQNs for execution. 

That's doable, but requires a certain amount effort to maintain the list of APQNs per mechanism. Also, we have hot-plug capability of APQNs, so how to handle APQNs the come and go in regards of the tagging to specific mechanisms.

Also, for  a user point of view, I don't think that such a behavior would be expected. The user configures a set of APQNs to be used for the token, so he assumes fail over within those APQNs. Using only certain APQNs for certain mechanisms undergoes this fail over capability, and may lead to  a situation where the user thinks that he has configured it in a save way for fail over if one APQN breaks, but in fact this is not true for some mechanisms. So IF we ever do something like this, then we should have a config option that needs to be specified to allow such behavior explicitly.

hfreude:
On 17.08.20 13:01, Ingo Franzki wrote:
>> Can we figure out a way - an option or similar - to provide also mechanisms which are not backed up by all the APQNs within a configuration ?
> To support this, I would need to remember those APQNs out of the configured set of APQNs that support those mechanism (i.e. kind of tag the mechanisms with the APQNs that support it), and when executing the mechanism I would need to select only those APQNs for execution. 
>
> That's doable, but requires a certain amount effort to maintain the list of APQNs per mechanism. Also, we have hot-plug capability of APQNs, so how to handle APQNs the come and go in regards of the tagging to specific mechanisms.
>
> Also, for  a user point of view, I don't think that such a behavior would be expected. The user configures a set of APQNs to be used for the token, so he assumes fail over within those APQNs. Using only certain APQNs for certain mechanisms undergoes this fail over capability, and may lead to  a situation where the user thinks that he has configured it in a save way for fail over if one APQN breaks, but in fact this is not true for some mechanisms. So IF we ever do something like this, then we should have a config option that needs to be specified to allow such behavior explicitly.
>
>
This is only one side of the coin. I think there are customers who think, ock should smart detect what each APQN can do and do a
smart dispatching on all APQNs which are able torun the requested crypto function. However, there should be a warning or so if
there is no backup APQN for an mechanismus.

But Ingo, I do not expect to have this solved with this pull request. I just think, we should think into this direction and find a solution for
example when a customer runs a z15 with let's say 2 CEX6 and 2 CEX7 and uses a simple configuration like one domain assigned to an
LPAR and ANY in the ep11.conf and now wants to run Dilitium Load (or any other load only available only with CEX7).
We are currently sticking to the max model (the set of functions is the intersection of all functions provided by each APQN).
But a min model (the set of functions is the union of all functions provided by each APQN) would also make sense.

The hotplug issue is another topic. However, long term we should detect hot(un)plug and have a way to dynamically figure out the set of available functions.

>

PR #313, opened 2020-09-24
Fixes for potential vulnerabilities

juergenchrist:
Don't use environment variables in all cases.  If running under setuid, we
should not trust the environment without additional sanitization.  Especially
for loading libraries, we cannot use input from untrusted environment
variables.

Some paths were written under some assumptions that paths never overflow.
Instead, we use a checked path creation.  If the path name overflows, the
function that tries to use the path name fails.

This commit only includes reworks on getenv and sprintf.  Further functions
should still be checked.

There are also some minor cosmetic fixes:
- Don't calloc if we anyway write all allocated bytes directly afterwards.
- Don't allocate paths larger than PATH_MAX.

### DIFF
--- a/usr/lib/common/loadsave.c
+++ b/usr/lib/common/loadsave.c
@@ -41,6 +41,7 @@
 #include "sw_crypt.h"
 #include "trace.h"
 #include "ock_syslog.h"
+#include "slotmgr.h" // for ock_snprintf
 
 extern void set_perm(int);
 
@@ -50,16 +51,71 @@ CK_RV reload_token_object_old(STDLL_TokData_t *tokdata, OBJECT *obj);
 CK_RV save_public_token_object_old(STDLL_TokData_t *tokdata, OBJECT *obj);
 CK_RV load_public_token_objects_old(STDLL_TokData_t *tokdata);
 
-char *get_pk_dir(STDLL_TokData_t *tokdata, char *fname)
+static int get_token_object_path(char *buf, size_t buflen,
+                                 STDLL_TokData_t *tokdata, char *path)
 {
+    if (ock_snprintf(buf, buflen, "%s/" PK_LITE_OBJ_DIR "/%s",
+                     tokdata->data_store, path) != 0) {
+        TRACE_ERROR("buffer overflow for object path %s", path);
+        return -1;
+    }
+    return 0;
+}
+
+static FILE *open_token_object_path(char *buf, size_t buflen,
+                                    STDLL_TokData_t *tokdata, char *path,
+                                    char *mode)
+{
+    if (get_token_object_path(buf, buflen, tokdata, path) < 0)
+        return NULL;
+    return fopen(buf, mode);
+}
+
+static int get_token_data_store_path(char *buf, size_t buflen,
+                                     STDLL_TokData_t *tokdata, char *path)
+{
+    if (ock_snprintf(buf, buflen, "%s/%s", tokdata->data_store, path)) {
+        TRACE_ERROR("buffer overflow for path %s", path);
+        return -1;
+    }
+    return 0;
+}
+
+static FILE *open_token_data_store_path(char *buf, size_t buflen,
+                                        STDLL_TokData_t *tokdata, char *path,
+                                        char *mode)
+{
+    if (get_token_data_store_path(buf, buflen, tokdata, path) < 0)
+        return NULL;
+    return fopen(buf, mode);
+}
+
+static FILE *open_token_object_index(char *buf, size_t buflen,
+                                     STDLL_TokData_t *tokdata, char *mode)
+{
+    return open_token_object_path(buf, buflen, tokdata, PK_LITE_OBJ_IDX, mode);
+}
+
+static FILE *open_token_nvdat(char *buf, size_t buflen,
+                              STDLL_TokData_t *tokdata, char *mode)
+{
+    if (ock_snprintf(buf, buflen, "%s/" PK_LITE_NV, tokdata->data_store)) {
+        TRACE_ERROR("NVDAT.TOK file name buffer overflow\n");
+        return NULL;
+    }
+    return fopen(buf, mode);
+}
+
+char *get_pk_dir(STDLL_TokData_t *tokdata, char *fname, size_t len)
+{
+    int snres;
     struct passwd *pw = NULL;
 
     if (token_specific.data_store.per_user && (pw = getpwuid(getuid())) != NULL)
-        sprintf(fname, "%s/%s", tokdata->pk_dir, pw->pw_name);
+        snres = ock_snprintf(fname, len, "%s/%s", tokdata->pk_dir, pw->pw_name);
     else
-        sprintf(fname, "%s", tokdata->pk_dir);
-
-    return fname;
+        snres = ock_snprintf(fname, len, "%s", tokdata->pk_dir);
+    return snres != 0 ? NULL : fname;
 }
 
 void set_perm(int file)
@@ -112,9 +168,7 @@ CK_RV save_token_object(STDLL_TokData_t *tokdata, OBJECT *obj)
         return rc;
 
     // update the index file if it exists
-    sprintf(fname, "%s/%s/%s", tokdata->data_store, PK_LITE_OBJ_DIR,
-            PK_LITE_OBJ_IDX);
-    fp = fopen(fname, "r");
+    fp = open_token_object_index(fname, sizeof(fname), tokdata, "r");
     if (fp) {
         set_perm(fileno(fp));
         while (fgets(line, 50, fp)) {
@@ -151,11 +205,6 @@ CK_RV delete_token_object(STDLL_TokData_t *tokdata, OBJECT *obj)
     FILE *fp1, *fp2;
     char objidx[PATH_MAX], idxtmp[PATH_MAX], fname[PATH_MAX], line[256];
 
-    sprintf(objidx, "%s/%s/%s", tokdata->data_store,
-            PK_LITE_OBJ_DIR, PK_LITE_OBJ_IDX);
-    sprintf(idxtmp, "%s/%s/%s", tokdata->data_store,
-            PK_LITE_OBJ_DIR, "IDX.TMP");
-
     // FIXME:  on UNIX, we need to make sure these guys aren't symlinks
     //         before we blindly write to these files...
     //
@@ -163,8 +212,9 @@ CK_RV delete_token_object(STDLL_TokData_t *tokdata, OBJECT *obj)
     // remove the object from the index file
     //
 
-    fp1 = fopen(objidx, "r");
-    fp2 = fopen(idxtmp, "w");
+    fp1 = open_token_object_index(objidx, sizeof(objidx), tokdata, "r");
+    fp2 = open_token_object_path(idxtmp, sizeof(idxtmp),
+                                 tokdata, "IDX.TMP", "w");
     if (!fp1 || !fp2) {
         if (fp1)
             fclose(fp1);
@@ -206,9 +256,11 @@ CK_RV delete_token_object(STDLL_TokData_t *tokdata, OBJECT *obj)
     fclose(fp1);
     fclose(fp2);
 
-    sprintf(fname, "%s/%s/%s", tokdata->data_store,
-            PK_LITE_OBJ_DIR, (char *) obj->name);
-    unlink(fname);
+    if (get_token_object_path(fname, sizeof(fname), tokdata,
+                              (char *) obj->name) < 0)
+       TRACE_DEVEL("file name buffer overflow in obj unlink\n");
+    else
+        unlink(fname);
 
     return CKR_OK;
 }
@@ -239,36 +291,43 @@ done:
     return rc;
 }
 
-void init_data_store(STDLL_TokData_t *tokdata, char *directory,
-                     char *data_store)
+CK_RV init_data_store(STDLL_TokData_t *tokdata, char *directory,
+                      char *data_store, size_t len)
 {
     char *pkdir;
+    int pklen;
 
     if (tokdata->pk_dir != NULL) {
         free(tokdata->pk_dir);
         tokdata->pk_dir = NULL;
     }
 
-    if ((pkdir = getenv("PKCS_APP_STORE")) != NULL) {
-        tokdata->pk_dir = (char *) malloc(strlen(pkdir) + 1024);
-        memset(tokdata->pk_dir, 0, strlen(pkdir) + 1024);
-        sprintf(tokdata->pk_dir, "%s/%s", pkdir, SUB_DIR);
-        get_pk_dir(tokdata, data_store);
-        return;
+    if ((pkdir = secure_getenv("PKCS_APP_STORE")) != NULL) {
+        pklen = strlen(pkdir) + 1024;
+        tokdata->pk_dir = (char *) calloc(pklen, 1);
+        if (!(tokdata->pk_dir))
+            return CKR_HOST_MEMORY;
+        if (ock_snprintf(tokdata->pk_dir, pklen, "%s/%s", pkdir, SUB_DIR) != 0)
+            return CKR_FUNCTION_FAILED;
+        return get_pk_dir(tokdata, data_store, len) ? CKR_OK : CKR_FUNCTION_FAILED;
     }
 
     if (directory) {
-        tokdata->pk_dir = (char *) malloc(strlen(directory) + 25);
-        memset(tokdata->pk_dir, 0, strlen(directory) + 25);
-        sprintf(tokdata->pk_dir, "%s", directory);
+        pklen = strlen(directory) + 1;
+        tokdata->pk_dir = (char *) calloc(pklen, 1);
+        if (!(tokdata->pk_dir))
+            return CKR_HOST_MEMORY;
+        if (ock_snprintf(tokdata->pk_dir, pklen, "%s", directory) != 0)
+            return CKR_FUNCTION_FAILED;
     } else {
-        tokdata->pk_dir = (char *) malloc(strlen(PK_DIR) + 25);
-        memset(tokdata->pk_dir, 0, strlen(PK_DIR) + 25);
-        sprintf(tokdata->pk_dir, "%s", PK_DIR);
+        pklen = strlen(PK_DIR) + 1;
+        tokdata->pk_dir = (char *) calloc(pklen, 1);
+        if (!(tokdata->pk_dir))
+            return CKR_HOST_MEMORY;
+        if (ock_snprintf(tokdata->pk_dir, pklen, "%s", PK_DIR) != 0)
+            return CKR_FUNCTION_FAILED;
     }
-    get_pk_dir(tokdata, data_store);
-
-    return;
+    return get_pk_dir(tokdata, data_store, len) ? CKR_OK : CKR_FUNCTION_FAILED;
 }
 
 void final_data_store(STDLL_TokData_t * tokdata)
@@ -580,8 +639,7 @@ CK_RV load_token_data_old(STDLL_TokData_t *tokdata, CK_SLOT_ID slot_id)
         goto out_nolock;
     }
 
-    sprintf(fname, "%s/%s", tokdata->data_store, PK_LITE_NV);
-    fp = fopen(fname, "r");
+    fp = open_token_nvdat(fname, sizeof(fname), tokdata, "r");
     if (!fp) {
         /* Better error checking added */
         if (errno == ENOENT) {
@@ -654,8 +712,7 @@ CK_RV save_token_data_old(STDLL_TokData_t *tokdata, CK_SLOT_ID slot_id)
         goto out_nolock;
     }
 
-    sprintf(fname, "%s/%s", tokdata->data_store, PK_LITE_NV);
-    fp = fopen(fname, "w");
+    fp = open_token_nvdat(fname, sizeof(fname), tokdata, "w");
     if (!fp) {
         TRACE_ERROR("fopen(%s): %s\n", fname, strerror(errno));
         rc = CKR_FUNCTION_FAILED;
@@ -784,8 +841,12 @@ static CK_RV save_private_token_object_old(STDLL_TokData_t *tokdata, OBJECT *obj
         goto error;
     }
 
-    sprintf(fname, "%s/%s/", tokdata->data_store, PK_LITE_OBJ_DIR);
-    strncat(fname, (char *) obj->name, 8);
+    if (ock_snprintf(fname, PATH_MAX, "%s/%s/%.8s", tokdata->data_store,
+                     PK_LITE_OBJ_DIR, (char *)obj->name) != 0) {
+        TRACE_ERROR("private token object old name buffer overflow\n");
+        rc = CKR_FUNCTION_FAILED;
+        goto error;
+    }
     fp = fopen(fname, "w");
     if (!fp) {
         TRACE_ERROR("fopen(%s): %s\n", fname, strerror(errno));
@@ -844,20 +905,14 @@ CK_RV load_private_token_objects_old(STDLL_TokData_t *tokdata)
     CK_RV rc;
     size_t read_size;
 
-    sprintf(iname, "%s/%s/%s", tokdata->data_store, PK_LITE_OBJ_DIR,
-            PK_LITE_OBJ_IDX);
-
-    fp1 = fopen(iname, "r");
+    fp1 = open_token_object_index(iname, sizeof(iname), tokdata, "r");
     if (!fp1)
         return CKR_OK;          // no token objects
 
     while (fgets(tmp, 50, fp1)) {
         tmp[strlen(tmp) - 1] = 0;
 
-        sprintf(fname, "%s/%s/", tokdata->data_store, PK_LITE_OBJ_DIR);
-        strcat(fname, tmp);
-
-        fp2 = fopen(fname, "r");
+        fp2 = open_token_object_path(fname, sizeof(fname), tokdata, tmp, "r");
         if (!fp2)
             continue;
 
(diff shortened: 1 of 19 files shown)


usr/sbin/pkcsslotd/log.c
```diff
@@ -268,11 +268,9 @@ BOOL NewLoggingFacility(char *ID, pLoggingFacility pStuff)
...
     pInfo->LogLevel = pStuff->LogLevel;
 
-    sprintf(pInfo->Descrip, "%s %s", pStuff->Label, ID);
-
-    /* ensure that the last character is a NULL */
-    pInfo->Descrip[sizeof(pInfo->Descrip) - 1] = '\0';
-
+    /* truncate description to fit buffer */
+    snprintf(pInfo->Descrip, sizeof(pInfo->Descrip), "%s %s",
+             pStuff->Label, ID);
```
ifranzki, line 273: Don't we need a zero termination here? What if snprintf ran to sizeof(pInfo->Descrip) and thus did not add the zero termination?

juergenchrist: snprintf always adds a 0-termination.  It copies at most n-1 chars and then places the 0 according to C standard.

ifranzki: You are right, I was at strncpy, which does not....

usr/lib/common/loadsave.c
```diff
@@ -239,36 +300,43 @@ CK_RV delete_token_data(STDLL_TokData_t *tokdata)
...
-    if ((pkdir = getenv("PKCS_APP_STORE")) != NULL) {
-        tokdata->pk_dir = (char *) malloc(strlen(pkdir) + 1024);
-        memset(tokdata->pk_dir, 0, strlen(pkdir) + 1024);
-        sprintf(tokdata->pk_dir, "%s/%s", pkdir, SUB_DIR);
-        get_pk_dir(tokdata, data_store);
-        return;
+    if ((pkdir = secure_getenv("PKCS_APP_STORE")) != NULL) {
+        pklen = strlen(pkdir) + 1024;
+        tokdata->pk_dir = (char *) malloc(pklen);
+        memset(tokdata->pk_dir, 0, pklen);
```
ifranzki, line 317: Use calloc here?  But ist fine with me either way.

usr/lib/common/loadsave.c
```diff
@@ -239,36 +300,43 @@ CK_RV delete_token_data(STDLL_TokData_t *tokdata)
...
+            return -1;
+        return get_pk_dir(tokdata, data_store, len) ? CKR_OK : CKR_FUNCTION_FAILED;
     }
 
     if (directory) {
-        tokdata->pk_dir = (char *) malloc(strlen(directory) + 25);
-        memset(tokdata->pk_dir, 0, strlen(directory) + 25);
-        sprintf(tokdata->pk_dir, "%s", directory);
+        pklen = strlen(directory) + 25;
+        tokdata->pk_dir = (char *) malloc(pklen);
```
ifranzki, line 327: same here

usr/lib/common/loadsave.c
```diff
@@ -239,36 +300,43 @@ CK_RV delete_token_data(STDLL_TokData_t *tokdata)
...
+        if (splen < 0 || splen >= pklen)
+            return -1;
+        return get_pk_dir(tokdata, data_store, len) ? CKR_OK : CKR_FUNCTION_FAILED;
     }
 
     if (directory) {
-        tokdata->pk_dir = (char *) malloc(strlen(directory) + 25);
-        memset(tokdata->pk_dir, 0, strlen(directory) + 25);
-        sprintf(tokdata->pk_dir, "%s", directory);
+        pklen = strlen(directory) + 25;
```
ifranzki, line 326: Did you find put why we add 25 here?  The s(n)printf a few lines down just fills it with exactly strlen()+1 characters.... So I would understand +1, but why +25?

usr/lib/common/new_host.c
```diff
@@ -115,13 +115,27 @@ CK_RV ST_Initialize(API_Slot_t *sltp, CK_SLOT_ID SlotNumber,
     bt_init(&sltp->TokData->publ_token_obj_btree, call_object_free);
 
     if (strlen(sinfp->tokname)) {
-        sprintf(abs_tokdir_name, "%s/%s", CONFIG_PATH, sinfp->tokname);
+        int snres;
```
ifranzki, line 118: Not sure if it is a good idea to declare a variable in an if block. I remember that I  run into problems with such for certain distros, where the compiler flagged this as an error... Better declare the variable at the beginning of the function

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -6883,7 +6884,11 @@ static int read_adapter_config_file(STDLL_TokData_t * tokdata,
     int whitemode = 0;
     int anymode = 0;
     int apqn_i = 0;             /* how many APQN numbers */
-    char *conf_dir = getenv("OCK_EP11_TOKEN_DIR");
+    /* Since the ep11 token config contains the path to libica that
+     * will later be dlopen()ed, we cannot use a token directory from
```
ifranzki, line 6888: I would call that the 'token config directory'. The 'token directory' usually is the directory where the token stores its data, e.g. /var/lib/opencryptoki/<token-name>/. 
Yes the environment variable name is misleading as well, but lets at least precise in the comment :-)

usr/lib/ep11_stdll/new_host.c
```diff
@@ -106,13 +106,26 @@ CK_RV ST_Initialize(API_Slot_t *sltp, CK_SLOT_ID SlotNumber,
     bt_init(&sltp->TokData->publ_token_obj_btree, call_object_free);
 
     if (strlen(sinfp->tokname)) {
-        sprintf(abs_tokdir_name, "%s/%s", CONFIG_PATH, sinfp->tokname);
-        TRACE_DEVEL("Token directory: %s\n", abs_tokdir_name);
-        init_data_store(sltp->TokData, (char *) abs_tokdir_name,
-                        sltp->TokData->data_store);
+        int snres = snprintf(abs_tokdir_name, PATH_MAX, "%s/%s",
```
ifranzki, line 109: same comment as for the other ST_Initialize functions in the other new_host.c files. 
Also this one is slightly different to the previous. 
If we duplicate code, then we should at least make sure that the duplicated code is the same as much as possible. This makes it easier to keep the copies in sync.

ifranzki, line 109: same comment as for the other ST_Initialize functions in the other new_host.c files. 
Also this one is slightly different to the previous. 
If we duplicate code, then we should at least make sure that the duplicated code is the same as much as possible. This makes it easier to keep the copies in sync.

usr/lib/icsf_stdll/new_host.c
```diff
@@ -104,13 +104,25 @@ CK_RV ST_Initialize(API_Slot_t *sltp, CK_SLOT_ID SlotNumber,
     bt_init(&sltp->TokData->publ_token_obj_btree, call_object_free);
 
     if (strlen(sinfp->tokname)) {
-        sprintf(abs_tokdir_name, "%s/%s", CONFIG_PATH, sinfp->tokname);
-        TRACE_DEVEL("Token directory: %s\n", abs_tokdir_name);
-        init_data_store(sltp->TokData, (char *) abs_tokdir_name,
-                        sltp->TokData->data_store);
+        int snres = snprintf(abs_tokdir_name, PATH_MAX, "%s/%s",
```
ifranzki, line 107: same here again

usr/sbin/pkcsslotd/log.c
```diff
@@ -421,7 +420,8 @@ BOOL PKCS_Log(pLogHandle phLog, char *Format, va_list ap)
         }
     }
 
-    if (vsprintf(&(Buffer[0]), Format, ap) < 0) {
+    snres = vsnprintf(&(Buffer[0]), PATH_MAX, Format, ap);
```
ifranzki, line 423: while you are at it, lets get rid of  this ugly `&(Buffer[0])`...
Why not just `Buffer` ?

usr/lib/common/loadsave.c
```diff
@@ -864,20 +936,14 @@ CK_RV load_private_token_objects_old(STDLL_TokData_t *tokdata)
...
 
-    sprintf(iname, "%s/%s/%s", tokdata->data_store, PK_LITE_OBJ_DIR,
-            PK_LITE_OBJ_IDX);
-
-    fp1 = fopen(iname, "r");
+    fp1 = open_token_object_index(iname, sizeof(iname), tokdata, "r");
     if (!fp1)
         return CKR_OK;          // no token objects
 
     while (fgets(tmp, 50, fp1)) {
```
hfreude, line 943: where does this "50" come from ?
Can't this be replaced by some kind of sizeof(tmp) ?

usr/lib/common/loadsave.c
```diff
@@ -978,10 +1044,9 @@ CK_RV load_masterkey_so_old(STDLL_TokData_t *tokdata)
     // this file gets created on C_InitToken so we can assume that it always
     // exists
     //
-    sprintf(fname, "%s/MK_SO", tokdata->data_store);
-    fp = fopen(fname, "r");
+    fp = open_token_data_store_path(fname, sizeof(fname), tokdata, "MK_SO",
+                                    "r");
```
hfreude, line 1048: are you really sure you want to break this line with just an "r"); on the next ?
I would not hesitate to move this back to the prev line (and get some complain from any checker or whatever).

usr/lib/common/loadsave.c
```diff
@@ -1081,10 +1146,9 @@ CK_RV load_masterkey_user_old(STDLL_TokData_t *tokdata)
     // this file gets created on C_InitToken so we can assume that it always
     // exists
     //
-    sprintf(fname, "%s/MK_USER", tokdata->data_store);
-    fp = fopen(fname, "r");
+    fp = open_token_data_store_path(fname, sizeof(fname), tokdata, "MK_USER",
+                                    "r");
```
hfreude, line 1150: same here - it is much more better readable to have the "r"); moved back to were it belongs.

usr/lib/common/loadsave.c
```diff
@@ -1205,10 +1269,9 @@ CK_RV save_masterkey_so_old(STDLL_TokData_t *tokdata)
     //
     // probably ought to ensure the permissions are correct
     //
-    sprintf(fname, "%s/MK_SO", tokdata->data_store);
-    fp = fopen(fname, "w");
+    fp = open_token_data_store_path(fname, sizeof(fname), tokdata, "MK_SO",
+                                    "w");
```
hfreude, line 1273: the linux kernel now allows 100 characters per line if it is better for the readability

usr/lib/common/loadsave.c
```diff
@@ -1293,10 +1356,9 @@ CK_RV save_masterkey_user_old(STDLL_TokData_t *tokdata)
     //
     // probably ought to ensure the permissions are correct
     //
-    sprintf(fname, "%s/MK_USER", tokdata->data_store);
-    fp = fopen(fname, "w");
+    fp = open_token_data_store_path(fname, sizeof(fname), tokdata, "MK_USER",
+                                    "w");
```
hfreude, line 1360: poor lonely "w"...

usr/lib/common/loadsave.c
```diff
@@ -1604,20 +1675,14 @@ CK_RV load_public_token_objects_old(STDLL_TokData_t *tokdata)
...
 
-    sprintf(iname, "%s/%s/%s", tokdata->data_store, PK_LITE_OBJ_DIR,
-            PK_LITE_OBJ_IDX);
-
-    fp1 = fopen(iname, "r");
+    fp1 = open_token_object_index(iname, sizeof(iname), tokdata, "r");
     if (!fp1)
         return CKR_OK;          // no token objects
 
     while (fgets(tmp, 50, fp1)) {
```
hfreude, line 1682: again this 50 value out of the blue - check for replace with some sizeof(tmp)

usr/lib/common/loadsave.c
```diff
@@ -2415,20 +2474,14 @@ CK_RV load_private_token_objects(STDLL_TokData_t *tokdata)
...
```
hfreude, line 2481: check for sizeof(tmp) instead of "50"

usr/lib/common/new_host.c
hfreude: Ingo this is C standart. With every block you can declare variables. What's the issue with this ?
I know that C99 and C++ allow variables in the middle of code, but Ansi C90 restricts to the beginnning
of a code block - but not to the beginning of a function.

usr/lib/common/utility.c
```diff
@@ -256,9 +256,13 @@ CK_RV CreateXProcLock(char *tokname, STDLL_TokData_t *tokdata)
         /** create lock subdir for each token if it doesn't exist.
          *  The root directory should be created in slotmgr daemon **/
         if (strlen(tokname) > 0)
-            sprintf(lockdir, "%s/%s", LOCKDIR_PATH, tokname);
+            snres = snprintf(lockdir, PATH_MAX, "%s/%s", LOCKDIR_PATH, tokname);
         else
-            sprintf(lockdir, "%s/%s", LOCKDIR_PATH, SUB_DIR);
+            snres = snprintf(lockdir, PATH_MAX, "%s/%s", LOCKDIR_PATH, SUB_DIR);
+        if (snres < 0 || snres >= PATH_MAX) {
+            OCK_SYSLOG(LOG_ERR, "lock directory path too long\n");
```
hfreude, line 263: please add an ock error trace also here.

hfreude: or maybe, the ock_syslock does this already ?

usr/lib/common/utility.c
```diff
@@ -292,9 +296,15 @@ CK_RV CreateXProcLock(char *tokname, STDLL_TokData_t *tokdata)
...
-            sprintf(lockfile, "%s/%s/LCK..%s", LOCKDIR_PATH, tokname, tokname);
+            snres = snprintf(lockfile, PATH_MAX, "%s/%s/LCK..%s",
+                             LOCKDIR_PATH, tokname, tokname);
         else
-            sprintf(lockfile, "%s/%s/LCK..%s", LOCKDIR_PATH, SUB_DIR, SUB_DIR);
+            snres = snprintf(lockfile, sizeof(lockfile), "%s/%s/LCK..%s",
+                             LOCKDIR_PATH, SUB_DIR, SUB_DIR);
+        if (snres < 0 || snres >= PATH_MAX) {
+            OCK_SYSLOG(LOG_ERR, "lock file path too long\n");
+            goto err;
```
hfreude, line 306: same here, if ock_syslog does not also emit an ock error log

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -13,6 +13,7 @@
      ==========
 ****************************************************************************/
 
+#define _GNU_SOURCE
```
hfreude, line 16: why do we need this ? maybe add a comment
may it make sense to add this to the global compile options ?

usr/lib/common/loadsave.c
jschmidb: Yes, we had this discussion also when reviewing the token migration tool. There we changed the code similar to:
```
char tmp[PATH_MAX];
while (fgets(tmp, PATH_MAX, fp)) { ... }
```
However, we are just reading token object names here. They only have 8 chars.

usr/sbin/pkcstok_migrate/pkcstok_migrate.c
```diff
@@ -127,7 +164,7 @@ static CK_RV migrate_public_token_object(const char *data_store, const char *nam
                                          unsigned char *data, unsigned long len)
 {
     const char *tokobj = "TOK_OBJ";
-    char fname[PATH_MAX + 1 + strlen(tokobj) + 1 + strlen(name) + 1];
+    char fname[PATH_MAX];
```
jschmidb, line 167: I agree. In the original code this ugly length calculation was necessary to avoid compiler warnings.

usr/lib/common/new_host.c
ifranzki: Maybe I confused this with another case. Back then I hade something like this:
`for(int i=0, .... ` and the compiler flagged this as only allowed in C99.

usr/lib/common/loadsave.c
ifranzki: I guess Juergen did not change this since it is not related to his fix.
Although I agree that it could be enhanced, but please in a different commit.

usr/lib/tpm_stdll/tpm_specific.c
```diff
@@ -193,17 +194,23 @@ CK_RV token_specific_init(STDLL_TokData_t * tokdata, CK_SLOT_ID SlotNumber,
...
+            TRACE_ERROR("mkdir(%s): %s\n", fname, strerror(errno));
             return CKR_FUNCTION_FAILED;
         }
     }
     // now create userdir/TOK_OBJ if it doesn't exist
-    strncat(path_buf, "/", sizeof(path_buf) - (strlen(path_buf) + 1));
-    strncat(path_buf, PK_LITE_OBJ_DIR,
-            sizeof(path_buf) - (strlen(PK_LITE_OBJ_DIR) + 1));
+    snres = snprintf(path_buf, PATH_MAX, "%s/%s", fname, PK_LITE_OBJ_DIR);
+    /* Support for old glibc: snprintf might return -1 */
```
hfreude, line 209: can you please remove this comment ? it does not tell me anything.

usr/sbin/pkcsslotd/log.c
hfreude: I'd also recomment to rename Buffer to buffer ... or buf and Formant to format or fmt ...

usr/sbin/pkcsslotd/slotmgr.c
```diff
@@ -54,7 +54,7 @@ extern BOOL IveDaemonized;
 void DumpSharedMemory(void)
 {
     u_int32 *p;
-    char Buf[PATH_MAX];
+    char Buf[4 * 8 + 4];
```
hfreude, line 57: and please rename to "buf"

usr/lib/common/loadsave.c
juergenchrist: Will do.

usr/lib/common/loadsave.c
juergenchrist: done

usr/lib/common/loadsave.c
juergenchrist: No, I kept the size for now without further investigation on why 25.  I just checked that it is not too small.  Changed this to 1.

usr/lib/common/new_host.c
juergenchrist: I moved it to the top of the function.

usr/lib/ep11_stdll/ep11_specific.c
juergenchrist: done

usr/lib/ep11_stdll/new_host.c
juergenchrist: Moved the declaration to the top of the function and kept assignment here.

juergenchrist: done

usr/lib/icsf_stdll/new_host.c
juergenchrist: done

usr/sbin/pkcsslotd/log.c
juergenchrist: Renamed to buf and fmt and fixed the name clash with the time buffer (called "buf" previously and "timebuf" now).

usr/lib/common/loadsave.c
juergenchrist: Changed this.  Maybe we should think about a different dimension for tmp than PATH_MAX?

usr/lib/common/loadsave.c
juergenchrist: If it is okay with the project, I am all for it.  Felt silly splitting the line there, but according to the project rules 

> the limit on the length of lines is 80 columns and this is a strongly preferred limit

Will change it anyway.

usr/lib/common/loadsave.c
juergenchrist: Agreed and changed.

usr/lib/common/loadsave.c
juergenchrist: That line is now 81 characters...

usr/lib/common/loadsave.c
juergenchrist: made it less lonely

usr/lib/common/loadsave.c
juergenchrist: Done.  Actually, checks for fgets were not done in these changes.  They are still pending.  But due to the size of the changes I wanted to upload that amount already in advance.

usr/lib/common/loadsave.c
juergenchrist: fgets is still on my todo list.  Will change it in a separate commit later.

usr/lib/common/loadsave.c
jschmidb: There are 4 places in OCK with fgets(tmp, 50 ...). The token object names have just 8 chars, but when reviewing the token migration tool, we changed this to PATH_MAX. Why not do the same here?

usr/lib/common/utility.c
juergenchrist: OCK_SYSLOCK does not add an error trace outside the syslog.  Added a TRACE_ERROR line with the same message.

usr/lib/common/utility.c
juergenchrist: done

usr/sbin/pkcstok_migrate/pkcstok_migrate.c
```diff
@@ -1678,10 +1717,8 @@ static CK_RV create_NVTOK_DAT_312(const char *data_store, const char *sopin,
...
 
     /* Create new file NVTOK.DAT_312 */
-    sprintf(fname, "%s/%s", data_store, nvtok);
-    fp = fopen(fname, "w");
+    fp = open_datastore_file(fname, sizeof(fname), data_store, nvtok, "w");
     if (!fp) {
-        TRACE_ERROR("fopen(%s) failed, errno=%s\n", fname, strerror(errno));
         ret = CKR_FUNCTION_FAILED;
         goto done;
     }
```
hfreude, line 1724: I would suggest to write a function
inline int __attribute__((format(printf,3,4))) ock_snprintf(buf, buflen, fmt, ...)
{
    va_list ap;
    int n;

    va_start(ap, fmt);
    n = vsnprintf(buf, buflen, fmt, ap);
    va_end(ap);

    if (n < 0 || n >= buflen)
        return -1;
  
    return 0;
}

and use this

if (ock_snprintf(buf, sizeof(buf), "%s %s ...", str1, str2) != 0) {
    TRACE_ERROR("...");
    // do something more
}

usr/lib/common/new_host.c
hfreude: and ingo the compiler is right. A variable declaration like you show here is not allowed with Ansi C.

usr/lib/ep11_stdll/ep11_specific.c
juergenchrist: secure_getenv is only available if _GNU_SOURCE is defined.  Do we need it everywhere?  I added a comment.

usr/sbin/pkcstok_migrate/pkcstok_migrate.c
juergenchrist: Well, if a path that long is needed, fopen will return ENAMETOOLONG.  So no point in using more memory and a dynamically-sized array.

usr/lib/tpm_stdll/tpm_specific.c
juergenchrist: Done, but newer glibc cannot return -1 from snprintf anymore.  That is what the comment is about (also see NOTES in man 3 snprintf).

usr/sbin/pkcsslotd/slotmgr.c
juergenchrist: Done, but this file contains a lot of CamelCase variables.  Maybe that is something for a different set of patches.

usr/sbin/pkcstok_migrate/pkcstok_migrate.c
juergenchrist: I don't understand this comment.  What does it refer to?  Error tracing is moved into the open_datastore_file function.  Where would I need this new function?

hfreude: It refers to all the code above. You could declare a new function ock_snprintf like above and use it
everywhere in ock where as of now snprintf is used. It delegates the right return code handling to this
function and the caller does not even need to declare the int snres variable but only check if the returncode
is != 0 for failure.
That's a suggestion I would do - you may have another opinion about this.

usr/lib/common/new_host.c
p-steuer: I like to declare a variable that is only used within a block at the top of that block to limit its scope. Such code is usually easier to review (and conforms to ansi c). Declaring variables in the middle of a code block or for loop is definitely not ansi c and IMO should be avoided.

p-steuer: However, im also fine with declaring all variables at the top of a function. Variables that are intended to be used only inside the scope of a block statement should then only be initialized inside that block such that you get an undef var warning when using the variable out of its intended scope. Just my 2 cents..

usr/lib/common/loadsave.c
p-steuer: For users like me who like to vertically split their editors multiple times, im still a fan of the 80 columns limit, but i dont mind a few characters more here and there ..
Regarding line breaks in function argmument lists, i like to break the line such that all arguments stay between the brakets i.e., the last line ist the longest e.g.,
```
fp = open_token_data_store_path(fname, sizeof(fname),
                                tokdata, "MK_SO", "r");
```
However, since its just about a few characters more here, im okay with it.

p-steuer (approved):
Looks good to me.

usr/sbin/pkcstok_migrate/pkcstok_migrate.c
juergenchrist: Okay, this would remove the need for the additional local variable that stores the return value.  However, I think this requires a new header that should probably not be in usr/include, right?

juergenchrist: Done.  Also simplified one more piece of code (at the cost of an additional local variable).

PR #314, opened 2020-09-25
Restructure config parser location and reuse for token migration tool

### LINKED ISSUE #312: opencryptoki.conf parser mismatch in migration tool
In pkcstok_migrate.c we have the lines
`    snprintf(parm, sizeof(parm), "slot %ld", slot_id);
    while (fgets(line, sizeof(line), fp)) {
        if (!begins_with(line, '#') && strstr(line, parm)) {
            ret = get_stdll_name(fp, dll_name);
            goto done;
        }
    }
`
This parser is in no way aligned to the parser of the slot daemon.  Furthermore, it wrongly parses a lot of configurations:

- "slot  1" (with two spaces after the "slot" keyword)
- searching for slot 1 might hit slot 10 (since we do a prefix search)
- if the line does not contain a "{" after the slot number, the config line does not get rejected
- 80 spaces after "slot" can never be parsed with this simple solution, but according to the grammar of the slot daemon, this would be valid.

We should align these parsers.

juergenchrist:
Since now multiple tools have to read the opencryptoki.conf file, I restructure the parser and move it to a common location.  Furthermore, both the slot daemon and the migration tool are updated to use the parser.

Contains numerous fixes for potential buffer overflows in the parser.  As a design decision, tokens whose configuration would lead to a buffer overflow are skipped/ignored.

The token migration tool only uses the in-memory representation of the configuration to locate the STDLL for the token.  For updating the configuration we still use string operations.  These got changed from a prefix-search to a regexp-search for the token to update.

Fixes #312


juergenchrist:
Superseded by #315

PR #315, opened 2020-09-28
Use streaming parser

### LINKED ISSUE #312: opencryptoki.conf parser mismatch in migration tool

juergenchrist:
Restructured the parser as a streaming parser that emits events to an external
processing entity.  With this method we can use the parser also in the
migration tool to find the correct stdll and also update the configuration
file while preserving comments and newlines (tabs and spaces are not
necessarily preserved).

The rework of the parser also updated the grammar to include empty lines and
whole line comments within the slot definition.  This resolves the ugliness

```
slot 1 {
 description = something # comment allowed here
}
```

versus

```
slot 1 {
# comment not allowed here
  description = something
}
```
Fixes #312 


ifranzki:
Looks pretty nice! I was expecting/fearing a more complex change, but this seems to be pretty straight forward. 

One minor thing: 
lexer.l seems to have bene moved with 'git mv', so git still knows where the file was located previously.
However, for parser.y this does not seem to be the case? 
```
---
 usr/lib/common/configparser.h               |  63 +++++
 usr/{sbin/pkcsslotd => lib/common}/lexer.l  |  15 +-
 usr/lib/common/parser.y                     | 217 +++++++++++++++
 usr/sbin/pkcsslotd/parser.y                 | 289 --------------------
```

juergenchrist:
Weird thing is that neither file has been moved via git mv.  Will try to get this as proper tracking of the move.

juergenchrist:
Ups, forgot about the find-renames heuristic.  Since the files changed a lot, this heuristic build into git does not detect them with the default threshold of 50% (more than 50% of the file stay the same => rename; otherwise add/delete pair).  A quick check shows that lexer.l has a 93% similarity, parser.y only a 34%.  So try the option `--find-renames=34` to get this as a rename.

usr/lib/common/parser.y
```diff
@@ -0,0 +1,217 @@
...
+
+	conf = fopen(configfile, "r");
+
+	if (!conf) {
+		fprintf(stderr, "Failed to open %s: %s\n", configfile, strerror(errno));
+		return -1;
+	}
+
+	yyin = conf;
+    parsefuncs = funcs;
```
p-steuer, line 204: looks like something is wrong with identation here.

usr/lib/common/parser.y
```diff
@@ -0,0 +1,217 @@
...
+    parsedata = private;
+    
+	do {
+		yyparse();
+
+	} while (!feof(yyin));
+
+	fclose(conf);
+    parsefuncs = NULL;
```
p-steuer, line 213: and here

p-steuer:
Great piece of work. Any ideas how to test that old config files do not become invalid (language inclusion is generally undecidable even for context free grammars). We dont have fuzzing in place yet. Maybe add a few small well-formed and ill-formed config file examples to the test dir and check for known anwers ? Those could later be used as input hints for the fuzzer.

usr/lib/common/parser.y
juergenchrist: Unfortunately the file was wrongly detected by my setup and indented with spaces.  Fixed by now.

usr/lib/common/parser.y
juergenchrist: Fixed.

juergenchrist:
I have some tests mostly around spacing.  Those are some tests that the old parser refused while the new parser allows them.  I was searching for a good tool to fuzz-test this, but I did not find a tool that takes a scanner- and a parser-definition and generates grammar-based input.

I will compile a set of config files that I will then turn into a regression test for the current parser (both, accept and reject cases).

ifranzki (approved):
lgtm

PR #318, opened 2020-09-30
ICSF: replace PBKDF implementation by libcrypto call

p-steuer:
We want to replace all of openCryptoki's SW implementations
by calling to libcrypto. Moreover, the HMAC interface used
by openCryptoki's PBKDF implementation is deprecated in
OpenSSL master.

Edit: It looks like there may be a difference between OCK's old PBKDF implementation and what OpenSSL implements, so this one is not working yet. Not planning to include this one in the october release.


p-steuer:
@juergenchrist found the difference between OpenSSL and OCK implementation.

ifranzki:
Resolved via https://github.com/opencryptoki/opencryptoki/pull/510

PR #320, opened 2020-10-12
Fix GCC 10 compiler warning in pkcstok_migrate.c

juergenchrist:
GCC 10 complains about a snprintf.  The real problem is the strcat in the next
line.  However, the lines can easily be simplified to also remove the warning.


ifranzki (approved):
lgtm

p-steuer (approved):
lgtm

PR #323, opened 2020-10-13
Conform to PKCS 11 v3.0 Baseline Provider profile

p-steuer:


ifranzki:
You could also update testcases/misc/reencrypt.c to use the vendor interface instead of dlysm'ing C_IBM_ReencryptSingle. But it works either way.....

usr/lib/common/profile_obj.c
```diff
@@ -0,0 +1,80 @@
...
+
+    class_attr->type = CKA_CLASS;
+    class_attr->ulValueLen = sizeof(CK_OBJECT_CLASS);
+    class_attr->pValue = (CK_BYTE *)class_attr + sizeof(CK_ATTRIBUTE);
+    *(CK_OBJECT_CLASS *)class_attr->pValue = CKO_PROFILE;
+
+    profile_id_attr->type = CKA_PROFILE_ID;
+    profile_id_attr->ulValueLen = sizeof(CK_PROFILE_ID);
+    profile_id_attr->pValue = (CK_BYTE *)profile_id_attr + sizeof(CK_ATTRIBUTE);;
+    *(CK_PROFILE_ID *)profile_id_attr->pValue = CKP_INVALID_ID; //XXX
```
ifranzki, line 57: What does the XXX mean? Is this a left over TODO ?

p-steuer: I was not sure what a reasonable default value is here, since the pkcs11 spec doesnt tell me, so this is kind of a question mark to reviewers.

testcases/common/common.c
```diff
@@ -1102,34 +1110,81 @@ int do_ParseArgs(int argc, char **argv)
...
+    CK_INTERFACE *interface;
+    CK_VERSION version;
+    CK_FLAGS flags;
+    CK_RV rc, rv;
     CK_RV(*pfoo) ();
     char *e;
     char *f = "libopencryptoki.so";
+    CK_ULONG nmemb = 0;
+
+    rv = FALSE;
```
jschmidb, line 1122: Looks like rv is a CK_RV, not a CK_BBOOL, are you intentionally assigning rv = FALSE?

usr/include/pkcs11types.h
```diff
@@ -580,6 +592,10 @@ typedef struct CK_DATE {
     CK_CHAR day[2];             /* the day   ("01" - "31") */
 } CK_DATE;
 
+/* CK_PROFILE_ID is an unsigend ulong value
+ * represting a specific token profile. */
```
jschmidb, line 596: representing

usr/lib/common/profile_obj.c
p-steuer: .. such that an application could create such an object and see what profile the lib conforms to based on the contents of the objct ? Thats another way of looking at it..

usr/include/pkcs11types.h
jschmidb: ... and unsigned

usr/lib/api/api_interface.c
```diff
@@ -57,13 +57,208 @@ void api_init();
...
+    C_GenerateKeyPair,
+    C_WrapKey,
+    C_UnwrapKey,
+    C_DeriveKey,
+    C_SeedRandom,
+    C_GenerateRandom,
+    C_GetFunctionStatus,
+    C_CancelFunction,
+    C_WaitForSlotEvent,
+
```
jschmidb, line 208: an intended blank line?

usr/lib/cca_stdll/cca_stdll.mk
```diff
@@ -33,8 +33,10 @@ opencryptoki_stdll_libpkcs11_cca_la_SOURCES =				\
 	usr/lib/common/utility.c usr/lib/common/mech_des.c		\
 	usr/lib/common/mech_des3.c usr/lib/common/mech_md5.c		\
 	usr/lib/common/mech_ssl3.c usr/lib/common/verify_mgr.c		\
-	usr/lib/common/p11util.c usr/lib/common/sw_crypt.c		\
-	usr/lib/common/shared_memory.c usr/lib/cca_stdll/cca_specific.c
+	usr/lib/common/p11util.c usr/lib/common/sw_crypt.c		\					\
```
jschmidb, line 36: This line contains 2 backslashes

usr/lib/common/profile_obj.c
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    switch (attr->type) {
+    case CKA_PROFILE_ID:
+        return CKR_OK;
+    default:
+        return template_validate_base_attribute(tmpl, attr, mode);
+    }
+
+    return CKR_OK;
```
jschmidb, line 79: Looks like this line is never reached. We always return in the above switch stmt.

usr/lib/common/template.c
```diff
@@ -999,10 +1037,16 @@ CK_RV template_set_default_common_attributes(TEMPLATE *tmpl)
     label_attr->ulValueLen = 0; // empty string
     label_attr->pValue = NULL;
 
+    unique_id_attr->type = CKA_UNIQUE_ID;
+    unique_id_attr->ulValueLen = sizeof(64);
```
jschmidb, line 1041: ok, the ulValueLen is sizeof(64), but what type does 64 have? It's just a number ...

p-steuer: Should just be 64 i.e. 2*UNIQUE_ID_LEN, dont know why comiler doesnt complain ..

usr/lib/common/profile_obj.c
p-steuer: Yes, but the base spec says "A given token can contain more than one profile ID" and which makes sense since a token could conform to multiple profiles. So i dont know how to make sense of it ..

p-steuer: My interpretation was that a token should contain a profile object for every profile it conforms to but this is no where stated explicitly.

ifranzki: right.... Not really well defined by PKCS#11 3.0 .....
So if profile objects are to be only created by the token itself, then you should actually hinder an application from creating one, I guess. So profile objects would be kind of read-only objects.

p-steuer: hmm the spec doesnt say those objects cannot be created , copyed etc by an application either.

testcases/common/common.c
p-steuer: done

testcases/common/common.c
```diff
@@ -1102,34 +1110,81 @@ int do_ParseArgs(int argc, char **argv)
...
+        testcase_error("C_GetInterface rc=%s", p11_get_ckr(rc));
+        goto ret;
     }
+    funcs3 = interface->pFunctionList;
 
-    atexit(unload_pkcslib);
-    return TRUE;
+    rv = TRUE;
+ret:
+    if (rv == TRUE) {
```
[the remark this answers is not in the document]
p-steuer: done

usr/include/pkcs11types.h
p-steuer: done

usr/lib/common/profile_obj.c
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    class_attr->type = CKA_CLASS;
+    class_attr->ulValueLen = sizeof(CK_OBJECT_CLASS);
+    class_attr->pValue = (CK_BYTE *)class_attr + sizeof(CK_ATTRIBUTE);
+    *(CK_OBJECT_CLASS *)class_attr->pValue = CKO_PROFILE;
+
+    profile_id_attr->type = CKA_PROFILE_ID;
+    profile_id_attr->ulValueLen = sizeof(CK_PROFILE_ID);
+    profile_id_attr->pValue = (CK_BYTE *)profile_id_attr + sizeof(CK_ATTRIBUTE);;
```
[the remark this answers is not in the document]
p-steuer: done

usr/lib/cca_stdll/cca_stdll.mk
p-steuer: done

usr/lib/common/profile_obj.c
p-steuer: done

usr/lib/common/template.c
p-steuer: done

usr/lib/common/profile_obj.c
p-steuer: So i implemented support for profile objects analogous to other object classes for now since the spec does not seem to treat them special. If it turns out tokens are required to come with r/o profile objects to conform to the baseline provider profile, i think its better to fix this as a bug than to publish something now that is not in the specs.

I added subtests for profile objects to the obj_mgmt_tests testcase.

p-steuer:
Added tests for profile objects and unique id attributes.

usr/lib/common/profile_obj.c
ifranzki: Agreed. Lets keep it as is for now.

If a token is supposed to contain these profile objects, and they are supposed to be found using C_FindObject, then the token should not only hinder an application from making changes to it, but it should also hinder an application from destroying/deleting such objects.  Otherwise one application could delete them and a second application would not find them anymore....
But that's a lot of guessing, so I would expect the PKCS #11 3.0 standard to specify such things more explicitly.

ifranzki (approved):

PR #324, opened 2020-10-13
Bumping up to version-3.15.0

p-steuer:


ChangeLog
```diff
@@ -1,4 +1,12 @@
-+ Opencryptoki 3.14
++ openCryptoki 3.15
+- common: conform to PKCS 11 3.0 Baseline Provider profile
+- CCA: support key wrapping
+- EP11: support m_ReEncryptSingle
```
ifranzki, line 4: C_IBM_ReencryptSingle is actually available for all tokens, and is of special interest for the secure key EP11 and CCA tokens.

ifranzki (approved):
With the minor comment on the changelog: LGTM

ChangeLog
```diff
@@ -1,4 +1,12 @@
...
+- SOFT: support ECC
+- p11sak tool: remove key command
```
jschmidb, line 6: a very minor comment: I would change "remove key command" to "add remove-key command"

p-steuer: good point. done.

ChangeLog
p-steuer: right, i changed that to also mention the name of our new vendor defined interface and the actual interface name that is exported rather than the EP11 function.

jschmidb (approved):
LGTM

PR #335, opened 2020-10-28
Bumping up to version 3.15.1

p-steuer:


ifranzki (approved):
OK, compared to the commit for 3.15.0, you don't update the README.md and usr/sbin/pkcsslotd/opencryptoki.conf for 3.15.1, but that's OK as far as I can tell.

p-steuer:
Its intentional, we never updated those on patchlevel releases.

PR #336, opened 2020-10-30
Fixes and small enhancements to make Opencryptoki even more PKCS#11 conforming

ifranzki:


testcases/pkcs11/attribute.c
```diff
@@ -138,11 +184,164 @@ CK_RV do_TestAttributes(void)
     else
         testcase_pass("Successfully verified updated attributes.");
 
+    /* Try to update attributes of an object which has CKA_MODIFIABLE=FALSE */
+    testcase_new_assertion();
+
+    /* create a public key object with CKA_MODIFIABLE=FALSE*/
+    rc = funcs->C_CreateObject(session, pub_template_no_modify, 7,
+                               &obj_handle_no_mod);
```
jschmidb, line 192: We could introduce a length variable, like
CK_ULONG pub_template_no_modify_len = sizeof(pub_template_no_modify) / sizeof(CK_ATTRIBUTE);
and use it here instead of explicitly coding 7.

testcases/pkcs11/copyobjects.c
```diff
@@ -99,6 +117,18 @@ CK_RV do_CopyObjects(void)
...
+    if (rc != CKR_OK) {
+        testcase_error("C_CreateObject() rc = %s", p11_get_ckr(rc));
+        goto testcase_cleanup;
+    }
+
+    rc = funcs->C_CreateObject(session, aes_tmpl_copy, 5, &keyobj_copy);
+    if (rc != CKR_OK) {
+        testcase_error("C_CreateObject() rc = %s", p11_get_ckr(rc));
+        goto testcase_cleanup;
+    }
```
jschmidb, line 130: Same as above, specifying the template length ...

testcases/pkcs11/destroyobjects.c
```diff
@@ -197,6 +212,29 @@ CK_RV do_DestroyObjects(void)
 
     testcase_pass("All objects were deleted.");
 
+    testcase_new_assertion();
+    /* Create a key object with CKA_DESTROYABLE=FALSE */
+    rc = funcs->C_CreateObject(session, aes_tmpl_no_destroy, 5,
+                               &keyobj_no_destroy);
```
jschmidb, line 218: Same as above, specifying the template length ...

usr/lib/common/cert.c
```diff
@@ -60,26 +60,154 @@ CK_RV cert_check_required_attributes(TEMPLATE *tmpl, CK_ULONG mode)
...
 
 // cert_validate_attribute()
 //
-CK_RV cert_validate_attribute(TEMPLATE *tmpl, CK_ATTRIBUTE *attr,
+CK_RV cert_validate_attribute(STDLL_TokData_t *tokdata,
+                              TEMPLATE *tmpl, CK_ATTRIBUTE *attr,
                               CK_ULONG mode)
 {
     CK_CERTIFICATE_TYPE type;
+    CK_CERTIFICATE_CATEGORY categoy;
```
jschmidb, line 157: categoy -> category

usr/lib/common/dp_obj.c
```diff
@@ -32,14 +32,14 @@
...
+    CK_RV rc;
 
-    found = template_attribute_find(tmpl, CKA_KEY_TYPE, &attr);
-    if (!found) {
+    rc = template_attribute_get_ulong(tmpl, CKA_KEY_TYPE, &val);
+    if (rc != CKR_OK) {
         if (mode == MODE_CREATE) {
-            TRACE_ERROR("%s\n", ock_err(ERR_TEMPLATE_INCOMPLETE));
-            return CKR_TEMPLATE_INCOMPLETE;
+            TRACE_ERROR("Could not find CKA_KEY_TYPE \n");
```
jschmidb, line 41: remove trailing blank?

usr/lib/common/dp_obj.c
```diff
@@ -132,15 +134,15 @@ CK_RV dp_x9dh_check_required_attributes(TEMPLATE *tmpl, CK_ULONG mode)
             return CKR_TEMPLATE_INCOMPLETE;
         }
     } else if (mode == MODE_KEYGEN) {
-        found = template_attribute_find(tmpl, CKA_PRIME_BITS, &attr);
-        if (!found) {
-            TRACE_ERROR("%s\n", ock_err(ERR_TEMPLATE_INCOMPLETE));
-            return CKR_TEMPLATE_INCOMPLETE;
+        rc = template_attribute_get_ulong(tmpl, CKA_PRIME_BITS, &val);
+        if (rc != CKR_OK) {
+            TRACE_ERROR("Could not find CKA_PRIME_BITS \n");
```
jschmidb, line 139: remove trailing blank?

usr/lib/common/dp_obj.c
```diff
@@ -132,15 +134,15 @@ CK_RV dp_x9dh_check_required_attributes(TEMPLATE *tmpl, CK_ULONG mode)
...
+            return rc;
         }
-        found = template_attribute_find(tmpl, CKA_SUBPRIME_BITS, &attr);
-        if (!found) {
-            TRACE_ERROR("%s\n", ock_err(ERR_TEMPLATE_INCOMPLETE));
-            return CKR_TEMPLATE_INCOMPLETE;
+        rc = template_attribute_get_ulong(tmpl, CKA_SUBPRIME_BITS, &val);
+        if (rc != CKR_OK) {
+            TRACE_ERROR("Could not find CKA_SUBPRIME_BITS \n");
```
jschmidb, line 144: remove trailing blank?

usr/lib/common/hwf_obj.c
```diff
@@ -32,14 +32,14 @@
...
+    CK_RV rc;
 
-    found = template_attribute_find(tmpl, CKA_HW_FEATURE_TYPE, &attr);
-    if (!found) {
+    rc = template_attribute_get_ulong(tmpl, CKA_HW_FEATURE_TYPE, &val);
+    if (rc != CKR_OK) {
         if (mode == MODE_CREATE) {
-            TRACE_ERROR("%s\n", ock_err(ERR_TEMPLATE_INCOMPLETE));
-            return CKR_TEMPLATE_INCOMPLETE;
+            TRACE_ERROR("Could not find CKA_HW_FEATURE_TYPE \n");
```
jschmidb, line 41: remove trailing blank?

usr/lib/common/hwf_obj.c
```diff
@@ -74,16 +75,16 @@ CK_RV counter_check_required_attributes(TEMPLATE *tmpl, CK_ULONG mode)
             return CKR_TEMPLATE_INCOMPLETE;
         }
 
-        found = template_attribute_find(tmpl, CKA_HAS_RESET, &attr);
-        if (!found) {
-            TRACE_ERROR("%s\n", ock_err(ERR_TEMPLATE_INCOMPLETE));
-            return CKR_TEMPLATE_INCOMPLETE;
+        rc = template_attribute_get_bool(tmpl, CKA_HAS_RESET, &val);
+        if (rc != CKR_OK) {
+            TRACE_ERROR("Could not find CKA_HAS_RESET \n");
```
jschmidb, line 80: remove trailing blank?

usr/lib/common/hwf_obj.c
```diff
@@ -74,16 +75,16 @@ CK_RV counter_check_required_attributes(TEMPLATE *tmpl, CK_ULONG mode)
...
+            return rc;
         }
 
-        found = template_attribute_find(tmpl, CKA_RESET_ON_INIT, &attr);
-        if (!found) {
-            TRACE_ERROR("%s\n", ock_err(ERR_TEMPLATE_INCOMPLETE));
-            return CKR_TEMPLATE_INCOMPLETE;
+        rc = template_attribute_get_bool(tmpl, CKA_RESET_ON_INIT, &val);
+        if (rc != CKR_OK) {
+            TRACE_ERROR("Could not find CKA_RESET_ON_INIT \n");
```
jschmidb, line 86: remove trailing blank?

usr/lib/common/key.c
```diff
@@ -140,14 +140,14 @@
...
+    CK_RV rc;
 
-    found = template_attribute_find(tmpl, CKA_KEY_TYPE, &attr);
-    if (!found) {
+    rc = template_attribute_get_ulong(tmpl, CKA_KEY_TYPE, &val);
+    if (rc != CKR_OK) {
         if (mode == MODE_CREATE) {
-            TRACE_ERROR("%s\n", ock_err(ERR_TEMPLATE_INCOMPLETE));
-            return CKR_TEMPLATE_INCOMPLETE;
+            TRACE_ERROR("Could not find CKA_KEY_TYPE \n");
```
jschmidb, line 149: remove trailing blank?

usr/lib/common/key.c
```diff
@@ -1135,11 +1353,11 @@ CK_RV rsa_publ_check_required_attributes(TEMPLATE *tmpl, CK_ULONG mode)
...
     }
 
-    found = template_attribute_find(tmpl, CKA_MODULUS_BITS, &attr);
-    if (!found) {
+    rc = template_attribute_get_ulong(tmpl, CKA_MODULUS_BITS, &val);
+    if (rc != CKR_OK) {
         if (mode == MODE_KEYGEN) {
-            TRACE_ERROR("%s\n", ock_err(ERR_TEMPLATE_INCOMPLETE));
-            return CKR_TEMPLATE_INCOMPLETE;
+            TRACE_ERROR("Could not find CKA_MODULUS_BITS \n");
```
jschmidb, line 1359: remove trailing blank?

usr/lib/common/mech_dsa.c
```diff
@@ -39,31 +39,26 @@ CK_RV dsa_sign(STDLL_TokData_t *tokdata,
...
     CK_OBJECT_CLASS class;
-    CK_BBOOL flag;
     CK_RV rc;
 
     UNUSED(sess);
 
     rc = object_mgr_find_in_map1(tokdata, ctx->key, &key_obj, READ_LOCK);
     if (rc != CKR_OK) {
-        TRACE_ERROR("Failed to acquire key from specified handle");
+        TRACE_ERROR("Failed to acquire key from specified handl.\n");
```
jschmidb, line 50: handle.

usr/lib/common/mech_ec.c
```diff
@@ -1376,3 +1424,98 @@ CK_RV ec_uncompress_public_key(CK_BYTE *curve, CK_ULONG curve_len,
...
+done:
+    if (pub_key)
+        EC_POINT_free(pub_key);
+    if (eckey)
+        EC_KEY_free(eckey);
+    BN_clear_free(bn_x);
+    BN_clear_free(bn_y);
+    BN_clear_free(bn_d);
+    if (ec_point != NULL)
+        free(ec_point);
```
jschmidb, line 1518: free() internally checks for null

usr/lib/common/mech_ssl3.c
```diff
@@ -991,28 +1007,41 @@ CK_RV ssl3_master_key_derive(STDLL_TokData_t *tokdata,
...
+    }
+
+    base_key_len = attr->ulValueLen;
+    base_key_value = attr->pValue;
+
+    if (base_key_len != 48) {
+        TRACE_ERROR("The base key's length is not 48.\n");
+        rc = CKR_KEY_FUNCTION_NOT_PERMITTED;
+        goto error;
+    }
```
jschmidb, line 1044: Hm ... from where do the 48 bytes come from? Can we possibly write out more detailed info?

usr/lib/common/mech_ssl3.c
```diff
@@ -1022,33 +1051,40 @@ CK_RV ssl3_master_key_derive(STDLL_TokData_t *tokdata,
...
+
+    rc = get_ulong_attribute_by_type(pTemplate, ulCount, CKA_VALUE_LEN,
+                                     &value_len);
+    if (rc == CKR_ATTRIBUTE_VALUE_INVALID) {
+        TRACE_ERROR("%s\n", ock_err(ERR_ATTRIBUTE_VALUE_INVALID));
+        return CKR_ATTRIBUTE_VALUE_INVALID;
+    }
+    if (rc == CKR_OK && value_len != 48) {
+        TRACE_ERROR("The derived key's length is not 48.\n");
```
jschmidb, line 1085: ... same as above

usr/lib/common/obj_mgr.c
```diff
@@ -711,6 +762,9 @@ CK_RV object_mgr_create_final(STDLL_TokData_t *tokdata,
         }
     }
 
+    if (rc == CKR_OK)
+        TRACE_DEVEL("Object Created: handle: %lu\n", *handle);
```
jschmidb, line 766: Object created ...

usr/lib/common/mech_ssl3.c
ifranzki: I don't know...  It was that way before, I did not dig into it....

usr/lib/api/api_interface.c
```diff
@@ -3553,7 +3553,7 @@ CK_RV C_SignFinal(CK_SESSION_HANDLE hSession,
     STDLL_FcnList_t *fcn;
     ST_SESSION_T rSession;
 
-    TRACE_INFO("C_SignEncryptUpdate\n");
+    TRACE_INFO("C_SignFinal\n");
```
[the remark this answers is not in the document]
ifranzki: yes, good idea. but this would be a commit on its own changing all the similar places

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -3222,7 +3222,8 @@ static CK_BBOOL ep11tok_ec_curve_supported(STDLL_TokData_t *tokdata,
     key_obj = NULL;
 
     for (i = 0; i < NUMEC; i++) {
-        if ((memcmp(attr->pValue, der_ec_supported[i].data,
+        if ((CK_ULONG)der_ec_supported[i].data == attr->ulValueLen &&
+            (memcmp(attr->pValue, der_ec_supported[i].data,
```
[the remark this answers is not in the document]
ifranzki: right, this should be 'der_ec_supported[i].data_size == attr->ulValueLen' of course....

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -4247,14 +4268,12 @@ CK_RV token_specific_generic_secret_key_gen(STDLL_TokData_t * tokdata,
...
+    rc = template_attribute_get_ulong(template, CKA_VALUE_LEN, &keylength);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("CKA_VALUE_LEN missing in (HMAC) key template\n");
+        return rc;
     }
 
-    keylength = *(CK_ULONG *) attr->pValue;
-
     /* HMAC key length needs to be 80-2048 bits */
     if (((8 * keylength) < 80) || ((8 * keylength) > 2048)) {
```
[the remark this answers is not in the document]
ifranzki: good catch!

usr/lib/common/attributes.c
```diff
@@ -121,6 +125,49 @@ CK_ATTRIBUTE_PTR get_attribute_by_type(CK_ATTRIBUTE_PTR attrs,
...
+                                   CK_ULONG *value)
+{
+    CK_ATTRIBUTE_PTR attr;
+
+    attr = get_attribute_by_type(attrs, attrs_len, type);
+    if (attr == NULL)
+        return CKR_TEMPLATE_INCOMPLETE;
+
+    if (attr->ulValueLen != sizeof(CK_ULONG) || attr->pValue == NULL) {
+        TRACE_ERROR("%s\n", ock_err(ERR_ATTRIBUTE_VALUE_INVALID));
```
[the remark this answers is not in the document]
ifranzki: yep, will do.

usr/lib/common/h_extern.h
```diff
@@ -2244,6 +2247,13 @@ CK_RV template_add_default_attributes(TEMPLATE *tmpl,
 CK_BBOOL template_attribute_find(TEMPLATE *tmpl,
                                  CK_ATTRIBUTE_TYPE type, CK_ATTRIBUTE **attr);
 
+CK_RV template_attribute_get_ulong(TEMPLATE *tmpl, CK_ATTRIBUTE_TYPE type,
```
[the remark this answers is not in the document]
ifranzki: well, because the other already existing functions are called template_attribute_xxx (find, find_multible), so I kept this naming scheme.

usr/lib/common/key.c
```diff
@@ -2580,24 +2692,26 @@ CK_RV ecdsa_priv_wrap_get_data(TEMPLATE *tmpl,
...
+    rc = template_attribute_get_non_empty(tmpl, CKA_VALUE, &point);
+    if (rc != CKR_OK) {
         TRACE_ERROR("Could not find EC Point for the key.\n");
-        return CKR_FUNCTION_FAILED;
+        return rc;
     }
 
     /* check if optional public-key part was defined */
-    template_attribute_find(tmpl, CKA_EC_POINT, &pubkey);
+    template_attribute_get_non_empty(tmpl, CKA_EC_POINT, &pubkey);
```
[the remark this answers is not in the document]
ifranzki: Yes template_attribute_get_non_empty should set attr to NULL if returning an error. Will fix this.

usr/lib/common/key_mgr.c
```diff
@@ -70,19 +70,23 @@ CK_RV key_mgr_generate_key(STDLL_TokData_t *tokdata,
...
-                return CKR_TEMPLATE_INCONSISTENT;
-            }
-        }
-
-        if (pTemplate[i].type == CKA_KEY_TYPE)
-            subclass = *(CK_ULONG *) pTemplate[i].pValue;
+    rc = get_ulong_attribute_by_type(pTemplate, ulCount, CKA_CLASS,
+                                     &keyclass);
+    if (rc == CKR_ATTRIBUTE_VALUE_INVALID) {
+        TRACE_ERROR("%s\n", ock_err(ERR_ATTRIBUTE_VALUE_INVALID));
```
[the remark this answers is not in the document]
ifranzki: From the trace message in get_ulong_attribute_by_type you can't tell from where get_ulong_attribute_by_type as called. This trace mesage has a line number, so you know exactly where the error was detected.

usr/lib/common/key_mgr.c
```diff
@@ -194,46 +198,37 @@ CK_RV key_mgr_generate_key(STDLL_TokData_t *tokdata,
...
         goto error;
     }
 
+    rc = build_attribute(CKA_ALWAYS_SENSITIVE, &flag, sizeof(CK_BBOOL),
+                         &new_attr);
+    if (rc != CKR_OK) {
+        TRACE_DEVEL("build attribute failed.\n");
+        goto error;
+    }
+    template_update_attribute(key_obj->template, new_attr);
```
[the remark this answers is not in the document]
ifranzki: template_update_attribute  can only fail if one of its arguments is NULL, which is the callers responsibility. There are no other error cases. 
However, I agree that this is not really nice the way is is. I would also prefer to have return code checking there. Unfortunately, there are lots of places where template_update_attribute is used without error checking....

usr/lib/common/mech_ssl3.c
```diff
@@ -1165,26 +1170,21 @@ CK_RV ssl3_master_key_derive(STDLL_TokData_t *tokdata,
...
 
-    flag = *(CK_BBOOL *) attr->pValue;
     if (flag == TRUE) {
-        rc = template_attribute_find(derived_key_obj->template, CKA_EXTRACTABLE,
-                                     &attr);
-        if (rc == FALSE) {
+        rc = template_attribute_get_bool(derived_key_obj->template,
+                                         CKA_EXTRACTABLE, &flag);
+        if (rc != CKR_OK) {
             TRACE_DEVEL("Failed to build CKA_EXTRACTABLE attribute.\n");
```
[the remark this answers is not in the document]
ifranzki: good catch. Looks like you are really reading all the code :-)

usr/lib/common/template.c
```diff
@@ -276,6 +282,82 @@ CK_BBOOL template_attribute_find(TEMPLATE *tmpl, CK_ATTRIBUTE_TYPE type,
...
+ * invalid size.
+ */
+CK_RV template_attribute_get_non_empty(TEMPLATE *tmpl, CK_ATTRIBUTE_TYPE type,
+                                       CK_ATTRIBUTE **attr)
+{
+    CK_BBOOL found;
+
+    found = template_attribute_find(tmpl, type, attr);
+    if (!found || *attr == NULL)
+        return CKR_TEMPLATE_INCOMPLETE;
```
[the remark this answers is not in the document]
ifranzki: yep, fixed.

usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -2005,9 +2014,11 @@ static CK_RV os_specific_rsa_encrypt(STDLL_TokData_t *tokdata,
     CK_RV rc;
 
     /* mech_sra.c:ckm_rsa_encrypt accepts only CKO_PUBLIC_KEY */
-    template_attribute_find(key_obj->template, CKA_MODULUS, &modulus);
-    template_attribute_find(key_obj->template, CKA_MODULUS_BITS, &mod_bits);
-    template_attribute_find(key_obj->template, CKA_PUBLIC_EXPONENT, &pub_exp);
+    template_attribute_get_non_empty(key_obj->template, CKA_MODULUS, &modulus);
+    template_attribute_get_non_empty(key_obj->template, CKA_MODULUS_BITS,
+                                     &mod_bits);
+    template_attribute_get_non_empty(key_obj->template, CKA_PUBLIC_EXPONENT,
```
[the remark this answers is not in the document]
ifranzki: yes, fixed in template_attribute_get_non_empty

usr/lib/soft_stdll/soft_specific.c
```diff
@@ -844,26 +849,23 @@ static void *rsa_convert_private_key(OBJECT *key_obj)
...
+    template_attribute_get_non_empty(key_obj->template, CKA_PRIVATE_EXPONENT,
+                                     &priv_exp);
+    template_attribute_get_non_empty(key_obj->template, CKA_PRIME_1, &prime1);
+    template_attribute_get_non_empty(key_obj->template, CKA_PRIME_2, &prime2);
+    template_attribute_get_non_empty(key_obj->template, CKA_EXPONENT_1, &exp1);
+    template_attribute_get_non_empty(key_obj->template, CKA_EXPONENT_2,&exp2);
+    template_attribute_get_non_empty(key_obj->template, CKA_COEFFICIENT,
+                                     &coeff);
 
     if (!prime2 && !modulus) {
```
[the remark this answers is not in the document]
ifranzki: yes, fixed in template_attribute_get_non_empty

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -1237,11 +1238,31 @@ static CK_RV rawkey_2_blob(STDLL_TokData_t * tokdata, SESSION * sess,
         TRACE_INFO("%s unwrap blen=%zd rc=0x%lx\n", __func__, *blen, rc);
     }
 
+    if (cslen >= 3) {
```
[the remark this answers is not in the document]
ifranzki: good catch.

usr/lib/common/key_mgr.c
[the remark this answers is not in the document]
juergenchrist: Looking at the code of template_update_attribute, there is another way it can fail: memory error.  In that case, the function forgets about all attributes in the template and still returns success.  See bug #339 which might be more theoretic, but nevertheless I think we should fix that.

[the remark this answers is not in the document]
ifranzki: Yes, agreed. We should fix template_update_attribute to return an error in case of a memory error (and not destroy the template). Once we have changed template_update_attribute we must also update all callers of template_update_attribute and add proper return code checking.
But this will be a separate commit in any case.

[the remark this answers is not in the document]
ifranzki: I have just pushed a commit for fixing the template_update_attribute problem (https://github.com/opencryptoki/opencryptoki/issues/339) and also all its callers to check the return code:
https://github.com/opencryptoki/opencryptoki/pull/336/commits/136570f2708d89e1de47a659587b137319c02a0f
When you look at the size of commit you probably understand why I wanted to fix this in a separate commit... ;-)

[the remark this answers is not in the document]
juergenchrist: Nice work.  Sorry, there was a misunderstanding.  I did not want to suggest to fix this in this commit.  I just saw this and commented even though it has nothing to do with this commit...

usr/lib/common/key.c
```diff
@@ -1910,44 +2253,118 @@ CK_RV rsa_priv_unwrap(TEMPLATE *tmpl, CK_BYTE *data, CK_ULONG total_length)
...
+    if (rc != CKR_OK) {
+        TRACE_DEVEL("ber_decode_RSAPublicKey failed\n");
+        return rc;
     }
 
-    p11_attribute_trim(modulus);
-    p11_attribute_trim(publ_exp);
-
+    p11_attribute_trim(modulus);
+    p11_attribute_trim(publ_exp);
```
juergenchrist, line 2344: Just out of curiosity: What is the diff here?  Whitespace?  Looks the same to me...

usr/lib/common/hwf_obj.c
```diff
@@ -249,9 +256,34 @@ CK_RV counter_set_default_attributes(TEMPLATE *tmpl, CK_ULONG mode)
         (CK_BYTE *) resetoninit_attr + sizeof(CK_ATTRIBUTE);
     *(CK_BBOOL *) resetoninit_attr->pValue = FALSE;
 
-    template_update_attribute(tmpl, value_attr);
-    template_update_attribute(tmpl, hasreset_attr);
-    template_update_attribute(tmpl, resetoninit_attr);
+    rc = template_update_attribute(tmpl, value_attr);
+    if (rc != CKR_OK) {
+        TRACE_DEVEL("template_update_attribute failed\n");
+        return rc;
```
juergenchrist, line 262: memory leak on `value_attr`.  Use `goto error` instead.

usr/lib/common/hwf_obj.c
```diff
@@ -249,9 +256,34 @@ CK_RV counter_set_default_attributes(TEMPLATE *tmpl, CK_ULONG mode)
...
+    }
+    value_attr = NULL;
+    rc = template_update_attribute(tmpl, hasreset_attr);
+    if (rc != CKR_OK) {
+        TRACE_DEVEL("template_update_attribute failed\n");
+        return rc;
```
juergenchrist, line 268: memory leak on `hasreset_attr`.  Use `goto error` instead.

usr/lib/common/hwf_obj.c
```diff
@@ -249,9 +256,34 @@ CK_RV counter_set_default_attributes(TEMPLATE *tmpl, CK_ULONG mode)
...
+    }
+    hasreset_attr = NULL;
+    rc = template_update_attribute(tmpl, resetoninit_attr);
+    if (rc != CKR_OK) {
+        TRACE_DEVEL("template_update_attribute failed\n");
+        return rc;
```
juergenchrist, line 274: memory leak on `resetoninit_attr`.  Use `goto error` instead.

usr/lib/common/mech_aes.c
```diff
@@ -3460,16 +3456,44 @@ CK_RV ckm_aes_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl)
...
     *(CK_BBOOL *) local_attr->pValue = TRUE;
 
-    template_update_attribute(tmpl, value_attr);
-    template_update_attribute(tmpl, key_type_attr);
-    template_update_attribute(tmpl, class_attr);
-    template_update_attribute(tmpl, local_attr);
+    rc = template_update_attribute(tmpl, value_attr);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("template_update_attribute failed\n");
+        return rc;
```
juergenchrist, line 3462: memory leak.  Use `goto err`.

usr/lib/common/mech_aes.c
```diff
@@ -3460,16 +3456,44 @@ CK_RV ckm_aes_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl)
...
+    }
+    value_attr = NULL;
+    rc = template_update_attribute(tmpl, key_type_attr);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("template_update_attribute failed\n");
+        return rc;
```
juergenchrist, line 3468: memory leak.  Use `goto err`.

usr/lib/common/mech_aes.c
```diff
@@ -3460,16 +3456,44 @@ CK_RV ckm_aes_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl)
...
+    }
+    key_type_attr = NULL;
+    rc = template_update_attribute(tmpl, class_attr);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("template_update_attribute failed\n");
+        return rc;
```
juergenchrist, line 3474: memory leak.  Use `goto err`.

usr/lib/common/mech_aes.c
```diff
@@ -3460,16 +3456,44 @@ CK_RV ckm_aes_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl)
...
+    }
+    class_attr = NULL;
+    rc = template_update_attribute(tmpl, local_attr);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("template_update_attribute failed\n");
+        return rc;
```
juergenchrist, line 3480: memory leak.  Use `goto err`.

usr/lib/common/mech_des.c
```diff
@@ -1308,16 +1304,44 @@ CK_RV ckm_des_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl)
...
     *(CK_BBOOL *) local_attr->pValue = TRUE;
 
-    template_update_attribute(tmpl, value_attr);
-    template_update_attribute(tmpl, key_type_attr);
-    template_update_attribute(tmpl, class_attr);
-    template_update_attribute(tmpl, local_attr);
+    rc = template_update_attribute(tmpl, value_attr);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("template_update_attribute failed\n");
+        return rc;
```
juergenchrist, line 1310: memory leak.  Use `goto err`.

usr/lib/common/mech_des.c
```diff
@@ -1308,16 +1304,44 @@ CK_RV ckm_des_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl)
...
+    }
+    value_attr = NULL;
+    rc = template_update_attribute(tmpl, key_type_attr);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("template_update_attribute failed\n");
+        return rc;
```
juergenchrist, line 1316: memory leak.  Use `goto err`.

usr/lib/common/mech_des.c
```diff
@@ -1308,16 +1304,44 @@ CK_RV ckm_des_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl)
...
+    }
+    key_type_attr = NULL;
+    rc = template_update_attribute(tmpl, class_attr);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("template_update_attribute failed\n");
+        return rc;
```
juergenchrist, line 1322: memory leak.  Use `goto err`.

usr/lib/common/mech_des.c
```diff
@@ -1308,16 +1304,44 @@ CK_RV ckm_des_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl)
...
+    }
+    class_attr = NULL;
+    rc = template_update_attribute(tmpl, local_attr);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("template_update_attribute failed\n");
+        return rc;
```
juergenchrist, line 1328: memory leak.  Use `goto err`.

usr/lib/common/mech_des.c
```diff
@@ -1388,12 +1404,44 @@ CK_RV ckm_cdmf_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl)
...
     *(CK_BBOOL *) local_attr->pValue = TRUE;
 
-    template_update_attribute(tmpl, value_attr);
-    template_update_attribute(tmpl, key_type_attr);
-    template_update_attribute(tmpl, class_attr);
-    template_update_attribute(tmpl, local_attr);
+    rc = template_update_attribute(tmpl, value_attr);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("template_update_attribute failed\n");
+        return rc;
```
juergenchrist, line 1410: memory leak.  Use `goto error`.

usr/lib/common/mech_des.c
```diff
@@ -1388,12 +1404,44 @@ CK_RV ckm_cdmf_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl)
...
+    }
+    value_attr = NULL;
+    rc = template_update_attribute(tmpl, key_type_attr);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("template_update_attribute failed\n");
+        return rc;
```
juergenchrist, line 1416: memory leak.  Use `goto error`.

usr/lib/common/mech_des.c
```diff
@@ -1388,12 +1404,44 @@ CK_RV ckm_cdmf_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl)
...
+    rc = template_update_attribute(tmpl, class_attr);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("template_update_attribute failed\n");
+        return rc;
+    }
+    class_attr = NULL;
+    rc = template_update_attribute(tmpl, local_attr);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("template_update_attribute failed\n");
+        return rc;
```
juergenchrist, line 1428: memory leak.  Use `goto error`.

usr/lib/common/mech_des3.c
```diff
@@ -2796,16 +2792,44 @@ CK_RV ckm_des3_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl)
...
     *(CK_BBOOL *) local_attr->pValue = TRUE;
 
-    template_update_attribute(tmpl, value_attr);
-    template_update_attribute(tmpl, key_type_attr);
-    template_update_attribute(tmpl, class_attr);
-    template_update_attribute(tmpl, local_attr);
+    rc = template_update_attribute(tmpl, value_attr);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("template_update_attribute failed\n");
+        return rc;
```
juergenchrist, line 2798: memory leak.  Use `goto error`.

usr/lib/common/mech_des3.c
```diff
@@ -2796,16 +2792,44 @@ CK_RV ckm_des3_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl)
...
+    }
+    value_attr = NULL;
+    rc = template_update_attribute(tmpl, key_type_attr);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("template_update_attribute failed\n");
+        return rc;
```
juergenchrist, line 2804: memory leak.  Use `goto error`.

usr/lib/common/mech_des3.c
```diff
@@ -2796,16 +2792,44 @@ CK_RV ckm_des3_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl)
...
+    }
+    key_type_attr = NULL;
+    rc = template_update_attribute(tmpl, class_attr);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("template_update_attribute failed\n");
+        return rc;
```
juergenchrist, line 2810: memory leak.  Use `goto error`.

usr/lib/common/mech_des3.c
```diff
@@ -2796,16 +2792,44 @@ CK_RV ckm_des3_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl)
...
+    }
+    class_attr = NULL;
+    rc = template_update_attribute(tmpl, local_attr);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("template_update_attribute failed\n");
+        return rc;
```
juergenchrist, line 2816: memory leak.  Use `goto error`.

juergenchrist (commented):
Great work!

usr/lib/common/key.c
ifranzki: hm, in the current version on github this lines don't show any difference.....

usr/lib/common/mech_des3.c
ifranzki: good catch! Thanks for reading so carefully!
Will fix this and all the others as well.

ifranzki:
I should not have completed the development for this PR. 
The following commits were added after you have started to review: 
- TESTCASES: Test CKA_WRAP_TEMPLATE and CKA_UNWRAP_TEMPLATE
- Support CKA_UNWRAP_TEMPLATE attribute
- Support CKA_WRAP_TEMPLATE attribute
- Add support for attribute-array attributes
- API: Move certain parameter checking into the token DLLs
- Don't allow to set attribute CKA_ALWAYS_AUTHENTICATE to TRUE
- Add support for CKA_ALLOWED_MECHANISMS to restrict key usage
- TESTCASE: Test CKM_ECDH1_DERIVE without CKA_VALUE_LEN
- Key derivation via CKM_ECDH1_DERIVE: Make CKA_VALUE_LEN optional
- Correct error handling of C_Sign/Verify/Update/Final
- SOFT: Enable CKM_SHAx_RSA_PKCS mechanisms
- EP11: Use EP11 host library header files as-is

The following commit was added within the first few commits to fix a bug that was revealed with "* Fix sign/verify of empty message": 
- Fix RSA signature/decryption of empty message with CKM_RSA_PKCS

The commits you have reviewed so far might have been changed to fix findings by you or me.

usr/lib/common/template.c
```diff
@@ -633,20 +673,128 @@ CK_RV template_copy(TEMPLATE *dest, TEMPLATE *src)
...
+                element_32.pValue = 0x00;
+
+                if (flatten_ulong_attribute_as_ulong32(attrs[i].type) &&
+                    attrs[i].ulValueLen != 0) {
+                    element_32.ulValueLen = sizeof(CK_ULONG_32);
+
+                    memcpy(ptr, &attr_32, sizeof(CK_ATTRIBUTE_32));
+                    ptr += sizeof(CK_ATTRIBUTE_32);
+
+                    Val_32 = (CK_ULONG_32) *((CK_ULONG *) attrs[i].pValue);
```
juergenchrist, line 765: I don't think this goes through with ASAN checks (alignment of CK_ULONG vs. alignment of CK_BYTE).  I think the correct way would be to memcpy first into a properly aligned CK_ULONG, then assign to a properly aligned CK_ULONG_32, and then memcpy into ptr.

usr/lib/common/mech_rsa.c
```diff
@@ -281,7 +292,7 @@ CK_RV rsa_parse_block(CK_BYTE *in_data,
         return CKR_ENCRYPTED_DATA_INVALID;
     }
 
-    if (in_data_len <= i) {
+    if (in_data_len < i) {
```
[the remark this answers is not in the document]
ifranzki: OK, will change.

usr/lib/common/key_mgr.c
```diff
@@ -584,6 +584,20 @@ CK_RV key_mgr_wrap_key(STDLL_TokData_t *tokdata,
...
+    /* Is a wrapping key with CKA_TRUSTED = CK_TRUE required? */
+    if (template_attribute_find(key_obj->template, CKA_WRAP_WITH_TRUSTED,
+                                &attr) &&
+        attr != NULL && *(CK_BBOOL *)attr->pValue == TRUE) {
+
+        if (template_attribute_find(wrapping_key_obj->template, CKA_TRUSTED,
+                                    &attr) &&
+            attr != NULL && *(CK_BBOOL *)attr->pValue == FALSE) {
+            TRACE_ERROR("%s\n", ock_err(ERR_KEY_NOT_WRAPPABLE));
+            rc = CKR_KEY_NOT_WRAPPABLE;
```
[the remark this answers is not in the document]
ifranzki: There is no error CKR_WRAPPING_KEY_INVALID, only CKR_WRAPPING_KEY_HANDLE_INVALID, which doe snot really fit. So lets keep it as is (CKR_KEY_NOT_WRAPPABLE).

usr/lib/common/obj_mgr.c
```diff
@@ -732,6 +739,13 @@ CK_RV object_mgr_destroy_object(STDLL_TokData_t *tokdata,
         return CKR_OBJECT_HANDLE_INVALID;
     }
 
+    if (!object_is_destroyable(o)) {
+        TRACE_ERROR("Object is not destroyable\n");
+        object_put(tokdata, o, TRUE);
+        o = NULL;
```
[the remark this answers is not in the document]
ifranzki: I know, but since object_put decreases the ref count of the object, and the object might have been deleted if the ref count has reached zero, I use this as an indication that o should not be accessed any more.  You will find this an many places....

testcases/pkcs11/destroyobjects.c
```diff
@@ -197,6 +212,29 @@ CK_RV do_DestroyObjects(void)
...
+    }
+
+    /*
+     * Try to delete the non-destroyable object, should fail with
+     * CKR_ACTION_PROHIBITED
+     */
+    rc = funcs->C_DestroyObject(session, keyobj_no_destroy);
+    if (rc == CKR_ACTION_PROHIBITED)
+        testcase_pass("C_DestroyObject() did not delete the object. rc = %s "
+                "(as expetded)", p11_get_ckr(rc));
```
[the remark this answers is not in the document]
ifranzki: good point, will add code to remove the object after setting CKA_DESTROYABLE  to true.

usr/lib/common/key.c
```diff
@@ -1658,6 +1711,13 @@ CK_RV secret_key_validate_attribute(STDLL_TokData_t *tokdata, TEMPLATE *tmpl,
             return CKR_ATTRIBUTE_READ_ONLY;
         }
         return CKR_OK;
+    case CKA_WRAP_TEMPLATE:
+        if (attr->ulValueLen == 0|| attr->pValue == NULL ||
```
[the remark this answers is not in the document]
ifranzki: yes, right, will change.

usr/lib/common/attributes.c
```diff
@@ -57,59 +78,90 @@ void free_attribute_array(CK_ATTRIBUTE_PTR attrs, CK_ULONG attrs_len)
...
 
-    /* Allocate the new array */
-    dest_len = orig_len;
-    dest = malloc(dest_len * sizeof(*dest));
-    if (dest == NULL) {
-        TRACE_ERROR("%s\n", ock_err(ERR_HOST_MEMORY));
-        return CKR_HOST_MEMORY;
-    }
-    memset(dest, 0, dest_len);
+    memset(dest, 0, num_attrs);
```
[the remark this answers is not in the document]
ifranzki: right.

usr/lib/common/attributes.c
```diff
@@ -202,3 +269,61 @@ CK_RV add_to_attribute_array(CK_ATTRIBUTE_PTR *p_attrs,
...
+    if (a1->type != a2->type)
+         return FALSE;
+
+    if (a1->ulValueLen != a2->ulValueLen)
+         return FALSE;
+
+     if (a1->ulValueLen == 0)
+         return TRUE;
+
+     if ((a1->pValue == NULL && a1->ulValueLen > 0) ||
```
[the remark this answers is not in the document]
ifranzki: right, will change

usr/lib/common/p11util.c
```diff
@@ -207,6 +230,23 @@ CK_BBOOL is_attribute_defined(CK_ATTRIBUTE_TYPE type)
     return FALSE;
 }
 
+/*
+ * Returns true if the attribute is an attribute array type.
+ */
+CK_BBOOL is_attribute_attr_array(CK_ATTRIBUTE_TYPE type)
```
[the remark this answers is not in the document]
ifranzki: Well, is_attribute_type_array is not quite what the purpose is. 
There are attribute types (e.g. CKA_ALLOWED_MECHANISMS) that are an array, but not an array of attributes. 
To check if an attribute type is an array (any kind of array), you can simply check `type & CKF_ARRAY_ATTRIBUTE`, so no helper function needed for this.
But attribute types that are an array of attributes is a special case, and needs to be handled other than regular array attributes.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -973,10 +936,12 @@ static const_info_t ep11_mechanisms[] = {
     CONSTINFO(CKM_IBM_SHA512_256_KEY_DERIVATION),
     CONSTINFO(CKM_IBM_SHA512_224_KEY_DERIVATION),
     CONSTINFO(CKM_IBM_EC_C25519),
```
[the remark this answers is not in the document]
ifranzki: Yes, because it is used in ep11_get_mechanisms_by_name which itself is used in read_cp_filter_config_file to parse the CP filter config file. If there is an old config filter file that uses the old name, we still want to be able to parse it.

usr/lib/common/object.c
```diff
@@ -464,7 +522,7 @@ CK_RV object_get_attribute_values(OBJECT * obj,
             TRACE_ERROR("%s: %lx\n", ock_err(ERR_ATTRIBUTE_SENSITIVE),
                         pTemplate[i].type);
             rc = CKR_ATTRIBUTE_SENSITIVE;
```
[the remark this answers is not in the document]
ifranzki: PKCS#11 states in the description for C_GetAttributeValue:

> If case 1 applies to any of the requested attributes, then the call should return the value
> CKR_ATTRIBUTE_SENSITIVE. If case 2 applies to any of the requested attributes, then the call should
> return the value CKR_ATTRIBUTE_TYPE_INVALID. If case 5 applies to any of the requested attributes,
> then the call should return the value CKR_BUFFER_TOO_SMALL. **As usual, if more than one of these
> error codes is applicable, Cryptoki may return any of them**.

usr/lib/common/template.c
```diff
@@ -657,44 +805,43 @@ CK_RV template_flatten(TEMPLATE *tmpl, CK_BYTE *dest)
...
+
+                attr_32.ulValueLen = sizeof(CK_ULONG_32);
+
+                memcpy(ptr, &attr_32, sizeof(CK_ATTRIBUTE_32));
                 ptr += sizeof(CK_ATTRIBUTE_32);
 
-                pVal = (CK_ULONG *) attr->pValue;
-                Val = *pVal;
-                Val_32 = (CK_ULONG_32) Val;
+                Val_32 = (CK_ULONG_32) *((CK_ULONG *) attr->pValue);
```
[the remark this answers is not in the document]
ifranzki: Yes I do run with `-fstack-protector-all -D_FORTIFY_SOURCE=2 -fsanitize=address,undefined"` and it does pass.
Why would you think it should not pass? 
`*((CK_ULONG *) attr->pValue)` is used all over the place. The only difference here is that it is then casted to a CK_ULONG32 afterwards.

usr/lib/common/template.c
```diff
@@ -943,12 +1280,42 @@ CK_ULONG template_get_size(TEMPLATE *tmpl)
...
+static CK_ULONG attribute_get_compressed_size(CK_ATTRIBUTE_PTR attr)
+{
+    CK_ULONG size = 0, i, num_attrs;
+    CK_ATTRIBUTE_PTR attrs;
+
+    size += sizeof(CK_ATTRIBUTE_32);
+    if (flatten_ulong_attribute_as_ulong32(attr->type)
+        && attr->ulValueLen != 0) {
+        size += sizeof(CK_ULONG_32);
+    } else {
```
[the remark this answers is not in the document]
ifranzki: will change

p-steuer (approved):

p-steuer:
Ive just gone over what the commits do rather than all LOCs in detail. Looks good to me. If this one is approved also by @juergenchrist  or @jschmidb  , im going to merge this.

juergenchrist (approved):

PR #338, opened 2020-10-31
Make testcases-folder world writable for su tests

juergenchrist:
When running tests via "make installcheck" with "PKCS11_TEST_USER", that user
must be granted write access to the testcases folder since it will generate
the log-files in this folder.  Without this, ock_test.sh will not be able to
run successfully.


p-steuer (approved):
looks good to me.

PR #342, opened 2020-11-13
Add ci-installcheck make target

juergenchrist:
This target is supposed to contain all checks that should only run in the CI
but not on a "make installcheck".  Examples include token pin changes and
master key changes.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -199,17 +199,23 @@ endif
 
 
 if ENABLE_TESTCASES
+PKCS11_SO_PIN ?= 76543210
+PKCS11_USER_PIN ?= 01234567
+
 installcheck-local: all
 	killall -HUP pkcsslotd || true
 	@sbindir@/pkcsslotd
 	if test ! -z ${PKCS11_TEST_USER}; then				\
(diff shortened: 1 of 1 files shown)


p-steuer (approved):

ifranzki:
While you are are it, could you maybe also enhance the make file to not use the hardcoded PINS (12345678)? 
I usually habe initialized the tokens already with a different pin. Running `make installcheck` thus will lock out the pins.... 

ock_tests.sh already checks if the token is initialized, and only initializes them, if not.  But the make file unconditionally passes the hardcoded pins to it anyway, so it will try to run the test with that pin and will therefore lock out the pins if they don't match.

Can you maybe check if PKCS11_SO_PIN and PKCS11_USER_PIN is set, and only assign the default pins if not already set?

juergenchrist:
> Can you maybe check if PKCS11_SO_PIN and PKCS11_USER_PIN is set, and only assign the default pins if not already set?

Thankfully make has special syntax for this.  Please review the new change.  It should enable you to pass the values either via environment or even on the make command line.

ifranzki (approved):
Looks good, thanks!

PR #343, opened 2020-11-17
CCA: store only public token in IBM_OPAQUE attribute

hfreude:
Rework on RSA and EC keypair generate: Extract public cca token
part from private cca token and store only the public token
into CKA_IBM_OPAQUE attribute value of the public key object.

### DIFF
--- a/usr/lib/cca_stdll/cca_specific.c
+++ b/usr/lib/cca_stdll/cca_specific.c
@@ -238,6 +238,27 @@ static const MECH_LIST_ELEMENT cca_mech_list[] = {
 static const CK_ULONG cca_mech_list_len =
                         (sizeof(cca_mech_list) / sizeof(MECH_LIST_ELEMENT));
 
+
+/* Helper function: build attribute and update template */
+static CK_RV build_update_attribute(TEMPLATE * tmpl,
+                                    CK_ATTRIBUTE_TYPE type,
+                                    CK_BYTE * data, CK_ULONG data_len)
+{
+    CK_ATTRIBUTE *attr;
+    CK_RV rv;
+
+    if ((rv = build_attribute(type, data, data_len, &attr))) {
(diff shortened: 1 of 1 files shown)


usr/lib/cca_stdll/cca_specific.c
```diff
@@ -773,12 +775,21 @@ CK_RV token_create_keypair_object(TEMPLATE * tmpl, CK_ULONG tok_len,
...
+
     /* Add the opaque key object to the template */
-    if ((rv = build_attribute(CKA_IBM_OPAQUE, tok, tok_len, &opaque_key))) {
-        TRACE_DEVEL("build_attribute for opaque key failed. rv=0x%lx\n", rv);
+    if ((rv = build_attribute(CKA_IBM_OPAQUE, tok, tok_len, &opaque))) {
+        TRACE_DEVEL("build_attribute for opaque attribute failed. rv=0x%lx\n", rv);
         return rv;
     }
-    template_update_attribute(tmpl, opaque_key);
+    template_update_attribute(tmpl, opaque);
```
ifranzki, line 792: Use build_update_attribute() instead of this whole function? 

If you choose to use it the way you have it, then please add error checking for template_update_attribute as it is done in the commit in my PR: https://github.com/opencryptoki/opencryptoki/pull/336/commits/80e624f94615aa80ba751596f80f160791555cbc#diff-a14e5f28856cbf95c58b0212766dd738551a826b75496d26aa711ac06c9ac14aR830

hfreude: Ingo, you are right. I'll update the pull request and use the build_update_attribute() function. :-)

hfreude: pull request is updated - please note I added a whitespace commit in front of the patch because my editor
permanently wants to fix these issues and is driving me mad with mixing my code changes with whitespace updates.

hfreude:
drop - no way, needs total rework

PR #346, opened 2020-11-20
CCA: Support import of CCA secure key token objects

hfreude:
Please note this is only a heads up - the patch is still work in progress.
I only want to show what I intent to provide soon.

The secure key blob for all the keys used with the CCA token is
stored in an attribute CKA_IBM_OPAQUE. The content is a so called
secure CCA key token and a description can be found in the
  "Secure Key Solution with the Common Cryptographic
   Architecture Application Programmer's Guide".

There is a requirement to export and import these raw secure key token
objects. Export is simple: Just fetch the value the CKA_IBM_OPAQUE
attribute for a key object and dump it for example into a
file. Importing such a value was until now not supported. This patch
enables the possibility to import such a raw CCA key token object with
the C_CreateObject() pkcs#11 function.

To import an CCA DES key token, use a template with these mandatory
attributes:
  CKA_CLASS = CKO_SECRET_KEY
  CK_KEY_TYPE = CKK_DES
  CKA_VALUE - dummy but needs to have byte size of the clear key (8)
  CKA_IBM_OPAQUE - with the CCA DES key token

to import an CCA DES3 key token, use a template with these mandatory
attributes:
  CKA_CLASS = CKO_SECRET_KEY
  CK_KEY_TYPE = CKK_DES3
  CKA_VALUE - dummy but needs to have byte size of the clear key (24)
  CKA_IBM_OPAQUE - with the CCA DES3 key token

to import an CCA AES key token, use a template with these mandatory
attributes:
  CKA_CLASS = CKO_SECRET_KEY
  CK_KEY_TYPE = CKK_AES
  CKA_VALUE - dummy but needs to have byte size of the clear key (16,
	      24, or 32)
  CKA_IBM_OPAQUE - with the CCA AES key token
As of now only CCA AES data key tokens are supported. CCA AES cipher
keys do not work yet.

to import an CCA internal RSA private key token, use a template with
these mandatory attributes:
  CKA_CLASS = CKO_PRIVATE_KEY
  CKA_KEY_TYPE = CKK_RSA
  CKA_IBM_OPAQUE - with the CCA RSA private key token

to import an CCA RSA public key token, use a template with these
mandatory attributes:
  CKA_CLASS = CKO_PUBLIC_KEY
  CKA_KEY_TYPE = CKK_RSA
  CKA_IBM_OPAQUE - with the CCA RSA public key token

to import an CCA HMAC key token, use a template with these mandatory
attributes:
  CKA_CLASS = CKO_SECRET_KEY
  CKA_KEY_TYPE = CKK_GENERIC_SECRET
  CKA_VALUE - with a dummy value the size in bytes of the clear key
	      byte size.
  CKA_IBM_OPAQUE with the CCA RSA private key token.

CCA ECC is still work in progress, and maybe AES cipher key will also come.

### DIFF
--- a/testcases/common/common.c
+++ b/testcases/common/common.c
@@ -149,6 +149,9 @@ int create_AESKey(CK_SESSION_HANDLE session,
     };
 
     rc = funcs->C_CreateObject(session, keyTemplate, 5, h_key);
+    if (rc != CKR_OK) {
+        testcase_error("C_CreateObject rc=%s", p11_get_ckr(rc));
+    }
 
     return rc;
 }
@@ -167,6 +170,9 @@ int generate_AESKey(CK_SESSION_HANDLE session,
                                     key_gen_tmpl,
                                     1,
                                     h_key);
+    if (rc != CKR_OK) {
+	testcase_error("C_GenerateKey rc=%s", p11_get_ckr(rc));
+    }
 
     return rc;
 }
--- a/testcases/misc_tests/cca_export_import_test.c
+++ b/testcases/misc_tests/cca_export_import_test.c
@@ -1,0 +1,1451 @@
+/*
+ * import/export test for the CCA token
+ * by Harald Freudenberger <freude@de.ibm.com>
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <memory.h>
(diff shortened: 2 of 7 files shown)


hfreude:
no way, needs total rework

PR #347, opened 2020-11-24
PKCSCCA: Remove duplicated common code

ifranzki:
This is an attempt to remove the duplicated code parts that have been copied from common code files into pkcscca.c. Duplicating code causes a maintenance nightmare.

 Instead of duplicating the code, link the common code parts into pkcscca. Unfortunately, almost all common code parts  need tobe included, to resolve all referenced symbols. That way lots of unused code is included into pkcscca, but at least its the exact same code as for the tokens.

Take a look and let me know what you think. If you prefer to keep the code duplication, thats fine with me, too.....

### DIFF
--- a/usr/sbin/pkcscca/pkcscca.c
+++ b/usr/sbin/pkcscca/pkcscca.c
@@ -34,10 +34,26 @@
 #include <pkcs11types.h>
 
 #include "sw_crypt.h"
+#include "defs.h"
+#include "host_defs.h"
+#include "local_types.h"
+#include "h_extern.h"
+#include "slotmgr.h" // for ock_snprintf
+
+#define OCK_TOOL
+#include "pkcs_utils.h"
+
 #include "pkcscca.h"
 
+const char manuf[] = "IBM";
+const char model[] = "CCA";
+const char descr[] = "IBM CCA Token";
+const char label[] = "ccatok";
 
-int v_flag = 0;
+pkcs_trace_level_t trace_level = TRACE_LEVEL_NONE;
+token_spec_t token_specific = { 0 };
+
+int v_level = 0;
 void *p11_lib = NULL;
 void (*CSNDKTC) ();
 void (*CSNBKTC) ();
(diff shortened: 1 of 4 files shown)


usr/sbin/pkcscca/pkcscca.mk
```diff
@@ -1,13 +1,51 @@
...
+noinst_HEADERS += usr/lib/common/pkcs_utils.h
 
-usr_sbin_pkcscca_pkcscca_LDFLAGS = -lcrypto -ldl
+usr_sbin_pkcscca_pkcscca_LDFLAGS = -lcrypto -ldl -lrt
 
 usr_sbin_pkcscca_pkcscca_CFLAGS  =					\
 	-DSTDLL_NAME=\"pkcscca\"					\
-	-I${srcdir}/usr/include -I${srcdir}/usr/lib/common		\
+	-DTOK_NEW_DATA_STORE=0x0003000c					\
+	-DLINUX -DNOCDMF -DNODSA -DNODH -DNOECB				\
```
juergenchrist, line 14: I don't find any usages of `-DLINUX` and `-DNOECB`.  Do we really need them?  Also, seeing that we are using a configure script, shouldn't `-DLINUX` come via this script?

usr/sbin/pkcscca/pkcscca.mk
```diff
@@ -1,13 +1,51 @@
...
+	-DOCK_NO_SET_PERM -DOCK_NO_LOCAL_RNG				\
+	-I${srcdir}/usr/include 					\
+	-I${srcdir}/usr/lib/common					\
 	-I${srcdir}/usr/sbin/pkcscca
 
 usr_sbin_pkcscca_pkcscca_SOURCES =					\
```
juergenchrist, line 20: This definition pulls in all the symbols from these files.  Why not create a convenience library "libcommon.a" that we only use here to link against?

juergenchrist (changes requested):

jschmidb:
I introduced pkcs_util.c when working on the pkcstok_migrate utility for exact this reason: separate duplicate code from the tools in one common code source. So when adding any #ifdef in pkcs_util.c, check the side effects on pkcstok_migrate, which already uses pkcs_util.c.

ifranzki:
> I introduced pkcs_util.c when working on the pkcstok_migrate utility for exact this reason: separate duplicate code from the tools in one common code source. So when adding any #ifdef in pkcs_util.c, check the side effects on pkcstok_migrate, which already uses pkcs_util.c.

Yes I am aware about this. I added `#ifndef OCK_NO_xxx` (if NOT defined) to it to skip some functions that would still be duplicate when pkcs_util.c  and the other common code parts are linked together. Since these symbols are NOT defined by default, it does not hurt any existing code. Only for those that intentionally define these symbols it has an effect, i.e. to skip these functions in pkcs_util.c.

usr/sbin/pkcscca/pkcscca.c
```diff
@@ -2181,29 +1343,48 @@ void usage(char *progname)
...
-    printf(" -m keys.\t\t\tUnwraps private keys with the");
+    printf(" -m keys\t\t\tUnwraps private keys with the");
     printf(" old CCA master\n\t\t\t\tkey and wraps them with the");
     printf(" new CCA master key\n");
     printf(" -s, --slotid SLOTID\t\tPKCS slot number\n");
     printf(" -k aes|apka|asym|sym\t\tMigrate selected keytype\n\n");
     printf(" Options:\n");
     printf(" -d, --datastore DATASTORE\tCCA token datastore location\n");
-    printf(" -v, --verbose\t\t\tProvide more detailed output\n");
+    printf(" -v, --verbose LEVEL\t\tset verbose level (optional):\n");
```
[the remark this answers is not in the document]
ifranzki: will fix this.

usr/sbin/pkcscca/pkcscca.c
```diff
@@ -2262,7 +1443,13 @@ int main(int argc, char **argv)
             slot_id = atoi(optarg);
             break;
         case 'v':
-            v_flag++;
+            verbose = strdup(optarg);
```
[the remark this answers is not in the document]
ifranzki: right, we don't need this here. Will fix.

usr/sbin/pkcscca/pkcscca.mk
ifranzki: I copied this from the CCA token make file. But you are right both are not used anywhere. So at least I will remove them here. We might want to clean up all of those at other places in a separate PR....

usr/sbin/pkcscca/pkcscca.mk
ifranzki: Well, yes that's certainly an idea, also because the way it is currently it builds all these source files separately for every token (and now also for the pkcscca tool). 

HOWEVER: the reason that it is that way is probably because each token may (and coes) build these common code parts with different compile options.  Not sure if it is really required to build them with different options, but that's how it is currently. I would not like to change this in this PR, nevertheless, we should try to clean this up someday, compile all the common code parts with the same options, and then create a libcommon.a out of them. This would certainly reduce compile time, and make tings much clearer.

juergenchrist (approved):

p-steuer (approved):
Nice

PR #348, opened 2020-11-24
EP11: Update attributes in key blob when attributes are changed

ifranzki:
The EP11 key blob also contains a set of attributes (e.g. boolean attributes). Make sure that when attributes are changed (e.g. via C_SetAttributeValues), that the key blob is also updated with the changed attributes. Otherwise the view of the attributes of a key as seen by OCK and the view of the attributes as seen by EP11 may differ. This may cause errors (e.g. rejects of certain operations)  due to different attributes.

### DIFF
--- a/usr/lib/common/object.c
+++ b/usr/lib/common/object.c
@@ -143,8 +143,7 @@ CK_RV object_copy(STDLL_TokData_t * tokdata,
                   CK_ATTRIBUTE * pTemplate,
                   CK_ULONG ulCount, OBJECT * old_obj, OBJECT ** new_obj)
 {
-    TEMPLATE *tmpl = NULL;
-    TEMPLATE *new_tmpl = NULL;
+    TEMPLATE *tmpl, *new_tmpl;
     OBJECT *o = NULL;
     CK_BBOOL found;
     CK_ULONG class, subclass;
@@ -175,6 +174,7 @@ CK_RV object_copy(STDLL_TokData_t * tokdata,
     memset(o, 0x0, sizeof(OBJECT));
     memset(tmpl, 0x0, sizeof(TEMPLATE));
     memset(new_tmpl, 0x0, sizeof(TEMPLATE));
+    o->template = tmpl;
 
     rc = object_init_lock(o);
     if (rc != CKR_OK)
@@ -182,11 +182,12 @@ CK_RV object_copy(STDLL_TokData_t * tokdata,
 
     // copy the original object's attribute template
     //
-    rc = template_copy(tmpl, old_obj->template);
+    rc = template_copy(o->template, old_obj->template);
     if (rc != CKR_OK) {
         TRACE_DEVEL("Failed to copy template.\n");
         goto error;
     }
+
     rc = template_add_attributes(new_tmpl, pTemplate, ulCount);
     if (rc != CKR_OK) {
         TRACE_DEVEL("template_add_attributes failed.\n");
@@ -200,7 +201,7 @@ CK_RV object_copy(STDLL_TokData_t * tokdata,
     //    4) conflicting attributes/values
     //
 
-    found = template_get_class(tmpl, &class, &subclass);
+    found = template_get_class(o->template, &class, &subclass);
     if (found == FALSE) {
         TRACE_ERROR("Could not find CKA_CLASS in object's template.\n");
         rc = CKR_TEMPLATE_INCONSISTENT;
@@ -220,9 +221,23 @@ CK_RV object_copy(STDLL_TokData_t * tokdata,
         TRACE_DEVEL("template_validate_attributes failed.\n");
         goto error;
     }
+
+    /*
+     * Let the token know that attributes of the copied object have been
+     * changed.
+     */
+    if (token_specific.t_set_attribute_values != NULL) {
+        rc = token_specific.t_set_attribute_values(tokdata, o, new_tmpl);
+        if (rc != CKR_OK) {
+            TRACE_DEVEL("token_specific_set_attribute_values failed with %lu\n",
+                        rc);
+            goto error;
+        }
+   }
+
     // merge in the new attributes
     //
-    rc = template_merge(tmpl, &new_tmpl);
+    rc = template_merge(o->template, &new_tmpl);
     if (rc != CKR_OK) {
         TRACE_DEVEL("template_merge failed.\n");
         goto error;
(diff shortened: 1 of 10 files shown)


usr/lib/common/object.c
```diff
@@ -143,8 +143,7 @@ CK_RV object_copy(STDLL_TokData_t * tokdata,
                   CK_ATTRIBUTE * pTemplate,
                   CK_ULONG ulCount, OBJECT * old_obj, OBJECT ** new_obj)
 {
-    TEMPLATE *tmpl = NULL;
-    TEMPLATE *new_tmpl = NULL;
+    TEMPLATE *tmpl, *new_tmpl = NULL;
```
jschmidb, line 146: As both variables, tmpl, and new_tmpl are unconditionally assigned in lines 158 and 159, initializing new_tmpl = NULL is not necessary, but also not wrong. However, I would either initialize both variables or none to avoid confusion.

usr/lib/common/object.c
```diff
@@ -220,9 +221,22 @@ CK_RV object_copy(STDLL_TokData_t * tokdata,
...
     }
+
+    /*
+     * Let the token know that attributes of the copied object have been
+     * changed.
+     */
+    if (token_specific.t_set_attribute_values != NULL) {
+        rc = token_specific.t_set_attribute_values(tokdata, o, new_tmpl);
+        if (rc != CKR_OK) {
+            TRACE_DEVEL("token_specific_set_attribute_values failed.\n");
```
jschmidb, line 232: I would also print out the rc value in the trace.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -9585,3 +9585,110 @@ static void free_ep11_target_for_apqn(target_t target)
...
+CK_RV token_specific_set_attribute_values(STDLL_TokData_t *tokdata, OBJECT *obj,
+                                          TEMPLATE *new_tmpl)
+{
+    ep11_private_data_t *ep11_data = tokdata->private_data;
+    CK_OBJECT_CLASS class;
+    size_t keyblobsize = 0;
+    CK_BYTE *keyblob;
+    DL_NODE *node;
+    CK_ATTRIBUTE *ibm_opaque_attr = NULL;
+    CK_ATTRIBUTE_PTR attributes = NULL;
```
jschmidb, line 9597: Just for better reading: I would either always use CK_ATTRIBUTE *attr or CK_ATTRIBUTE_PTR attr. I personally prefer the standard way, coding ptr variables with the '*'.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -9585,3 +9585,110 @@ static void free_ep11_target_for_apqn(target_t target)
...
+        case CKA_ENCRYPT:
+        case CKA_DERIVE:
+        case CKA_UNWRAP:
+        case CKA_WRAP:
+        case CKA_VERIFY:
+        case CKA_VERIFY_RECOVER:
+        case CKA_WRAP_WITH_TRUSTED:
+        case CKA_TRUSTED:
+        case CKA_IBM_RESTRICTABLE:
+        case CKA_IBM_USE_AS_DATA:
```
jschmidb, line 9643: There are additional boolean attributes, which can be changed, e.g. CKA_COPYABLE and CKA_DESTROYABLE. Did you intentionally leave them out?

usr/lib/common/object.c
ifranzki: yes, right, initializing don't make sense here. Will fix this.

usr/lib/common/object.c
ifranzki: yes, will add this.

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: I tend to use CK_ATTRIBUTE_PTR if it is an attribute array, and CK_ATTRIBUTE * if it is a single attribute. Not that this really makes a difference though....

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: I took those that are listed in the EP11 structure document under 2.2.6 Boolean attribute listing. And CKA_COPYABLE and CKA_DESTROYABLE are not listed there. 
Not sure if the doc just does not mention them, but would support them anyway, or if they really don't support them.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -9585,3 +9585,110 @@ static void free_ep11_target_for_apqn(target_t target)
...
+    CK_ATTRIBUTE *attr;
+    CK_RV rc;
+
+    rc = template_attribute_get_ulong(obj->template, CKA_CLASS, &class);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("%s CKA_CLASS is missing\n", __func__);
+        return rc;
+    }
+
+    switch (class) {
```
juergenchrist, line 9608: Boolean attributes seem to be supported also on public keys and change whenever changed on a public key.

jschmidb (approved):

usr/lib/common/object.c
```diff
@@ -612,6 +625,15 @@ CK_RV object_set_attribute_values(STDLL_TokData_t * tokdata,
         TRACE_DEVEL("template_validate_attributes failed.\n");
         goto error;
     }
+
+    if (token_specific.t_set_attribute_values != NULL) {
+        rc = token_specific.t_set_attribute_values(tokdata, obj, new_tmpl);
+        if (rc != CKR_OK) {
+            TRACE_DEVEL("token_specific_set_attribute_values failed.\n");
```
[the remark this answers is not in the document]
ifranzki: No, it goes to the error label and there it returns rc.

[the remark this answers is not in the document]
juergenchrist: In other places I saw we include `rc` in the trace message to have it as close to the place where it happens.  If we do not need that here, okay.  I guess the correct place to put the trace describing `rc` would anyway be the token specific `t_set_attribute_values`.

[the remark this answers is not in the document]
ifranzki: ah yes, right, I guess I should include it into the trace message here as well (like it already is in object_copy). Should not hurt, even if the token specific implementation also traces the return code.

ifranzki:
Added a 3rd patch to pass CKA_MODIFIED to the EP11 key blob so that public keys can be modified via C_SetAttributeValues.

juergenchrist (approved):

p-steuer (approved):

PR #349, opened 2020-11-24
Support import of raw cca tokens with C_ObjectCreate

hfreude:
3 commits:
- TESTCASES: new testcase to test CCA opaque key token export and import
- CCA: support import of raw cca tokens with C_ObjectCreate via CKA_IBM_OPAQUE
- CCA: store only public token in IBM_OPAQUE attribute
for more details pleas see header of each commit

### DIFF
--- a/testcases/misc_tests/cca_export_import_test.c
+++ b/testcases/misc_tests/cca_export_import_test.c
@@ -1,0 +1,1448 @@
+/*
+ * COPYRIGHT (c) International Business Machines Corp. 2020
+ *
+ * This program is provided under the terms of the Common Public License,
+ * version 1.0 (CPL-1.0). Any use, reproduction or distribution for this
+ * software constitutes recipient's acceptance of CPL-1.0 terms which can
+ * be found in the file LICENSE file or at
+ * https://opensource.org/licenses/cpl1.0.php
+ *
+ * import/export test for the CCA token
+ * by Harald Freudenberger <freude@de.ibm.com>
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <memory.h>
+#include <stdint.h>
+
+#include "pkcs11types.h"
+#include "ec_curves.h"
+
+#include "regress.h"
+#include "common.c"
+
+// define this to enable the AES cipher key import test:
+// #define CCA_AES_CIPHER_KEY_SUPPORTED
+
+static CK_RV export_ibm_opaque(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE handle,
+			       CK_BYTE **buf, CK_ULONG *buflen)
+{
+    CK_RV rc;
+    CK_ULONG len;
+    CK_ATTRIBUTE a_opaque = { CKA_IBM_OPAQUE, NULL_PTR, 0 };
+
+    rc = funcs->C_GetAttributeValue(session, handle, &a_opaque, 1);
+    if (rc != CKR_OK) {
+	testcase_error("C_GetAttributeValue() rc=%s", p11_get_ckr(rc));
+	return rc;
+    }
+    len = a_opaque.ulValueLen;
+    if (!len) {
+	testcase_error("opaque attribute len is 0");
+	return CKR_ATTRIBUTE_VALUE_INVALID;
+    }
+    a_opaque.pValue = malloc(len);
+    if (!a_opaque.pValue) {
+	testcase_error("malloc(%lu) failed", len);
+	return CKR_HOST_MEMORY;
+    }
+    rc = funcs->C_GetAttributeValue(session, handle, &a_opaque, 1);
+    if (rc != CKR_OK) {
+	testcase_error("C_GetAttributeValue() rc=%s", p11_get_ckr(rc));
+	return rc;
+    }
+
+    *buf = a_opaque.pValue;
+    *buflen = len;
+
+    return CKR_OK;
+}
+
+static CK_RV import_cca_des_key(CK_SESSION_HANDLE session,
+				const char *label,
+				CK_BYTE *ccatoken, CK_ULONG tokenlen,
+				CK_OBJECT_HANDLE *handle)
+{
+    CK_RV rc;
+    CK_OBJECT_CLASS keyClass = CKO_SECRET_KEY;
+    CK_KEY_TYPE keyType = CKK_DES;
+    CK_BBOOL true = TRUE;
+    CK_BBOOL false = FALSE;
+    CK_ATTRIBUTE template[] = {
+	{CKA_CLASS, &keyClass, sizeof(keyClass)},
+	{CKA_KEY_TYPE, &keyType, sizeof(keyType)},
+	{CKA_LABEL, (char *) label, strlen(label) + 1},
+	{CKA_ENCRYPT, &true, sizeof(true)},
+	{CKA_TOKEN, &false, sizeof(false)},
+	{CKA_IBM_OPAQUE, ccatoken, tokenlen}
+    };
+    CK_ULONG nattr = sizeof(template)/sizeof(CK_ATTRIBUTE);
+
+    rc = funcs->C_CreateObject(session, template, nattr, handle);
+    if (rc != CKR_OK) {
+	testcase_error("C_CreateObject() rc=%s", p11_get_ckr(rc));
+    }
+
+    return rc;
+}
+
+static CK_RV import_cca_des3_key(CK_SESSION_HANDLE session,
+				 const char *label,
+				 CK_BYTE *ccatoken, CK_ULONG tokenlen,
+				 CK_OBJECT_HANDLE *handle)
+{
+    CK_RV rc;
+    CK_OBJECT_CLASS keyClass = CKO_SECRET_KEY;
+    CK_KEY_TYPE keyType = CKK_DES3;
+    CK_BBOOL true = TRUE;
+    CK_BBOOL false = FALSE;
+    CK_ATTRIBUTE template[] = {
+	{CKA_CLASS, &keyClass, sizeof(keyClass)},
+	{CKA_KEY_TYPE, &keyType, sizeof(keyType)},
+	{CKA_LABEL, (char *) label, strlen(label) + 1},
+	{CKA_ENCRYPT, &true, sizeof(true)},
+	{CKA_TOKEN, &false, sizeof(false)},
+	{CKA_IBM_OPAQUE, ccatoken, tokenlen}
+    };
+    CK_ULONG nattr = sizeof(template)/sizeof(CK_ATTRIBUTE);
+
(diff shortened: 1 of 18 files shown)


hfreude:
PR is updated, now without AES cipher key support and without requirement for CKA_VALUE for des and aes keys.

usr/lib/common/new_host.c
```diff
@@ -3638,7 +3638,7 @@ CK_RV SC_GenerateKeyPair(STDLL_TokData_t *tokdata,
     }
 
     TRACE_DEBUG("Private Template:\n");
-    attr = pPublicKeyTemplate;
+    attr = pPrivateKeyTemplate;
```
ifranzki, line 3641: Please check if this has to be corrected in the other copies of new_host.c as well (ep11 token and icsf token).

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -239,6 +240,300 @@ static const CK_ULONG cca_mech_list_len =
                         (sizeof(cca_mech_list) / sizeof(MECH_LIST_ELEMENT));
 
 
+#ifdef DEBUG
+/* Debug function: hexdump buffer with TRACE_DEBUG */
+static void dump_mem(const char *prestr, void *buf, unsigned len)
```
ifranzki, line 245: In ep11_specific.c there is a macro defined to dump memory:
`#define TRACE_DEBUG_DUMP(_buf, _buflen) hexdump(_buf, _buflen)`
Function hexdump() is alos local there. 
Maybe we can add a function into trace.c and a macro like TRACE_DEBUG_DUMP to use it, so that it can be used everywhere?

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -239,6 +240,300 @@ static const CK_ULONG cca_mech_list_len =
...
+                p++; len--;
+            }
+        }
+        line[67] = '\0';
+        TRACE_DEBUG("%s%s\n", prestr, line);
+    }
+}
+
+/* Debug function: dump one attribute via TRACE_DEBUG */
+static void dump_attr(CK_ATTRIBUTE *a)
```
ifranzki, line 271: This should probably go into attributes.c or p11util.c so that it is available for others, too. 
In p11utilc, there are already functions p11_get_ckr() and p11_get_ckm() to get the textual representation of return codes and mechanisms. maybe it would make sense to add a p11_get_cka() as well?

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -239,6 +240,300 @@ static const CK_ULONG cca_mech_list_len =
...
+        break;
+    default:
+        TRACE_DEBUG("  %s: len=%lu value:\n", typestr, a->ulValueLen);
+        dump_mem("  ", a->pValue, a->ulValueLen);
+        break;
+    }
+}
+
+/* Debug function: dump list of attribues from a template via TRACE_DEBUG */
+static void dump_template(TEMPLATE *tmpl)
```
ifranzki, line 364: move to template.c, or p11util.c ?

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -239,6 +240,300 @@ static const CK_ULONG cca_mech_list_len =
...
+/*
+ * Helper function: Analyse given CCA token.
+ * returns TRUE and keytype and keybitsize if token is known and seems
+ * to be valid (only basic checks are done), otherwise FALSE.
+ */
+static CK_BBOOL analyse_cca_key_token(const CK_BYTE *t, CK_ULONG tlen,
+                                      enum cca_token_type *keytype,
+                                      unsigned int *keybitsize)
+{
+    if (t[0] == 0x01 && (t[4] == 0x00 || t[4] == 0x01)) {
```
ifranzki, line 399: I usually like struct to overlay over the buffer better than hard coded offsets, but since the CCA token uses the offset-way already, I am fine with this, too.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -239,6 +240,300 @@ static const CK_ULONG cca_mech_list_len =
...
+        else if (t[59] == 0x20)
+            *keybitsize = 24 * 8;
+        else {
+            TRACE_DEVEL("CCA DES data key token has invalid/unknown keysize 0x%02x\n", (int)t[59]);
+            return FALSE;
+        }
+        return TRUE;
+    }
+
+    if (t[0] == 0x01 && t[4] == 0x04) {
```
ifranzki, line 419: 2 levels of switch()-es would be easier to read maybe:
```
switch (t[0]) {
case 0x01:
    switch (t[4]) {
    case xxx:
    }
}
```

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -3988,331 +4607,553 @@ static CK_RV build_public_EC_key_value_structure(CK_BYTE *pubkey, CK_ULONG puble
...
-	if (!object) {
-		TRACE_ERROR("Invalid argument\n");
-		return CKR_FUNCTION_FAILED;
-	}
+    if (!object) {
+        TRACE_ERROR("Invalid argument\n");
+        return CKR_FUNCTION_FAILED;
+    }
 
+    /* we only deal with key objects here */
```
ifranzki, line 5060: Actually, you could check here already if an CKA_IBM_OPAQUE is there, and if so, then analyze it and find out about the key type and class from the token. This would lift the requirement for the user to specify CLA_KEY_TYPE and CKA_CLASS when importing an opaque key. The key token does in fact contain all this information already, so no need for the user to specify it.
The only question would be if an EP11 key token also allows to get the same kind of information from it. So once we do the key blob import for EP11, it then should behave the same as for CCA.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -4409,6 +5250,11 @@ CK_RV token_specific_generic_secret_key_gen(STDLL_TokData_t * tokdata,
         return rc;
     }
 
+#ifdef DEBUG
+    TRACE_DEBUG("%s: imported object template attributes:\n", __func__);
+    dump_template(template);
```
ifranzki, line 5255: If you add this here (which makes sense, now that we have this functions), why not also adding it to the other keygen functions as well?
Any maybe adjust the trace message, here is is not an 'imported' key, but a 'generated' one.

testcases/misc_tests/cca_export_import_test.c
```diff
@@ -0,0 +1,1440 @@
+/*
```
ifranzki, line 1: I guess we should have the standard file header here with copyright infos, license, etc.

usr/lib/common/new_host.c
hfreude: needs also to get fixed in the other new_host.c files ...
I'll do that - maybe split these fixes into a new commit.

usr/lib/cca_stdll/cca_specific.c
hfreude: yes, makes sense - will do this

usr/lib/cca_stdll/cca_specific.c
hfreude: I have the structs - but as most of the cca specific code deals with some kind of offsets and would need a total rework to adapt to structs, I've chosen to use these simple offset values.

usr/lib/cca_stdll/cca_specific.c
hfreude: I'll leave this to Reinhard to decide. It's doable but then a user does not need and provide any key information but only the CCA secure key token. So no way to check if this was the key type the user intended to import.

testcases/misc_tests/cca_export_import_test.c
hfreude: Ok, will do.

usr/lib/cca_stdll/cca_specific.c
hfreude: done, will be part of the next PR update

usr/lib/cca_stdll/cca_specific.c
hfreude: well, no
the idea of the whole function is to group it based on keytypes. Mixing of ifs and switches and more levels may give the function a 1us performance boost but the human readability is much more worse. At least this is my understanding.

hfreude:
Ok, hopefully this was the last update of the PR:
- added some more attributes for the dump_attr function.
- added CKA_SENSITIVE=true for the import of sym and priv keys (request from Reinhard)

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -238,6 +238,27 @@ static const MECH_LIST_ELEMENT cca_mech_list[] = {
 static const CK_ULONG cca_mech_list_len =
                         (sizeof(cca_mech_list) / sizeof(MECH_LIST_ELEMENT));
 
+
+/* Helper function: build attribute and update template */
+static CK_RV build_update_attribute(TEMPLATE * tmpl,
+                                    CK_ATTRIBUTE_TYPE type,
+                                    CK_BYTE * data, CK_ULONG data_len)
```
ifranzki, line 403: We could move this function to common code (e.g. utility.c, were build_attribute() is already) so that others could also utilize it, since many other places also have the sequence of build_attribute() followed by template_update_attribute(), but that's not something that should hold off this PR.

testcases/misc_tests/cca_export_import_test.c
```diff
@@ -0,0 +1,1448 @@
...
+#include <stdint.h>
+
+#include "pkcs11types.h"
+#include "ec_curves.h"
+
+#include "regress.h"
+#include "common.c"
+
+// define this to enable the AES cipher key import test:
+// #define CCA_AES_CIPHER_KEY_SUPPORTED
```
ifranzki, line 27: I don't find this define anywhere else except in this testcase, especially not in cca_specific.c. Is this a left over, or did you leave it there intentionally?

ifranzki (approved):
lgtm

testcases/misc_tests/cca_export_import_test.c
hfreude: You are right. This is local to the file.
The code to support import of CCA AES cipher blobs is easy - just add it to the switch within the symetric key import code function. However, I did not want to discard the test code as I spend some time to develop it and so I commented it out with this #ifdef CCA_AES_CIPHER_KEY_SUPPORTED

usr/lib/cca_stdll/cca_specific.c
hfreude: yes, this could and should be done. And - as this function avoids the free() if update_attribute() fails, all the places where build_attribute() + update_attribute() is used would benefit from this.

p-steuer (approved):
lgtm

PR #350, opened 2020-11-25
Fix compiling with c++

gleichdick:
Using this library with g++ yields
```
g++ -c -o/dev/null /usr/include/opencryptoki/pkcs11types.h
/usr/include/opencryptoki/pkcs11types.h:1486:16: error: using typedef-name ‘CK_IBM_FUNCTION_LIST_1_0_PTR’ after ‘struct’
 1486 | typedef struct CK_IBM_FUNCTION_LIST_1_0_PTR CK_PTR CK_IBM_FUNCTION_LIST_1_0_PTR_PTR;
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/opencryptoki/pkcs11types.h:1485:48: note: ‘CK_IBM_FUNCTION_LIST_1_0_PTR’ has a previous declaration here
 1485 | typedef struct CK_IBM_FUNCTION_LIST_1_0 CK_PTR CK_IBM_FUNCTION_LIST_1_0_PTR;
      |                                                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

gcc 10.2.1, openSuSE Tumbleweed, openCryptoki-devel-3.15.1-1.1.x86_64

PR #351, opened 2020-11-26
Add C++ build test

p-steuer:
Regression test for https://github.com/opencryptoki/opencryptoki/pull/350 . Travis is expected to fail until 350 is merged.

### DIFF
--- a/testcases/build/cpp_test.cpp
+++ b/testcases/build/cpp_test.cpp
@@ -1,0 +1,27 @@
+/*
+ * COPYRIGHT (c) International Business Machines Corp. 2020
+ *
+ * This program is provided under the terms of the Common Public License,
+ * version 1.0 (CPL-1.0). Any use, reproduction or distribution for this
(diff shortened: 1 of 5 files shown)


testcases/build/cpp_test.cpp
```diff
@@ -0,0 +1,27 @@
...
+#include "pkcs11.h"
+#include "ec_curves.h"
+
+int main(void)
+{
+    int rv;
+
+    rv = 0;
+    /* make sure we return non-zero if rv is non-zero */
+    return ((rv == 0) || (rv % 256) ? (int)rv : -1);
```
ifranzki, line 26: Does not make much sense if you set rv to zero just before....

ifranzki (approved):
Feel free to ignore my comment or address it. Approved either way.

testcases/build/cpp_test.cpp
p-steuer: The LOC translates the rv to something autotools can digest. I left it in intentionally.

PR #352, opened 2020-11-26
Bugfix and comment

jschmidb:
Just two small fixes:

- The wrong comment in template_update_attribute was obvious, but Ingo also suggested to change the return code from "function failed" to arguments bad", because this fits better.
- The malloc() in the four init functions for encr/decr/sign/verify caused a problem in aes_cmac_sign where the mac_len is calculated, because malloc(0) returned some undefined value, which got assigned to mech.pParameter. Therefore let's do the malloc only if a mech param is given.

### DIFF
--- a/usr/lib/ep11_stdll/new_host.c
+++ b/usr/lib/ep11_stdll/new_host.c
@@ -2058,10 +2058,21 @@ CK_RV SC_EncryptInit(STDLL_TokData_t *tokdata, ST_SESSION_HANDLE *sSession,
         sess->encr_ctx.key = hKey;
 
         sess->encr_ctx.mech.mechanism = pMechanism->mechanism;
-        sess->encr_ctx.mech.pParameter = malloc(pMechanism->ulParameterLen);
-        memcpy(sess->encr_ctx.mech.pParameter, pMechanism->pParameter,
-               pMechanism->ulParameterLen);
-        sess->encr_ctx.mech.ulParameterLen = pMechanism->ulParameterLen;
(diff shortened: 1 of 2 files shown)


usr/lib/ep11_stdll/new_host.c
```diff
@@ -2058,10 +2058,15 @@ CK_RV SC_EncryptInit(STDLL_TokData_t *tokdata, ST_SESSION_HANDLE *sSession,
         sess->encr_ctx.key = hKey;
 
         sess->encr_ctx.mech.mechanism = pMechanism->mechanism;
-        sess->encr_ctx.mech.pParameter = malloc(pMechanism->ulParameterLen);
-        memcpy(sess->encr_ctx.mech.pParameter, pMechanism->pParameter,
-               pMechanism->ulParameterLen);
-        sess->encr_ctx.mech.ulParameterLen = pMechanism->ulParameterLen;
+        if (pMechanism->pParameter && pMechanism->ulParameterLen > 0) {
+            sess->encr_ctx.mech.pParameter = malloc(pMechanism->ulParameterLen);
+            memcpy(sess->encr_ctx.mech.pParameter, pMechanism->pParameter,
```
ifranzki, line 2063: While you are at it, please also check if the malloc has worked, before attempting to memcpy.

jschmidb: yep, that makes sense. Done.

ifranzki (approved):
lgtm.

p-steuer (approved):

PR #355, opened 2020-12-09
Remove superfluous code

jschmidb:
As the vendor specific attributes are checked at the beginning of the
routine, they can be removed from the switch statement.


ifranzki (approved):
lgtm.

Although vendor specific attributes are all considered as being 'defined' in is_attribute_defined(), I am not sure if this means that they any vendor specific attribute can be specified in templates.
I guess template_validate_attribute() and friends still check if the specified attributes are valid for the specific object class, and may reject unknown vendor specific attributes. 

Nevertheless, this code change makes it easier to maintain the list of supported attributes.

p-steuer (approved):

PR #360, opened 2021-01-13
TPM fixes (as reported by issues from @grawity)

ifranzki:


ifranzki:
@grawity confirmed that this fixes the problems, see associated issues.

p-steuer (approved):
approved based on your positive testing result.

PR #365, opened 2021-01-27
Fix attribute-array flatten and unflatten

ifranzki:
Some ULONG attributes are always stored as 32 bit ULONG, while others are stored in the length that CK_ULONG is on the systems architecture. I.e. on a 64 bit system a CK_ULONG is 64 bit in size, while on a 32 bit system it is 32 bit in size.

The same logic must also be used when unflattening attribute-array elements, when an element is such an attribute that is always stored as 32 bit ULONG.

Also fix an error when flattening an attribute-array element.

### DIFF
--- a/testcases/pkcs11/attribute.c
+++ b/testcases/pkcs11/attribute.c
@@ -29,6 +29,8 @@ CK_RV do_TestAttributes(void)
     CK_ULONG user_pin_len;
     CK_BYTE so_pin[PKCS11_MAX_PIN_LEN];
     CK_ULONG so_pin_len;
+    CK_ULONG find_count;
+    CK_OBJECT_HANDLE obj_list[10];
 
     CK_BYTE modulus[] = {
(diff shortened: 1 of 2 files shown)


juergenchrist (approved):

p-steuer:
Is this change backward compatible ?

ifranzki:
The code that introduced the bug was added after 3.15.1, so it is not included in any distro so far.  
The bug caused that attribute-array attributes with > 0 elements were always stored incorrectly, so this did never really work since the code was introduced anyway.

ifranzki:
I added one more commit to enhance a testcase to test attribute-array attributes.

p-steuer (approved):

juergenchrist:
Please merge this PR soon.  It fixes some of our daily CI errors...

p-steuer (approved):

PR #366, opened 2021-01-28
Added test cases for the p11sak tool

reumann-ibm:
Added test cases for the p11sak tool incl. Changes for CI integration in the makefile. Signed-off-by: Matthias Reumann <matthias.reumann1@ibm.com

### DIFF
--- a/testcases/misc_tests/p11sak_test.sh
+++ b/testcases/misc_tests/p11sak_test.sh
@@ -1,0 +1,419 @@
+#!/bin/bash
+#
+# COPYRIGHT (c) International Business Machines Corp. 2020
+#
+# This program is provided under the terms of the Common Public License,
+# version 1.0 (CPL-1.0). Any use, reproduction or distribution for this software
+# constitutes recipient's acceptance of CPL-1.0 terms which can be found
+# in the file LICENSE file or at https://opensource.org/licenses/cpl1.0.php
+
+# - The PKCS11_USER_PIN environment variable must hold the user pin.
+# - The SLOT environment variable must hold the slot id of the token under test.
+#
+# sudo -E ./p11sak_test.sh
+
+
+
+echo "** Now executing 'p11sak_test.sh'"
+
+
+# tmp files
+
+P11SAK_DES_PRE=p11sak-des-pre.out
+P11SAK_DES_POST=p11sak-des-post.out
+P11SAK_3DES_PRE=p11sak-3des-pre.out
+P11SAK_3DES_POST=p11sak-3des-post.out
+P11SAK_AES_PRE=p11sak-aes-pre.out
+P11SAK_AES_POST=p11sak-aes-post.out
+P11SAK_RSA_PRE=p11sak-rsa-pre.out
+P11SAK_RSA_POST=p11sak-rsa-post.out
+P11SAK_EC_PRE=p11sak-ec-pre.out
+P11SAK_EC_POST=p11sak-ec-post.out
+
+
+
+echo "** Now generating keys - 'p11sak_test.sh'"
+
+
+# generate objects
+
+# des
+p11sak generate-key des --slot $SLOT --pin $PKCS11_USER_PIN --label p11sak-des
+# 3des
+p11sak generate-key 3des --slot $SLOT --pin $PKCS11_USER_PIN --label p11sak-3des
+# aes [128 | 192 | 256]
+p11sak generate-key aes 128 --slot $SLOT --pin $PKCS11_USER_PIN --label p11sak-aes-128
+p11sak generate-key aes 192 --slot $SLOT --pin $PKCS11_USER_PIN --label p11sak-aes-192
+p11sak generate-key aes 256 --slot $SLOT --pin $PKCS11_USER_PIN --label p11sak-aes-256
+# rsa [1024 | 2048 | 4096]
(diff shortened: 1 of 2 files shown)


ifranzki:
What's the difference between the 3rd and 4th Commit? Seems to be duplicate....

reumann-ibm:
after the 3rd commit, git status showed that the Makefile.am in the opencryptoki directory was not added in the commit. Therefore I commited the Makefile.am in the 4th commit to include the new p11sak-test.sh in the CI.

ifranzki:
But Makefile.am is already in the 3rd patch. 

Github also shows: 
This branch has conflicts that must be resolved
Use the web editor or the  to resolve conflicts.
Conflicting files
Makefile.am

ifranzki:
Now the Makefile.am is missing in your 3rd commit.... I guess your intent is to update it as well, right?

usr/sbin/p11sak/p11sak.c
```diff
@@ -2230,7 +2230,6 @@ static CK_RV finalize_destroy_object(char *label, CK_SESSION_HANDLE *session,
             goto done;
         }
         *boolDestroyFlag = CK_TRUE;
-        printf("DONE - Destroy Object with Label: %s\n", label);
```
ifranzki, line 2235: Why don't you combine patch 1 and 2 into just one patch? This seems to change a lone also touched in first patch.

testcases/misc_tests/p11sak_test.sh
```diff
@@ -0,0 +1,419 @@
...
+
+
+
+echo "** Now generating keys - 'p11sak_test.sh'"
+
+
+# generate objects
+
+# des
+p11sak generate-key des --slot $SLOT --pin $PKCS11_USER_PIN --label p11sak-des
```
ifranzki, line 41: How would the CI know if such a test step fails? 
I don't see any return code checking, etc.... The script would just continue to execute. Is there a subsequent process that scans the output for errors?

testcases/misc_tests/p11sak_test.sh
```diff
@@ -0,0 +1,419 @@
...
+p11sak generate-key ec prime256v1 --slot $SLOT --pin $PKCS11_USER_PIN --label p11sak-ec-prime256v1
+p11sak generate-key ec secp384r1 --slot $SLOT --pin $PKCS11_USER_PIN --label p11sak-ec-secp384r1
+p11sak generate-key ec secp521r1 --slot $SLOT --pin $PKCS11_USER_PIN --label p11sak-ec-secp521r1
+
+
+echo "** Now list keys and redirect output to pre-files - 'p11sak_test.sh'"
+
+
+# list objects
+p11sak list-key des --slot $SLOT --pin $PKCS11_USER_PIN &> $P11SAK_DES_PRE
```
ifranzki, line 62: How is checked if the list command did list the expected keys?

testcases/misc_tests/p11sak_test.sh
```diff
@@ -0,0 +1,419 @@
...
+p11sak list-key aes --slot $SLOT --pin $PKCS11_USER_PIN &> $P11SAK_AES_POST
+p11sak list-key rsa --slot $SLOT --pin $PKCS11_USER_PIN &> $P11SAK_RSA_POST
+p11sak list-key ec --slot $SLOT --pin $PKCS11_USER_PIN &> $P11SAK_EC_POST
+
+
+echo "** Now checking output files to determine PASS/FAIL of tests - 'p11sak_test.sh'"
+
+
+# check DES
+grep -q 'p11sak-des' $P11SAK_DES_PRE
```
ifranzki, line 116: OK, here you check if the list command shows the expected key. If the expected key is not there, then either the generate command has failed, or the list command has failed....
That's of course better than nothing, but still hard to distinguish what error has happened.

reumann-ibm: That is a good point. Generally, the assumption is that the list command cannot display a key if the generate command failed. 
This is an initial test case that I want to expand to include also output from the eg. the p11tool. i. e. another tool generates the key and the p11sak tool lists the key. Also, this can then be used to actually check if an e. g. AES 192 key generated by the p11sak tool and the p11tool compare. This way, one can further build test cases that can pin point more specifically where the tool failed.

testcases/misc_tests/p11sak_test.sh
reumann-ibm: See below

testcases/misc_tests/p11sak_test.sh
reumann-ibm: See below

reumann-ibm:
Updated pull request to resolve conflict in merging Makefile.am

juergenchrist (approved):

ifranzki:
> Updated pull request to resolve conflict in merging Makefile.am

OK, but now commits 1 and 2 are gone, now only the makefile and test script changes are there. Is this intended?

ifranzki (approved):
After talking to Matthias it turned out the the original commit #1 and #2 where left overs and have already been merged to master with another PR. So the remaining commit is the one that is intended to be merged with this PR.

This commit looks good for now, although the test script might be enhanced at a later time, in terms of more precise checking for errors of the run commands.

ifranzki:
We also might add the execution of p11sak_test.sh into .travis.yaml, so that p11sak also gets tested by travis.
Nothing to hold off this PR, but should be a simple thing.

p-steuer (approved):

PR #367, opened 2021-01-28
EP11 protected key option

jschmidb:
This pull request introduces an option to the ep11 token instance, defining that EP11 
secure keys of a certain type (currently only AES) shall be transformed into
protected keys, and the protected keys shall be used whenever possible. This significantly improves performance, because calls to CPACF are much cheaper than calls to an EP11 coprocessor card.


usr/lib/common/obj_mgr.c
```diff
@@ -1696,6 +1696,89 @@ CK_RV object_mgr_restore_obj_withSize(STDLL_TokData_t *tokdata, CK_BYTE *data,
...
+ */
+CK_RV object_mgr_save_token_object(STDLL_TokData_t *tokdata, OBJECT *obj)
+{
+    TOK_OBJ_ENTRY *entry = NULL;
+    CK_ULONG index;
+    CK_RV rc;
+
+    // I still think there's a race condition here if two processes are
+    // updating the same token object at the same time.  I don't know how
+    // to solve this short of assigning each token object it's own mutex...
```
ifranzki, line 1710: I guess these comments can be removed. This race should no longer exist (we have object level locking now).

usr/lib/common/template.c
```diff
@@ -1896,6 +1904,9 @@ CK_RV template_validate_base_attribute(TEMPLATE *tmpl, CK_ATTRIBUTE *attr,
         break;
     case CKA_UNIQUE_ID:
         break;
+    case CKA_IBM_PROTKEY_EXTRACTABLE:
```
ifranzki, line 1907: CKA_IBM_PROTKEY_EXTRACTABLE is added to key objects only in key_object_set_default_attributes(), so you should only allow it on key objects. I.e. this check must be moved to key_object_validate_attribute() in key.c. Otherwise one could set attribute CKA_IBM_PROTKEY_EXTRACTABLE on non-key objects also.

usr/lib/common/obj_mgr.c
```diff
@@ -1750,72 +1833,9 @@ CK_RV object_mgr_set_attribute_values(STDLL_TokData_t *tokdata,
...
-            }
-
-            entry = &tokdata->global_shm->publ_tok_objs[index];
-        }
-
-        entry->count_lo = obj->count_lo;
-        entry->count_hi = obj->count_hi;
-
-        rc = XProcUnLock(tokdata);
+        rc = object_mgr_save_token_object(tokdata, obj);
```
ifranzki, line 1836: You moved this into  a separate function. That's great, but this separate function is not used anywhere else in this commit. It might be clearer to have this change in the commit that uses object_mgr_save_token_object at the new place.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -489,6 +519,467 @@ static CK_RV cleanse_attribute(TEMPLATE *template,
...
+} __attribute__((packed)) wrapped_key_t;
+
+/**
+ * Callback function used by handle_all_ep11_cards() for creating a protected
+ * key via the given APQN (adaper,domain).
+ */
+static CK_RV ep11tok_pkey_wrap_handler(uint_32 adapter, uint_32 domain,
+                                       void *handler_data)
+{
+    CK_BYTE iv[16] = {
```
ifranzki, line 555: const ?

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -489,6 +519,467 @@ static CK_RV cleanse_attribute(TEMPLATE *template,
...
+    size_t csum_l = sizeof(csum);
+    CK_BYTE blob[MAX_BLOBSIZE];
+    size_t blobsize = sizeof(blob);
+    CK_ATTRIBUTE *pkey_attr = NULL, *blob_attr=NULL;
+    CK_RV ret;
+
+    /* Check if CPACF_WRAP mech supported */
+    if (ep11tok_is_mechanism_supported(tokdata, CKM_IBM_CPACF_WRAP) != CKR_OK) {
+        TRACE_INFO("CKM_IBM_CPACF_WRAP not supported on this system.\n");
+        ret = CKR_FUNCTION_NOT_SUPPORTED;
```
ifranzki, line 686: Does this mean that the token does not initialize if pkey is enabled in the config, but the system does not support it? If so, then I would also issue a syslog message here. Otherwise the user won't have a chance to know why the token does not initialize.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -489,6 +519,467 @@ static CK_RV cleanse_attribute(TEMPLATE *template,
...
+ */
+static CK_BBOOL ep11tok_pkey_is_valid(STDLL_TokData_t *tokdata, OBJECT *key_obj)
+{
+    ep11_private_data_t *ep11_data = tokdata->private_data;
+    CK_ATTRIBUTE *pkey_attr = NULL;
+    int vp_offset;
+
+    if (template_attribute_get_non_empty(key_obj->template, CKA_IBM_OPAQUE_PKEY,
+                                         &pkey_attr) == CKR_OK) {
+        if (pkey_attr->ulValueLen >= 16 + MK_VP_LENGTH) {
```
ifranzki, line 753: Use AES 128 key length symbol ?

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -489,6 +519,467 @@ static CK_RV cleanse_attribute(TEMPLATE *template,
...
+
+    object_put(tokdata, key_obj, TRUE);
+    key_obj = NULL;
+
+    return success;
+}
+
+/**
+ * This function is called whenever a new secret or private key object is
+ * created. It sets attribute CKA_IBM_PROTKEY_EXTRACTABLE according to the
```
ifranzki, line 926: Comment: Don't confuse secret or private/public keys with private objects. One means a key type, the other one an attribute of an object. 

I would assume that this function should be called for all key objects, regardless of the key type, and regardless if the object is private or not.

testcases/crypto/aes_func.c
```diff
@@ -2030,6 +2031,8 @@ int main(int argc, char **argv)
     }
 
     testcase_setup(0);          //TODO
+
+    pkey = CK_FALSE;
```
ifranzki, line 2327: Do you ever set this to TRUE? Looks like you run it only once with pkey=FALSE....

testcases/crypto/aes_func.c
```diff
@@ -2034,6 +2326,11 @@ int main(int argc, char **argv)
 
     pkey = CK_FALSE;
     rv = aes_funcs();
+
+    pkey = CK_TRUE;
```
ifranzki, line 2330: Ah OK, here it is, OK.

Could you skip this pkey= TRUE case for non-secure key tokens?  I.e. run it only for ep11 token for now.

jschmidb:
Looks like Travis doesn't like some of my C-constructs .... will fix them next week.

usr/lib/common/object.c
```diff
@@ -444,6 +444,47 @@ CK_BBOOL object_is_session_object(OBJECT * obj)
         return TRUE;
 }
 
+// object_is_extractable()
+//
+CK_BBOOL object_is_extractable(OBJECT * obj)
```
juergenchrist, line 449: Isn't extractable a property of a key?  Why do we have it on object level?

usr/lib/common/pkey_utils.c
```diff
@@ -0,0 +1,531 @@
...
+ */
+CK_RV pkey_aes_ecb(OBJECT *key_obj, CK_BYTE *in_data,
+                   CK_ULONG in_data_len, CK_BYTE *out_data,
+                   CK_ULONG_PTR p_output_data_len, CK_BYTE encrypt)
+{
+    CK_RV ret;
+    unsigned long fc;
+    CK_ATTRIBUTE *pkey_attr = NULL;
+    CK_ULONG clear_keylen;
+    struct __attribute__((packed)){
```
juergenchrist, line 272: Why packed?

usr/lib/common/pkey_utils.c
```diff
@@ -0,0 +1,531 @@
...
+
+/**
+ * Return the KM/KMC/KMAC function code for the given key length and
+ * encrypt/decrypt op.
+ */
+unsigned long get_function_code(CK_ULONG clear_keylen, CK_BYTE encrypt)
+{
+    unsigned long fc;
+
+    switch (clear_keylen) {
```
juergenchrist, line 239: how about `fc = ((clear_keylen - 16) >> 3) + 0x1a`?

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -489,6 +519,467 @@ static CK_RV cleanse_attribute(TEMPLATE *template,
...
+    /* Check if returned key type is what we expect:
+     *  - 0x1 = AES with bit length 128, 192 and 256
+     *  - 0x2 = 2DES and 3DES
+     *  - 0x3 = EC-P with bit length 192, 224, 256, 384 and 521
+     *  - 0x4 = ED25519 and ED448
+     */
+    wk = (wrapped_key_t *) &ep11_buf;
+    switch (wk->cpacf_key_type) {
+    case 1:
+    case 3:
```
juergenchrist, line 634: Are there symbolic names for these CPACF key types that we could use instead of the hard-coded constants?

usr/lib/common/object.c
```diff
@@ -795,6 +836,14 @@ CK_RV object_create_skel(STDLL_TokData_t * tokdata,
     if (rc != CKR_OK)
         goto done;
 
+    if ((class == CKO_SECRET_KEY || class == CKO_PRIVATE_KEY) &&
```
ifranzki, line 839: Why not also for public keys (CKO_PUBLIC_KEY) ?
The PKEY attribute is valid for All Key objects, so then this routine should be called for ALL key objects.

usr/lib/common/tok_spec_struct.h
```diff
@@ -275,6 +275,8 @@ struct token_specific_struct {
                                 CK_BYTE *, CK_ULONG , CK_BYTE *, CK_ULONG *);
 
     CK_RV(*t_set_attribute_values) (STDLL_TokData_t *, OBJECT *, TEMPLATE *);
+
+    CK_RV(*t_set_pkey_attr) (STDLL_TokData_t *, TEMPLATE *);
```
ifranzki, line 279: The name set_pkey_attr is still quite generic. It refers to pkey, that's great, and that it is somehow supposed to set attributes, but it is not clear from the name when this function is called (i.e. during key object creation - generate, create, wrap, unwrap), and it is not clear if the function is called when common code sets attributes, or to let the token set attributes (the second is the intend of this function) . 

Also, there is also a token_specific_set_attribute_values which kind of sounds similar (but has a totally different meaning). 

And there is another similar function: token_specific_object_add. This one is also called during object creation. What's the difference between your new function and and token_specific_object_add? Why can't you use token_specific_object_add instead of yours? I guess the answer is because token_specific_object_add is only called for C_CreateObjetc, but not for the other object creation cases (Generate, Unwrap, Derive), but yours is called for all those cases. 

I see 2 options: 
- either we make it really generic, so that it is called for ALL objects, and lets the token add (and modify?) attributes that it needs, totally independent of PKEY (but PKEY uses it as first),
- or make it very PKEY specific, i.e. change the name to something that makes it clear that it is called during key object creation. 

Remember: This is common code, so it should be generic, so that it could be used by all tokens.

usr/lib/common/key.c
```diff
@@ -7902,6 +7920,22 @@ CK_RV aes_set_default_attributes(TEMPLATE *tmpl, CK_ULONG mode)
     }
     value_attr = NULL;
 
+    /* If CKA_VALUE specified in base tmpl, add CKA_VALUE_LEN to tmpl */
+    if (template_attribute_find(basetmpl, CKA_VALUE, &value_attr)) {
+        keysize = value_attr->ulValueLen;
+        rc = build_attribute(CKA_VALUE_LEN, (CK_BYTE *)&keysize, sizeof(CK_ULONG), &len_attr);
```
ifranzki, line 7926: There are cases where CKA_VALUE_LEN is already specified by the user, i.e. for derive or Unwrap of an AES key. I guess in these cases CKA_VALUE is not specified, so your code would not do anything (as intended). But are there cases where you would overwrite a user supplied CKA_VALUE_LEN ?

usr/lib/common/object.c
ifranzki: Well, could be moved to key.c as key_object_is_extractable() , but since we don't yet have any other such key-object helpers I am OK with either way.

usr/lib/common/object.c
```diff
@@ -444,6 +444,47 @@ CK_BBOOL object_is_session_object(OBJECT * obj)
...
+// object_is_extractable()
+//
+CK_BBOOL object_is_extractable(OBJECT * obj)
+{
+    CK_BBOOL val;
+    CK_RV rc;
+
+    rc = template_attribute_get_bool(obj->template, CKA_EXTRACTABLE, &val);
+    if (rc != CKR_OK)
+        return TRUE;
```
ifranzki, line 456: Should it really return TRUE if the CKA_EXTRACTABLE attribute is not there? Since the attributes default is TRUE that might be OK though.

usr/lib/common/pkey_utils.c
```diff
@@ -0,0 +1,531 @@
...
+    CK_RV ret;
+
+    /* Unlock obj to remove READ_LOCK */
+    ret = object_unlock(key_obj);
+    if (ret != CKR_OK) {
+        TRACE_ERROR("object_unlock failed with rc=0x%lx\n", ret);
+        return ret;
+    }
+
+    /* Obtain write lock on key object */
```
ifranzki, line 171: Since you have unlocked the object here, it could be modified by other threads until you get the write lock. 

Could there be any problem when other threads modify the object or its attributes in any way here? 
This function will blindly add/update the PKEY attribute, so this function should be OK. 
But what about the caller(s) of this function? Do they somehow keep a reference on an attribute or attribute value of this object? This reference may no longer be valid on return of this function (due to unlock of the object and concurrent modification).

What happens if multiple threads try to set a new PKEY for the same object? I guess the last one will win, and the prot key will anyway be all the same, but please check carefully if such multithreading issues are cleanly handled..

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -489,6 +519,467 @@ static CK_RV cleanse_attribute(TEMPLATE *template,
...
+               MK_VP_LENGTH) != 0) {
+        TRACE_ERROR("vp of this pkey does not match with the one in ep11_data (should not occur)\n");
+        ret = CKR_FUNCTION_FAILED;
+        goto done;
+    }
+
+    /* Now update the key obj. If it's a token obj, it will be also updated
+     * in the repository. */
+    ret = pkey_update_and_save(tokdata, key_obj, pkey_attr);
+    if (ret != CKR_OK) {
```
ifranzki, line 807: Because pkey_update_and_save unlocks and relocks the object, it might get concurrently updated by other threads at this place. 

Since you just return from this function after returning from pkey_update_and_save it should be save for this function, but how about the callers?

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -489,6 +519,467 @@ static CK_RV cleanse_attribute(TEMPLATE *template,
...
+            goto done;
+        }
+        if (template_attribute_get_non_empty(key_obj->template,
+                                             CKA_IBM_OPAQUE_PKEY,
+                                             &opaque_attr) != CKR_OK ||
+            !ep11tok_pkey_is_valid(tokdata, key_obj)) {
+            /* this key has either no pkey attr, or it is not valid,
+             * try to create one */
+            ret = ep11tok_pkey_update(tokdata, key_obj);
+            if (ret != CKR_OK) {
```
ifranzki, line 875: Because pkey_update_and_save (called within ep11tok_pkey_update) unlocks and relocks the object, it might get concurrently updated by other threads at this place. 

Since you just return from this function after returning from ep11tok_pkey_update it should be save for this function, but how about the callers?

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -5507,6 +6072,14 @@ CK_RV ep11tok_sign_init(STDLL_TokData_t * tokdata, SESSION * session,
         goto done;
     }
 
+    if (ep11tok_pkey_usage_ok(tokdata, session, key_obj, mech)) {
+        rc = sign_mgr_init(tokdata, session, ctx, mech, recover_mode, key);
```
ifranzki, line 6076: This comment applies to all places where ep11tok_pkey_usage_ok is called:

Because pkey_update_and_save (called within ep11tok_pkey_update which is called by ep11tok_pkey_usage_ok) unlocks and relocks the object, it might get concurrently updated by other threads at this place.

As far as I can see, there are references to object attributes (e.g. keyblob is pValue from CKA_IBM_OPAQUE of the object) used below this point. Those references may no longer be valid ! 

If you can ensure that the unlock/relock is ONLY performed when ep11tok_pkey_usage_ok returns TRUE, and thus the sign_mgr_init path is taken, then it might be OK.

Please check all places where ep11tok_pkey_usage_ok is called!

ifranzki: You might think about moving h_opaque_2_blob below this point.... (Just an idea, don't know it this really works),

usr/lib/common/pkey_utils.h
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Modifier bit for CPACF instructions.
+ */
+#define CPACF_ENCRYPT         0x00
+#define CPACF_DECRYPT         0x80
+
+/**
+ * function codes for the KM/KMC/KMAC instruction.
+ */
+#define Encrypted_AES_128     0x1a
```
ifranzki, line 31: Can those be all UPPERCASE, like other defines?

usr/lib/common/pkey_utils.c
ifranzki: well, does this really making it easier to read?  :-)

usr/lib/common/pkey_utils.c
ifranzki: Why a struct at all?

usr/lib/common/pkey_utils.h
```diff
@@ -0,0 +1,73 @@
...
+#define Encrypted_AES_192     0x1b
+#define Encrypted_AES_256     0x1c
+
+
+typedef struct {
+    uint8_t  type;      /* 0x00 (TOKTYPE_NON_CCA) */
```
ifranzki, line 37: Can you add defines for TOKTYPE_NON_CCA and TOKVER_EP11_AES ?

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -434,7 +451,17 @@ typedef struct {
     short apqns[2 * MAX_APQN];
 } __attribute__ ((packed)) ep11_target_t;
 
+static CK_RV handle_all_ep11_cards(ep11_target_t * ep11_targets,
+                                   adapter_handler_t handler,
+                                   void *handler_data);
+
 /* EP11 token private data */
+#define MK_VP_LENGTH                32
```
ifranzki, line 459: HM, MK is ambiguous. There are crypto card MKs and LPAR/Virtual Guest MKs. How about naming this PKEY_MK_VP_LENTH?

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -489,6 +519,467 @@ static CK_RV cleanse_attribute(TEMPLATE *template,
...
+    CK_VOID_PTR secure_key;
+    CK_ULONG secure_key_len;
+    CK_BYTE *pkey_buf;
+    size_t *pkey_buflen_p;
+} pkey_wrap_handler_data_t;
+
+/* A wrapped key can have max 132 bytes for an EC-p521 key */
+#define MAX_WRAPPED_KEY_SIZE    (2 * (521 / 8 + 1))
+typedef struct {
+    uint16_t version;
```
ifranzki, line 538: Are there any defines for the version?

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -489,6 +519,467 @@ static CK_RV cleanse_attribute(TEMPLATE *template,
...
+    uint8_t res0[16];
+    uint32_t cpacf_key_type;
```
ifranzki, line 540: are there defines for the CPACF type?

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -489,6 +519,467 @@ static CK_RV cleanse_attribute(TEMPLATE *template,
...
+
+    ret = get_ep11_target_for_apqn(adapter, domain, &target, XCP_MFL_PROBE);
+    if (ret != CKR_OK)
+        goto done;
+
+    /* Create the protected key via CKM_IBM_CPACF_WRAP */
+    ret = dll_m_WrapKey(data->secure_key, data->secure_key_len,
+                        NULL, 0, NULL, 0, &mech,
+                        data->pkey_buf, data->pkey_buflen_p,
+                        target | XCP_TGTFL_SET_SCMD);
```
ifranzki, line 575: Is this really right? You or XCP_TGTFL_SET_SCMD to the target?  Dependent on the Ep11 library version, target is either a numeric value (handle), or a pointer (casted to uint64_t). I guess since protected keys only work with a host lib of >= 3 it should always be a handle, but I still don't think that this could be or'ed to the target handle. Shouldn't this go to the flags passed to get_ep11_target_for_apqn (aside with XCP_MFL_PROBE)?

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -1558,6 +2043,18 @@ CK_RV ep11tok_init(STDLL_TokData_t * tokdata, CK_SLOT_ID SlotNumber,
...
     }
 
+    if (!ep11tok_pkey_option_disabled(tokdata)) {
+        rc = ep11tok_pkey_get_firmware_mk_vp(tokdata);
+        if (rc != CKR_OK) {
+            /* Could not save mk_vp in ep11_data, pkey support not available.
+             * But the token should initialize ok, even if this happens.
+             * We are just running without protected key support, i.e. the
+             * pkey_wrap_supported flag in tokdata remains off. */
+            TRACE_INFO("Could not get mk_vp, protected key support not available.\n");
```
ifranzki, line 2053: Maybe issue syslog message ?

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -7369,6 +8248,8 @@ static int read_adapter_config_file(STDLL_TokData_t * tokdata,
                                31) == 0) {
                i = 0;
                ep11_data->optimize_single_ops = 1;
+            } else if (strncmp(token, "PKEY_MODE", 9) == 0) {
+               i = 6;
```
ifranzki, line 8252: In the final ese case: Don't you need to update the error message (trace + syslog) with the new keyword you are adding?
I.e. "Expected ..... <your new keyword>, found ...."

usr/lib/ep11_stdll/ep11tok.conf
```diff
@@ -53,6 +53,31 @@
 #
 #      USE_PRANDOM
 #
+# To optimize encrypt/decrypt and sign/verify performance, a corresponding
+# protected key can be created for AES secure keys and added to the secure key
+# via IBM specific attribute CKA_IBM_OPAQUE_PKEY. This protected key is then
```
ifranzki, line 58: Do we really want to mention the CKA_IBM_OPAQUE_PKEY attribute name here, although we do not allow that attribute to bet get or modified from an application?

testcases/common/common.c
```diff
@@ -26,6 +26,14 @@ CK_BBOOL no_stop;
 CK_BBOOL no_init;
 CK_BBOOL securekey;
 
+/*
+ * The pkey flag controls whether tests shall exploit the protected key
+ * option. To allow easy integration into the CI, this is not provided as
+ * a cmdline option, but affected tests are run twice, with and without
+ * pkey option.
+ */
+CK_BBOOL pkey;
```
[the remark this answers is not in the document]
ifranzki: Also maybe initialize it FALSE here?

testcases/crypto/aes_func.c
```diff
@@ -1915,6 +1929,298 @@ CK_RV do_SignVerifyMAC(struct published_mac_test_suite_info *tsuite)
...
+        goto testcase_cleanup;
+    }
+
+    testcase_begin("Generate AES key and change CKA_IBM_PROTKEY_EXTRACTABLE");
+    testcase_new_assertion();
+
+    /* Generate a test key */
+    rc = generate_AESKey(session, 16, CK_FALSE, &keygen_mech, &h_key);
+    if (rc != CKR_OK) {
+        testcase_fail("generate_AESKey rc=%s", p11_get_ckr(rc));
```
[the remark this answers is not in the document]
ifranzki: > Thinking about it, this is actually the correct place to fail the test and `generate_AESKey` should not fail the test. Instead, it should just return a failure code.

Yes, but generate_AESKey is used that way in many testcases, so changing this would also require to change all of them.

testcases/crypto/aes_func.c
ifranzki: And like Juergen mentioned, we need a way for a failing testcase to tell if it was running with PKEY or without. As best the testcase name should be different .

usr/lib/common/obj_mgr.c
jschmidb: Done.

usr/lib/common/template.c
jschmidb: Done.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: Yes, I already tried it, but it results in ugly casts when using the byte array later

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: Done.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: This function is indeed called in object_create_skel only if class = secret or private key. It just sets cka_ibm_protkey_extractable to true, depending on cka_extractable of this key, token option, session, etc. But it would not make sense for e.g. public keys or certificates.

testcases/crypto/aes_func.c
jschmidb: Yep ... below ...

testcases/crypto/aes_func.c
jschmidb: Yes, this is a good point: done.

testcases/common/common.c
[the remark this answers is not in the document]
jschmidb: OK, I changed the output to indicate the value of the pkey variable. Hmm, ok, I can also initialize pkey to false, just be safe, although it is explicitly set before each test suite.

testcases/crypto/aes_func.c
```diff
@@ -70,8 +70,8 @@ CK_RV do_EncryptDecryptAES(struct generated_test_suite_info *tsuite)
 
         /** generate key **/
         mechkey = aes_keygen;
-        rc = generate_AESKey(session, key_lens[i], &mechkey, &h_key);
-
+        rc = generate_AESKey(session, key_lens[i], pkey ? CK_TRUE : CK_FALSE,
```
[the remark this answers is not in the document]
jschmidb: Yes, I thought about this also, but as pkey is a test option, but the boolean parameter of generate_AESKey is the value of CKA_EXTRACTABLE of this key, I wanted to keep this separately.

testcases/crypto/aes_func.c
```diff
@@ -202,8 +202,8 @@ CK_RV do_EncryptDecryptUpdateAES(struct generated_test_suite_info * tsuite)
...
```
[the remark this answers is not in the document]
jschmidb: see above ...

testcases/crypto/aes_func.c
```diff
@@ -421,7 +421,7 @@ CK_RV do_EncryptAES(struct published_test_suite_info * tsuite)
         rc = CKR_OK;
 
         /** create key handle **/
-        rc = create_AESKey(session,
+        rc = create_AESKey(session, pkey ? CK_TRUE : CK_FALSE,
```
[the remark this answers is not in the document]
jschmidb: see above ...

testcases/crypto/aes_func.c
```diff
@@ -551,7 +551,7 @@ CK_RV do_EncryptUpdateAES(struct published_test_suite_info * tsuite)
...
```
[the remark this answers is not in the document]
jschmidb: see above ...

testcases/crypto/aes_func.c
```diff
@@ -1915,6 +1929,298 @@ CK_RV do_SignVerifyMAC(struct published_mac_test_suite_info *tsuite)
...
+    CK_ULONG user_pin_len;
+    CK_SLOT_ID slot_id = SLOT_ID;
+
+    testsuite_begin("do_EncryptDecryptAESPkey");
+    testcase_rw_session();
+    testcase_user_login();
+
+    /* Skip tests if the slot doesn't support protected keys*/
+    if (!is_ep11_token(SLOT_ID)) {
+        testsuite_skip(1, "Slot %u doesn't support protected keys",
```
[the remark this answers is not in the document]
jschmidb: Yes, good point! Done.

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: ok, if we keep the function pkey specific, then this makes sense, but I would still prefer to make it generic (see other comments about this topic). The way it is now it is too much specific for pkey, and has too much internal knowledge about how pkey works in common code.

testcases/crypto/aes_func.c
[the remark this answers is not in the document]
ifranzki: ???? but pkey is a boolean variable, why not simply pass it to the function? 
`pkey ? CK_TRUE : CK_FALSE` really is a no-op anyway.

usr/lib/common/object.c
jschmidb: ok, let's leave this for now.

usr/lib/common/pkey_utils.c
jschmidb: Well, it gets clearer when looking at the pkey_aes_cbc function below, where we have an additional iv field. The CPACF param block is different for the various functions, but for me it makes sense to code it in a very similar way. As the parm block goes to CPACF, I would always make it packed.

usr/lib/common/pkey_utils.c
jschmidb: Hm yes .... I would prefer it this way ...

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: No, at least not yet. These "key type" values are part of the specification of the CKM_IBM_CPACF_WRAP mechanism, as provided by zFirmware. They provide a means to check if we get what we wanted. So they are not really "key types", like e.g. AES, but more like "indications" of certain groups of keys. Sure, I can introduce constants for them, but as they are only used here, I thought it's sufficient to explain their meaning in the comment.

usr/lib/common/object.c
jschmidb: Because public keys are never used via protected keys. When e.g. performing a "verify", I take the public key attribute(s) of a given secure key and do the operation via "non-protected-key" CPACF functions. I mean, yes, we can allow CKA_IBM_PROTKEY_EXTRACTABLE = true for public keys, but they will never be used as such. Also, public keys will never get a CKA_IBM_OPAQUE_PKEY attribute.

usr/lib/common/tok_spec_struct.h
jschmidb: I like option 1 and also already thought about that: introduce a generic routine that allows any token add/modify any attributes, but pkey is just the first one. But there is some danger that such a generic routine would get overloaded. Well, on the other hand we dont' want to introduce one separate routine for separate attributes that might come in future.
Option 2 is more or less what the code currently does, so let me think about a better name and leave it for now.

usr/lib/common/object.c
jschmidb: Yes, that was my intention

usr/lib/common/object.c
ifranzki: OK, but then your code in key_object_validate_attribute and key_object_set_default_attributes is inconsistent. key_object_set_default_attributes sets it for ALL key objects, so either that attribute is valid for ALL key objects, or not. If it is just valid for certain key objects, then use the according xxx_key_object_validate_attribute / xxx_key_object_set_default_attributes functions for checing the attribute.

And, again, I would prefer to make this token specific callback function as generic as possible, i.e. call it for ALL objects, as discussed in other comments, too.

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: Yes, please introduce defines, to have readable code. The names of the defines should tell the reader what this value means.

usr/lib/common/pkey_utils.h
jschmidb: Sure, done.

usr/lib/common/pkey_utils.h
jschmidb: Done.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: Done.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: No. There is currently only one version 0x0001 supported/returned by zFirmware. But I see what you mean ... maybe it's better to also check this version field in routine ep11tok_pkey_skey2pkey() before extracting the protected key blob. I guess that the length of the whole wrapped_key_t struct would also change when they ever introduce a new version. So the check must be done together with the length check ....

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: see above ...

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: _ok, I added the check and introduced a new constant `EP11_WRAPPED_KEY_VERSION_1       0x0001`

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: Puh ... we really list all the keywords ... yes, in this case PKEY_MODE should also be listed. Done.

usr/lib/ep11_stdll/ep11tok.conf
jschmidb: Hmmm ... I tried to explain a bit how all this happens. But yes, as we will document the protected key option in our ep11 book anyway, we may leave this info out here. Done.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: Done.

testcases/crypto/aes_func.c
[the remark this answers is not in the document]
jschmidb: ok, done.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: The syslog message is now issued after returning from ep11tok_pkey_get_firmware_mk_vp.

usr/lib/common/object.c
jschmidb: Done.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: Yes, the target needs the XCP_TGTFL_SET_SCMD flag here. The ep11lib returns an error when leaving this out. The flag is defined in ep11.h as "protected key special command".

usr/lib/common/key.c
jschmidb: Hmm, this is indeed a good question ... I added a check to make sure basetmpl does not already contain CKA_VALUE_LEN, before adding it to tmpl (which is later merged into basetmpl). However, in every case I can think of, CKA_VALUE_LEN should in fact contain the correct length of CKA_VALUE.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: OK, I'm now calling token_specific_set_pkey_attr for all objects and I also added class and mode to its parm list. However, the two parms are currently unused.

ifranzki: > However, the two parms are currently unused.

Don't you need to check the class in the implementation in the EP11 token? You want to add the PKEY attribute only for key objects, but not to all objects (although the function is called for all objects). So at least the class parameter should be used, right?

And since this function is now generic (I like that), can we change the name and remove 'pkey' from it s name? I.e. token_specific_set_attrs_for_new_object or something like that?

usr/lib/common/obj_mgr.c
jschmidb: I agree, but the difficulty here is to separate the changes into meaningful commits at all ... I already put the new files pkey_utils.c/.h into the ep11-specific commit, because they are used here although they are common code. In theory we would have to merge the common-code commit with the ep11-specific commit at all. So if you want, I move obj_mgr.c to ep11-specific, but for me it's no problem to introduce some infrastructure in one commit and exploit it in another one. What do you think?

ifranzki: No, leave it as is, just remember for the future. The point is that you should make the separation into commits so that the Reviewers that don't already know what you are doing are able to understand it just from your commits. If you ever try to commit something into the Linux kernel and add a new function but don't use it, don't wonder if they reject it right away.....

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: I updated the return code handling of routine pkey_update_and_save to make sure it returns 0 only if all internal actions (unlock/relock, update attr) succeeded. There was in fact one case possible, where update attrs would fail, but the relocking succeeded afterwards, leading to a ret = 0. Regarding  h_opaque_2_blob: the flow is always: h_opaque_2_blob (obtains read lock) -> ep11tok_pkey_usage_ok, which may unlock/relock -> object_put to release the initial read-lock.

ifranzki: That's not the point I was making. 

Once you unlock/relock the object, the objects attributes can be modified by other threads. This can affect ALL attributes of the object, not just the PKEY related attributes that you care of here. 

h_opaque_2_blob returns a pointer to the value of the CKA_IBM_OPAQUE attribute of the object. Once the object has been unlocked/relocked this pointer may no longer be valid, if the CKA_IBM_OPAQUE attribute has been changed by another thread. The code in the EP11 fallback DOES use that pointer, and thus may segfault if the pointer is no longer valid.

So you MUST ensure that you obtain the pointer to the key blob freshly, if you did unlock/relock the object, but then afterwards still go the EP11 fallback path.

usr/lib/common/key.c
```diff
@@ -335,6 +351,9 @@ CK_RV key_object_validate_attribute(TEMPLATE *tmpl, CK_ATTRIBUTE *attr,
          */
         TRACE_ERROR("%s\n", ock_err(ERR_ATTRIBUTE_READ_ONLY));
         return CKR_ATTRIBUTE_READ_ONLY;
+    case CKA_IBM_PROTKEY_EXTRACTABLE:
```
ifranzki, line 354: The way it is currently, CKA_IBM_PROTKEY_EXTRACTABLE can be set or modified with out any restriction. So you can use C_SetAttributeValues after object creation to change the value of CKA_IBM_PROTKEY_EXTRACTABLE. 
Is this intended? 

The PKCS#11 CKA_EXTRACTABLE attribute is restricted in how it can be modified. Per PKCS#11 the CKA_EXTRACTABLE can be modified after object creation with C_SetAttributeValues (footnote 8), but it also has footnote 12 saying: "Attribute cannot be changed once set to CK_FALSE. It becomes a read only attribute".

Do we want to also restrict CKA_IBM_PROTKEY_EXTRACTABLE that way?

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: Yes, I'm now checking the class inside the token_specific routine. The routine only considers secret and private keys.

usr/lib/common/tok_spec_struct.h
jschmidb: The routine name is now: token_specific_set_attrs_for_new_object.

usr/lib/common/pkey_utils.c
jschmidb: Routine ep11tok_pkey_usage_ok is now reworked and renamed to ep11tok_pkey_check. (The second routine, which was ep11tok_pkey_usage_ok2, is now renamed to ep11tok_pkey_usage_ok).

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: See rework in all routines calling ep11tok_pkey_check. We fall back into the ep11 path only if no attempt was made to create a pkey, i.e. no unlock/relock occurred.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: see above.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: See places where ep11tok_pkey_check is called.

usr/lib/common/key.c
jschmidb: As this is an IBM internal attribute, I'm currently leaving it up to the ep11lib to decide which modification are valid. See e.g. testcase do_SetAttributeValuesPkey in aes_func.c. The ep11lib allows to change CKA_IBM_PROTKEY_EXTRACTABLE from true to false, but not vice versa.

ifranzki: OK, so the EP11 library does in fact enforce the same restrictions for CKA_IBM_PROTKEY_EXTRACTABLE as for  CKA_EXTRACTABLE. 

However, since key_object_validate_attribute is a common code function, and thus the CKA_IBM_PROTKEY_EXTRACTABLE can be used with all tokens, not only EP11, I think we should enforce the restrictions in common code as well. We want that the attributes behave the same across all tokens. This won't change anything for EP11, but for the other tokens.

All you need to do is to use the following code for the `case CKA_IBM_PROTKEY_EXTRACTABLE`: block instead of just returning CKR_OK:

```
            CK_BBOOL value;

            if (attr->ulValueLen != sizeof(CK_BBOOL) || attr->pValue == NULL) {
                TRACE_ERROR("%s\n", ock_err(ERR_ATTRIBUTE_VALUE_INVALID));
                return CKR_ATTRIBUTE_VALUE_INVALID;
            }

            value = *(CK_BBOOL *) attr->pValue;
            if (mode != MODE_CREATE && mode != MODE_DERIVE &&
                 mode != MODE_KEYGEN && mode != MODE_UNWRAP &&
                 value != FALSE) {
                TRACE_ERROR("%s\n", ock_err(ERR_ATTRIBUTE_READ_ONLY));
            }
           return CKR_OK;
```

jschmidb: Done.

p-steuer (approved):

usr/lib/ep11_stdll/ep11_specific.c
p-steuer: Im not sure about LOG_WARNING. We had the discussion if libraries should write to the syslog at all multiple times. Idk what openCryptoki does in similar situations, however, id prefer if all syslogging would be at debug level.

usr/lib/ep11_stdll/ep11_specific.c
p-steuer: I think unfortunately the pkcs11 types fail at constifying things, so if thats the case here, id rather leave it as it ist. Theres not sense in putting a const and later forcing a cast.

usr/lib/common/pkey_utils.c
p-steuer: Focus on readability and leave optimizations to the compiler ... I would only consider doing that kind of stuff if its in an absolutely performance critical code path.

testcases/crypto/aes_func.c
```diff
@@ -1915,6 +1929,298 @@ CK_RV do_SignVerifyMAC(struct published_mac_test_suite_info *tsuite)
...
+
+
+    testsuite_begin("do_SetAttributeValuesPkey");
+    testcase_rw_session();
+    testcase_user_login();
+
+    /* Skip tests if the slot doesn't support protected keys*/
+    if (!is_ep11_token(SLOT_ID)) {
+        testsuite_skip(1, "Slot %u doesn't support protected keys",
+                       (unsigned int) slot_id);
```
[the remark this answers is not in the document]
ifranzki: This one still seems to be unchanged. 

You use SLOT_ID in the if statement, but slot_id in the testsuite_skip...
The same can be found at other places as well.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -5546,8 +6174,26 @@ CK_RV ep11tok_sign(STDLL_TokData_t * tokdata, SESSION * session,
...
+
+    rc = ep11tok_pkey_check(tokdata, session, key_obj, &ctx->mech);
+    switch (rc) {
+    case CKR_OK:
+        rc = sign_mgr_sign(tokdata, session, length_only, ctx, in_data,
+                           in_data_len, signature, sig_len);
+        goto done;
+    default:
+        goto done;
+    }
```
ifranzki, line 6196: There is something wrong here.
All switch cases end with got done. So it will never ever go to the EP11 path (below the switch).  But this can't be what you want. 

If in ep11tok_sign_init() you got CKR_FUNCTION_NOT_SUPPORTED and thus took the EP11 way, you also need to take the EP11 way here. 

This applies to all similar places, (verify, etc).

Please also run test without PKEY option enabled. I would suppose that those cases would currently not work anymore.

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: I am still not fully convinced that this is correct.

target (to which the flag is or'ed) is a target identifyier built by get_ep11_target_for_apqn(). Dependent on if you run with an new or an old EP11 library (i.e. old or new targethandling), this is either a numeric value (where it might be OK to or the flat to), or an address casted to target_t. I doubt that you simply should or a flag into an address.....This will most likely seggfault.

So looks like this support only works with the new target handling, i.w. new EP11 library version.  Is there a check that ensures that this code only is executed when running with the new target handling?

testcases/crypto/aes_func.c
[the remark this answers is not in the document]
jschmidb: Good catch! I'm counting over 12 such occurrences in aes_func.c. In all these (old) functions there is a definition:
`CK_SLOT_ID slot_id = SLOT_ID;`
The new function do_EncryptDecryptAESPkey just does the same. So if I correct this here, I would also correct all other occurrences

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: Hm yes ... the CPACF_WRAP mechanism works only with the currently latest ep11 lib 3.0.1. At token init, the protected key of a test key is created to save the firmware master key vp. This is only done if the CPACF_WRAP mechanism is supported (checked via ep11tok_is_mechanism_supported). So we never reach routine ep11tok_pkey_skey2pkey and in turn ep11tok_pkey_wrap_handler when running with an old library.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: Ah yes, we talked about the cases where an ep11 fallback is possible. Being in a series of sign/update or verify/update, or even just sign/verify, via pkey, we cannot fall back to ep11, because the init was done for pkey alone. 
ep11tok_sign_init() and ep11tok_verify_init() have a different switch construct, including a case for the ep11 fallback: 
```
    switch (rc) {
    case CKR_OK:
        rc = sign_mgr_init(tokdata, session, ctx, mech, recover_mode, key);
        free(ep11_sign_state);
        goto done;
    case CKR_FUNCTION_NOT_SUPPORTED:
        /* ep11 fallback */
        break;
    default:
        goto done;
    }

```
So yes, if an internal error occurs in sign, sign/update etc. we cannot recover. But I agree, as I just removed the middle case CKR_FUNCTION_NOT_SUPPORTED this looks a bit weird.  I can change this to like if (rc == CKR_OK) call sign/verify mgr else goto done.

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: Hm, well, whether the CPACF_WRAP mech is supported or not depends on the hardware, not so much on the EP11 library level. So it can happen that you run with new hardware (that supports CPACF_WRAP), but using an old EP11 library. 
However, ep11tok_is_mechanism_supported() seems to check the library version, so this should prevent it from being used with an older library. So this looks OK.
But maybe add a comment to this function about requiring a EP11 library >= V3.

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: No you did not get my point. 
The point is that with your current code it NEVER ever goes to the EP11 path. 
It either goes to the PKEY path, or it goes to done (i.e. error). But you never go to the EP11 path, not even if PKEY is disabled, etc. A "if (rc == CKR_OK) call sign/verify mgr else goto done" would not change that.

I guess you simply need to add the CKR_FUNCTION_NOT_SUPPORTED case back in, doing a **break**, so that it falls through to the EP11 path, if PKEY is not supported/configured.

The only thing left is then if during init it gets CKR_OK, but in update it gets CKR_FUNCTION_NOT_SUPPORTED (or vice versa). Not sure if this can ever happen (in real life). This would lead to situations that probably cause other errors.....

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: Done.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: Yes, this was overoptimized. I added the CKR_FUNCTION_NOT_SUPPORTED case back in.
Done.

testcases/crypto/aes_func.c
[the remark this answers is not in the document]
jschmidb: I changed the check is_ep11_token(SLOT_ID) to is_ep11_token(slot_id) at all places. 
Done.

ifranzki:
When I run you code on my system, that has CEX6 as well as CEX7 EP11 cards, many testcases such as AES, 3DES, RSA, etc fail with 
```
* TESTCASE do_EncryptAES BEGIN AES_ECB Encryption with published test vector 0 and pkey=0.
* TESTCASE create_AESKey ERROR (./testcases/common/common.c:163)) C_CreateObject rc=CKR_ATTRIBUTE_TYPE_INVALID
* TESTCASE do_EncryptAES ERROR (testcases/crypto/aes_func.c:449)) C_CreateObject rc=CKR_ATTRIBUTE_TYPE_INVALID
* TESTCASE do_EncryptAES ERROR (testcases/crypto/aes_func.c:524)) C_DestroyObject rc=CKR_OBJECT_HANDLE_INVALID
```
The trace shows this sequence:
```
02/19/2021 11:00:46 4047309 [usr/lib/api/api_interface.c:495 api] INFO: C_CreateObject
02/19/2021 11:00:46 4047309 [usr/lib/api/api_interface.c:506 api] INFO: Valid Session handle id: 1
02/19/2021 11:00:46 4047309 [usr/lib/ep11_stdll/ep11_specific.c:1661 ep11tok] INFO: rawkey_2_blob encrypt ksize=0x10 clen=0x20 rc=0x0
02/19/2021 11:00:46 4047309 [usr/lib/ep11_stdll/ep11_specific.c:1688 ep11tok] ERROR: rawkey_2_blob unwrap blen=8192 rc=0x12
02/19/2021 11:00:46 4047309 [usr/lib/ep11_stdll/ep11_specific.c:3334 ep11tok] ERROR: token_specific_object_add rawkey_2_blob rc=0x12 blobsize=0x2000
```
I guess that older cards don't accept the CKA_IBM_PROTKEY_EXTRACTABLE attribute in EP11 calls. Note that I use the default config in ep11tok.conf. I guess you must filter out the CKA_IBM_PROTKEY_EXTRACTABLE attribute from any templates that you pass to EP11 if there is hardware configured that don't support PKEY (add a version check i.e. via ep11tok_is_mechanism_supported()).

testcases/crypto/aes_func.c
```diff
@@ -1929,6 +2006,308 @@ CK_RV do_SignVerifyMAC(struct published_mac_test_suite_info *tsuite)
...
+
+    /* Encrypt some data with this key object */
+    rc = funcs->C_EncryptInit(session, &encr_mech, h_key);
+    if (rc != CKR_OK) {
+        testcase_fail("C_EncryptInit rc=%s", p11_get_ckr(rc));
+        goto testcase_cleanup;
+    }
+
+    memset(output, 0, sizeof(output));
+    rc = funcs->C_Encrypt(session, input, sizeof(input), output, &outlen);
```
ifranzki, line 2170: outlen must be initialized (to sizof(output)). 

I get the following when its uninitialzed:
```
* TESTCASE do_EncryptDecryptAESPkey BEGIN Generate token key object and encrypt
* TESTCASE do_EncryptDecryptAESPkey FAIL (testcases/crypto/aes_func.c:2172) C_Encrypt rc=CKR_BUFFER_TOO_SMALL

```

testcases/crypto/aes_func.c
```diff
@@ -1929,6 +2006,308 @@ CK_RV do_SignVerifyMAC(struct published_mac_test_suite_info *tsuite)
...
+
+    /* Decrypt encrypted data with the retrieved token object */
+    rc = funcs->C_DecryptInit(session, &encr_mech, h_key2);
+    if (rc != CKR_OK) {
+        testcase_fail("C_DecryptInit rc=%s", p11_get_ckr(rc));
+        goto testcase_cleanup;
+    }
+
+    memset(decrypted, 0, sizeof(decrypted));
+    rc = funcs->C_Decrypt(session, output, sizeof(input), decrypted, &decrlen);
```
ifranzki, line 2240: decrlen must be initialized (to sizeof(decrypted)).

testcases/crypto/aes_func.c
```diff
@@ -1929,6 +2006,308 @@ CK_RV do_SignVerifyMAC(struct published_mac_test_suite_info *tsuite)
...
+    testcase_new_assertion();
+
+    /* Retrieve the key from repository */
+    rc = funcs->C_FindObjectsInit(session, find_tmpl, find_tmpl_len);
+    if (rc != CKR_OK) {
+        testcase_fail("C_FindObjectsInit rc=%s", p11_get_ckr(rc));
+        goto testcase_cleanup;
+    }
+
+    rc = funcs->C_FindObjects(session, &h_key2, 2, &count);
```
ifranzki, line 2203: h_key2 is just one object handle, so you can't pass 2 as array size to C_FindObjects.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -5571,10 +6230,32 @@ CK_RV ep11tok_sign_update(STDLL_TokData_t * tokdata, SESSION * session,
...
         return CKR_OK;
 
+    rc = h_opaque_2_blob(tokdata, ctx->key, &keyblob, &keyblobsize, &key_obj,
+                          READ_LOCK);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("%s h_opaque_2_blob, rc=0x%lx\n", __func__, rc);
+        return rc;
+    }
+
+    rc = ep11tok_pkey_check(tokdata, session, key_obj, &ctx->mech);
```
ifranzki, line 6247: To prevent the (theoretical) possibility that ep11tok_pkey_check() returns something else on the sign update of final call than it returned on the init call, we could add a boolean field in the SIGN_VERIFY_CONTEXT (and the other contexts as well).  Instead of calling ep11tok_pkey_check() in the sign update/final, we would just use that boolean to determine which path to take (PKEY or EP11). That way we ensure that we always take the same path that we took at init. 

As of today, ep11tok_pkey_check will most likely always return the same, but it could be that due to APQNs being disabled or getting offline, or new but older APQNs become online, that CKM_IBM_CPACF_WRAP as of a sudden may not be supported any more. Once we dynamically react on APQNs going away or coming online, this is definitely a possibility. So ep11_data->pkey_wrap_supported could change during run time..... IF this happens just in between an init and an update call this might lead to problems.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -7545,6 +8570,30 @@ static int read_adapter_config_file(STDLL_TokData_t * tokdata,
...
+        } else if (i == 6) {
+            /* expecting pkey_mode */
+            if (token == NULL) {
+                rc = APQN_FILE_UNEXPECTED_END_OF_FILE;
+                OCK_SYSLOG(LOG_ERR,"%s: Error: Unexpected end of file found"
+                           " in config file '%s', expected pkey_mode 0 .. 3\n",
+                           __func__, fname);
+                break;
+            }
+            if (strncmp(token, "DISABLED", 8) == 0)
```
ifranzki, line 8582: You never explicitely initialize ep11_data->pkey_mode, so it is 0 by default, because ep11_data is allocated using calloc, and thus is cleared with zeros. 
So this means it defaults to PKEY_MODE_DISABLED. Shouldn't it default to PKEY_MODE_DEFAULT instead? 
If so, you should probably initialize it to PKEY_MODE_DEFAULT  somewhere in ep11tok_init.

testcases/crypto/aes_func.c
jschmidb: Done.

testcases/crypto/aes_func.c
jschmidb: Done.

testcases/crypto/aes_func.c
jschmidb: Done.

jschmidb:
I added an additional check in aes_func, testcase do_SetAttributeValuesPkey. On an older system we get CKR_ATTRIBUTE_TYPE_INVALID instead of CKR_ATTRIBUTE_READ_ONLY when trying to change CKA_IBM_PROTKEY_EXTRACTABLE from false to true.

p-steuer:
hmm looks like you just introduced some build errors with the latest push.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -489,6 +519,519 @@ static CK_RV cleanse_attribute(TEMPLATE *template,
...
+
+    if (data->wrap_was_successful)
+        goto done;
+
+    ret = get_ep11_target_for_apqn(adapter, domain, &target, XCP_MFL_PROBE);
+    if (ret != CKR_OK)
+        goto done;
+
+    /* Create the protected key via CKM_IBM_CPACF_WRAP */
+    ret = dll_m_WrapKey(data->secure_key, data->secure_key_len,
```
ifranzki, line 582: I guess here is something missing. The way it currently is will cause failures when STRUCT_SESSION mode or VHSM_MODE is enabled for the EP11 token. Will do some further tests and let you know offline what changes are required..

ifranzki: Well, no, all is OK. dll_m_WrapKey does not have/need a pin blob, so no need to do anything special for STRICT_SESSION mode or VHSM_MODE.

Please ignore above comment ;-)

ifranzki:
> I added an additional check in aes_func, testcase do_SetAttributeValuesPkey. On an older system we get CKR_ATTRIBUTE_TYPE_INVALID instead of CKR_ATTRIBUTE_READ_ONLY when trying to change CKA_IBM_PROTKEY_EXTRACTABLE from false to true.

Is this because dll_m_SetAttributeValue in token_specific_set_attribute_values returns CKA_IBM_PROTKEY_EXTRACTABLE ? 

I am not sure if this ist the right way to handle it. We might want that CKA_IBM_PROTKEY_EXTRACTABLE can always be set in the OCK object, independent of the underlying EP11 version. You just don't set it in the  EP11 blob, if EP11 does not support this. So you need to check if it is supported inside token_specific_set_attribute_values and only pass it to dll_m_SetAttributeValue if it is supported.  However, you should not return an error, because this would hinder the attribute to be set in the OCK object also.

Lets talk offline about this if you like....

jschmidb:
> 
> 
> > I added an additional check in aes_func, testcase do_SetAttributeValuesPkey. On an older system we get CKR_ATTRIBUTE_TYPE_INVALID instead of CKR_ATTRIBUTE_READ_ONLY when trying to change CKA_IBM_PROTKEY_EXTRACTABLE from false to true.
> 
> Is this because dll_m_SetAttributeValue in token_specific_set_attribute_values returns CKA_IBM_PROTKEY_EXTRACTABLE ?
> 
> I am not sure if this ist the right way to handle it. We might want that CKA_IBM_PROTKEY_EXTRACTABLE can always be set in the OCK object, independent of the underlying EP11 version. You just don't set it in the EP11 blob, if EP11 does not support this. So you need to check if it is supported inside token_specific_set_attribute_values and only pass it to dll_m_SetAttributeValue if it is supported. However, you should not return an error, because this would hinder the attribute to be set in the OCK object also.
> 
> Lets talk offline about this if you like....

Yes, you are right. I'm now checking if the pkey option is supported, before trying to pass CKA_IBM_PROTKEY_EXTRACTABLE  to the ep11lib. I tested this on a z14 with CEX6P cards.

jschmidb:
> 
> 
> hmm looks like you just introduced some build errors with the latest push.

Fixed.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: Done.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: Done. I added a new boolean field pkey_active to the enc/decr and sign/verify contexts. This field is now checked in any encr/decr and sign/verify operations, except when initialising. At init, the field is set either to true when being in pkey mode, or false when being in ep11 mode.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -5546,6 +6249,23 @@ CK_RV ep11tok_sign_init(STDLL_TokData_t * tokdata, SESSION * session,
...
+    rc = ep11tok_pkey_check(tokdata, session, key_obj, mech);
+    switch (rc) {
+    case CKR_OK:
+        rc = sign_mgr_init(tokdata, session, ctx, mech, recover_mode, key);
+        if (rc == CKR_OK) {
+            ctx->pkey_active = TRUE;
+            free(ep11_sign_state);
+            goto done;
+        }
+        break;
```
ifranzki, line 6261: So if ep11tok_pkey_check() return d CKR_OK, but sign_mgr_init failed, the  you now go to the EP11 path? Is this intended? 
ep11tok_pkey_check() might have unlocked amd relocked the object, so keyblob obtained from the object before the object was unlocked, so it may no longer be valid. 

I guess if sign_mgr_init() failes here, you should always go to out. (and always free(ep11_sign_state) as well).

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -5746,6 +6524,22 @@ CK_RV ep11tok_verify_init(STDLL_TokData_t * tokdata, SESSION * session,
...
 
+    rc = ep11tok_pkey_check(tokdata, session, key_obj, mech);
+    switch (rc) {
+    case CKR_OK:
+        rc = verify_mgr_init(tokdata, session, ctx, mech, CK_FALSE, key);
+        if (rc == CKR_OK) {
+            ctx->pkey_active = TRUE;
+            goto done;
+        }
+        break;
```
ifranzki, line 6535: same here.
Also free ep11_sign_state here.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -6192,6 +7184,33 @@ static CK_RV ep11_ende_crypt_init(STDLL_TokData_t * tokdata, SESSION * session,
...
+        }
+        if (rc == CKR_OK) {
+            if (op == DECRYPT)
+                (&session->decr_ctx)->pkey_active = TRUE;
+            else
+                (&session->encr_ctx)->pkey_active = TRUE;
+            free(ep11_state);
+            goto done;
+        }
+        break;
```
ifranzki, line 7205: same here.

usr/lib/ep11_stdll/ep11_specific.h
```diff
@@ -163,4 +163,7 @@ CK_BBOOL ep11tok_libica_mech_available(STDLL_TokData_t *tokdata,
 void ep11tok_copy_firmware_info(STDLL_TokData_t *tokdata,
                                  CK_TOKEN_INFO_PTR pInfo);
 
+CK_BBOOL ep11tok_pkey_usage_ok(STDLL_TokData_t *tokdata, SESSION *session,
```
ifranzki, line 166: Is this really needed externally? I guess not anymore...

ifranzki: Ah yes, it is still needed, sorry.

testcases/crypto/aes_func.c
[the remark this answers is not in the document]
ifranzki: didn't you mean to change this to just pass pkey ?

testcases/crypto/aes_func.c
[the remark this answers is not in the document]
ifranzki: also here, and at all the other places....

testcases/crypto/aes_func.c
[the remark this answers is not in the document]
ifranzki: Ah its changed in the next commit, but not in this one.....

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: Done.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: Done.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: Done.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -5546,6 +6249,25 @@ CK_RV ep11tok_sign_init(STDLL_TokData_t * tokdata, SESSION * session,
...
+         * could have unlocked/relocked the obj so that the blob is no more
+         * valid.
+         */
+        free(ep11_sign_state);
+        goto done;
+    case CKR_FUNCTION_NOT_SUPPORTED:
+        /* ep11 fallback */
+        break;
+    default:
+        goto done;
```
ifranzki, line 6268: Also `free(ep11_sign_state);` here in the default case.

Same allies to the other xxx_init functions with similar switch statement.

ifranzki (approved):
looks good now.

p-steuer (approved):

PR #369, opened 2021-02-08
EP11: Check if curve is supported on EC key import via C_CreateObject

ifranzki:
When importing public or private EC keys that uses unsupported curves (i.e. Edwards or Montgomery curves), it depends on the EP11 firmware level if the import fails and how it fails. Even if the import works, the imported EC key would still not be usable with any mechanisms afterwards, if the underlying firmware does not support the curve.

Check the curve type when importing EC keys in the EP11 token, and allow Edwards or Montgomery curves only if all configured APQNs support it. Fail with CKR_CURVE_NOT_SUPPORTED if not.

### DIFF
--- a/usr/lib/ep11_stdll/ep11_specific.c
+++ b/usr/lib/ep11_stdll/ep11_specific.c
@@ -219,6 +219,10 @@ static CK_RV ep11_open_helper_session(STDLL_TokData_t * tokdata, SESSION * sess,
 static CK_RV ep11_close_helper_session(STDLL_TokData_t * tokdata,
                                        ST_SESSION_HANDLE * sSession);
 
+static CK_BBOOL ep11tok_ec_curve_supported2(STDLL_TokData_t *tokdata,
+                                            TEMPLATE *template,
+                                            const struct _ec **curve);
+
 static void free_cp_config(cp_config_t * cp);
 #ifdef DEBUG
 static const char *ep11_get_cp(unsigned int cp);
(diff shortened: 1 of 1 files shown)


ifranzki:
FYI: This fixes a sporadic error in the CI

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -1921,6 +1926,11 @@ static CK_RV import_EC_key(STDLL_TokData_t * tokdata, SESSION * sess,
         return rc;
     }
 
+    if (!ep11tok_ec_curve_supported2(tokdata, ec_key_obj->template, &curve)) {
+        TRACE_ERROR("Curve not supported.\n");
+        return CKR_CURVE_NOT_SUPPORTED;
+    }
```
jschmidb, line 1932: Looks good to me in general. Maybe a small trace enhancement: can you include the curve NID into this error message? A textual representation would be even better, but I think we don't have one here.

ifranzki: Well, I don't always have the NID of the curve here. I only have the OID in the template, and thats hard to print and read as a human. On a negative return of ep11tok_ec_curve_supported2 the returned curve parameter may be NULL, so I can't always access the curve details.

juergenchrist (approved):

p-steuer (approved):

PR #371, opened 2021-02-09
SOFT: Fix problem with C_Get/SetOperationState and digest contexts

ifranzki:
In commit 46829bf986d45262ad45c782c084a3f908f4acb8 the SOFT token was changed to use OpenSSL's EVP interface for implementing SHA digest. With this change, the OpenSSL digest context (EVP_MD_CTX) was saved in the DIGEST_CONTEXT's context field. Since EVP_MD_CTX is opaque, its length is not known, so context_len
was set to 1.

This hinders C_Get/SetOperationState to correctly save and restore the digest state, since the EVP_MD_CTX is not saved by C_GetOperationState, and C_SetOperationState also can't restore the digest state, leaving a subsequent
C_DigestUpdate or C_DigestFinal with an invalid EVP_MD_CTX. This most likely produces a segfault.

Fix this by saving the md_data from within the EVP_MD_CTX after each digest operation, and restoring md_data on every operation with a fresh initialized EVP_MD_CTX.


ifranzki:
Had to add some #ifdefs for very old OpenSSL versions (< 1.1.0) since the functions I use to get the digest context data are not available in those older versions.

juergenchrist:
Are these versions still supported?  If yes, how do we test against them?

ifranzki:
> Are these versions still supported? If yes, how do we test against them?

No, but the travis build on ppc obviously still uses it:
https://travis-ci.org/github/opencryptoki/opencryptoki/jobs/758235345

I would be very happy if we could get rid of it at some point in time.

juergenchrist (approved):

juergenchrist:
Just realized that RHEL 7 is also shipping with openssl 1.0.2...

ifranzki:
Still did not build on ppc.... 

I really don't like that I have to disable the complete SHA support on OpenSSL < 1.1.0  versions.
Do you know if the fields that I need to access are possibly not yet opaque in OpenSSL < 1.1.0? 
Its the md_data in EVP_MD_CTX and ctx_size in EVP_MD. If would could find a way how these fields can be accessed in OpenSSL < 1.1.0 then I would not have to disable the SHA mechs....

juergenchrist:
Technically I don't see them as opaque.  In <openssl/evp.h> I see
```
struct env_md_st {
    int type;
    int pkey_type;
    int md_size;
    unsigned long flags;
    int (*init) (EVP_MD_CTX *ctx);
    int (*update) (EVP_MD_CTX *ctx, const void *data, size_t count);
    int (*final) (EVP_MD_CTX *ctx, unsigned char *md);
    int (*copy) (EVP_MD_CTX *to, const EVP_MD_CTX *from);
    int (*cleanup) (EVP_MD_CTX *ctx);
    /* FIXME: prototype these some day */
    int (*sign) (int type, const unsigned char *m, unsigned int m_length,
                 unsigned char *sigret, unsigned int *siglen, void *key);
    int (*verify) (int type, const unsigned char *m, unsigned int m_length,
                   const unsigned char *sigbuf, unsigned int siglen,
                   void *key);
    int required_pkey_type[5];  /* EVP_PKEY_xxx */
    int block_size;
    int ctx_size;               /* how big does the ctx->md_data need to be */
    /* control function */
    int (*md_ctrl) (EVP_MD_CTX *ctx, int cmd, int p1, void *p2);
} /* EVP_MD */ ;
```
and
```
struct env_md_ctx_st {
    const EVP_MD *digest;
    ENGINE *engine;             /* functional reference if 'digest' is
                                 * ENGINE-provided */
    unsigned long flags;
    void *md_data;
    /* Public key context for sign/verify */
    EVP_PKEY_CTX *pctx;
    /* Update function: usually copied from EVP_MD */
    int (*update) (EVP_MD_CTX *ctx, const void *data, size_t count);
} /* EVP_MD_CTX */ ;
```
and from <openssl/ossl_types.h> I see
```
typedef struct env_md_st EVP_MD;
typedef struct env_md_ctx_st EVP_MD_CTX;
```
So we could get to the details...

juergenchrist:
For md_size in EVP_MD there is `int EVP_MD_size(const EVP_MD* md)`.  This function basically checks that the given EVP_MD is not NULL and then returns its md_size.  But there does not seem to be anything similar for EVP_MD_CTX.

ifranzki:
> For md_size in EVP_MD there is int EVP_MD_size(const EVP_MD* md). 

Nope: EVP_MD_size returns the size of the digest, not the size of its context.

ifranzki:
With older OpenSSL versions the EVP_MD and EVP_MD_CTX fields seem not to be opaque, so I can directly access the fields. For these older OpenSSL versions I think that's of low risk, since I don't think that the structures will ever change.

usr/lib/soft_stdll/soft_specific.c
```diff
@@ -3172,19 +3163,86 @@ CK_RV token_specific_sha_init(STDLL_TokData_t *tokdata, DIGEST_CONTEXT *ctx,
...
+        ctx->context_len = EVP_MD_meth_get_app_datasize(EVP_MD_CTX_md(md_ctx));
+        ctx->context = malloc(ctx->context_len);
+        if (ctx->context == NULL) {
+            TRACE_ERROR("malloc failed\n");
+    #if OPENSSL_VERSION_NUMBER < 0x10101000L
+            EVP_MD_CTX_destroy(md_ctx);
+    #else
+            EVP_MD_CTX_free(md_ctx);
+    #endif
+            ctx->context = NULL;
```
juergenchrist, line 3204: Assignment useless in this branch since `ctx->context == NULL` is the only condition leading here.

juergenchrist (approved):

usr/lib/soft_stdll/soft_specific.c
ifranzki: fixed

p-steuer (approved):

PR #373, opened 2021-02-10
EP11: Attribute bound keys

juergenchrist:
Add support for attribute bound keys to EP11.

### DIFF
--- a/testcases/crypto/abfunc.c
+++ b/testcases/crypto/abfunc.c
@@ -1,0 +1,1088 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <memory.h>
+
+#include "pkcs11types.h"
+#include "common.c"
+#include "regress.h"
+#include "mech_to_str.h"
+#include "ec_curves.h"
+
+#define NUMKEYS 15
+
+struct keys {
+    union keysunion {
+        struct details {
+            // AB keys
+            CK_OBJECT_HANDLE aes;
+            CK_OBJECT_HANDLE des2;
+            CK_OBJECT_HANDLE des3;
+            CK_OBJECT_HANDLE generic;
+            CK_OBJECT_HANDLE rsapub;
+            CK_OBJECT_HANDLE rsapriv;
+            CK_OBJECT_HANDLE ecpub;
+            CK_OBJECT_HANDLE ecpriv;
+            CK_OBJECT_HANDLE dsapub;
+            CK_OBJECT_HANDLE dsapriv;
+            CK_OBJECT_HANDLE dhpub;
+            CK_OBJECT_HANDLE dhpriv;
+            // non-AB key
+            CK_OBJECT_HANDLE nonabpub;
+            CK_OBJECT_HANDLE nonabpriv;
+            CK_OBJECT_HANDLE nonabecpub;
+            CK_OBJECT_HANDLE nonabecpriv;
+        } d;
+        CK_OBJECT_HANDLE keys[NUMKEYS];
+    } u;
+} keys;
+
+CK_BYTE DSA_PUBL_PRIME[128] = {
+    0xba, 0xa2, 0x5b, 0xd9, 0x77, 0xb3, 0xf0, 0x2d, 0xa1, 0x65,
+    0xf1, 0x83, 0xa7, 0xc9, 0xf0, 0x8a, 0x51, 0x3f, 0x74, 0xe8,
+    0xeb, 0x1f, 0xd7, 0x0a, 0xd5, 0x41, 0xfa, 0x52, 0x3c, 0x1f,
+    0x79, 0x15, 0x55, 0x18, 0x45, 0x41, 0x29, 0x27, 0x12, 0x4a,
+    0xb4, 0x32, 0xa6, 0xd2, 0xec, 0xe2, 0x82, 0x73, 0xf4, 0x30,
+    0x66, 0x1a, 0x31, 0x06, 0x37, 0xd2, 0xb0, 0xe4, 0x26, 0x39,
+    0x2a, 0x0e, 0x48, 0xf6, 0x77, 0x94, 0x47, 0xea, 0x7d, 0x99,
+    0x22, 0xce, 0x65, 0x61, 0x82, 0xd5, 0xe3, 0xfc, 0x15, 0x3f,
+    0xff, 0xff, 0xc8, 0xb9, 0x4f, 0x37, 0xbf, 0x7a, 0xa6, 0x6a,
+    0xbe, 0xff, 0xa9, 0xdf, 0xfd, 0xed, 0x4a, 0xb6, 0x83, 0xd6,
+    0x0f, 0xea, 0xf6, 0x90, 0x4f, 0x12, 0x8e, 0x09, 0x6e, 0x3c,
+    0x0a, 0x6d, 0x2e, 0xfb, 0xb3, 0x79, 0x90, 0x8e, 0x39, 0xc0,
+    0x86, 0x0e, 0x5d, 0xf0, 0x56, 0xcd, 0x26, 0x45
+};
+
+CK_BYTE DSA_PUBL_SUBPRIME[20] = {
+    0x9f, 0x3d, 0x47, 0x13, 0xa3, 0xff, 0x93, 0xbb, 0x4a, 0xa6,
+    0xb0, 0xf1, 0x7e, 0x54, 0x1e, 0xba, 0xf0, 0x66, 0x03, 0x61
+};
+
+
+CK_BYTE DSA_PUBL_BASE[128] = {
+    0x1a, 0x5b, 0xfe, 0x12, 0xba, 0x85, 0x8e, 0x9b, 0x08, 0x86,
+    0xd1, 0x43, 0x9b, 0x4a, 0xaf, 0x44, 0x31, 0xdf, 0xa1, 0x57,
+    0xd8, 0xe0, 0xec, 0x34, 0x07, 0x4b, 0x78, 0x8e, 0x3c, 0x62,
+    0x47, 0x4c, 0x2f, 0x5d, 0xd3, 0x31, 0x2c, 0xe9, 0xdd, 0x59,
+    0xc5, 0xe7, 0x2e, 0x06, 0x40, 0x6c, 0x72, 0x9c, 0x95, 0xc6,
+    0xa4, 0x2a, 0x1c, 0x1c, 0x45, 0xb9, 0xf3, 0xdc, 0x83, 0xb6,
+    0xc6, 0xdd, 0x94, 0x45, 0x4f, 0x74, 0xc6, 0x55, 0x36, 0x54,
+    0xba, 0x20, 0xad, 0x9a, 0xb6, 0xe3, 0x20, 0xf2, 0xdd, 0xd3,
+    0x66, 0x19, 0xeb, 0x53, 0xf5, 0x88, 0x35, 0xe1, 0xea, 0xe8,
+    0xd4, 0x57, 0xe1, 0x3d, 0xea, 0xd5, 0x00, 0xc2, 0xa4, 0xf5,
+    0xff, 0xfb, 0x0b, 0xfb, 0xa2, 0xb9, 0xf1, 0x49, 0x46, 0x9d,
+    0x11, 0xa5, 0xb1, 0x94, 0x52, 0x47, 0x6e, 0x2e, 0x79, 0x4b,
+    0xc5, 0x18, 0xe9, 0xbc, 0xff, 0xae, 0x34, 0x7f
+};
+
+CK_BYTE DH_PUBL_PRIME[128] = {
+    0xd5, 0xb1, 0xaa, 0x6a, 0x3b, 0x85, 0x50, 0xf0, 0xe2,
+    0xea, 0x6b, 0xec, 0x26, 0x3b, 0xe0, 0xbf, 0x7a, 0x82,
+    0x45, 0x1b, 0xa8, 0x0a, 0x54, 0x2e, 0x14, 0x2c, 0xc2,
+    0x58, 0xb1, 0xf5, 0x59, 0xec, 0x7d, 0x16, 0x9e, 0x00,
+    0x62, 0xb3, 0xa7, 0xdc, 0x38, 0x6f, 0x64, 0x40, 0xfc,
+    0x0d, 0x3e, 0x0b, 0x66, 0x13, 0x5e, 0xa5, 0x84, 0x90,
+    0x26, 0x62, 0xcf, 0x5a, 0x14, 0x72, 0x2d, 0x1b, 0x37,
+    0x7e, 0x8a, 0x4b, 0xc0, 0xb7, 0xf2, 0x63, 0xd1, 0xaa,
+    0x51, 0x92, 0x96, 0x18, 0xae, 0xb9, 0xfd, 0x5f, 0x9d,
+    0x5d, 0xdf, 0x75, 0xa9, 0x80, 0x3d, 0xaa, 0xc2, 0x54,
+    0x00, 0xcc, 0xc1, 0x9e, 0x31, 0x4d, 0x22, 0x31, 0x44,
+    0xe9, 0x69, 0x34, 0xae, 0xcf, 0xcd, 0x6d, 0xf6, 0xe9,
+    0x37, 0x20, 0xa4, 0xd3, 0x85, 0x24, 0xff, 0x9f, 0x39,
+    0xeb, 0x78, 0xf2, 0xd1, 0xc3, 0xf9, 0x66, 0xab, 0xbd,
+    0x2d, 0xd3
+};
+
+
+CK_BYTE DH_PUBL_BASE[128] = {
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x02
+};
+
+CK_BYTE AES_KEY_VAL[] = {
+    0x2b, 0x7e, 0x15, 0x16, 0x28, 0xae, 0xd2, 0xa6,
+    0xab, 0xf7, 0x15, 0x88, 0x09, 0xcf, 0x4f, 0x3c
+};
+
+CK_BYTE DSA1024_BASE[128] = {
+    0xf7, 0xe1, 0xa0, 0x85, 0xd6, 0x9b, 0x3d, 0xde, 0xcb, 0xbc,
+    0xab, 0x5c, 0x36, 0xb8, 0x57, 0xb9, 0x79, 0x94, 0xaf, 0xbb,
+    0xfa, 0x3a, 0xea, 0x82, 0xf9, 0x57, 0x4c, 0x0b, 0x3d, 0x07,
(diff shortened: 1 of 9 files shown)


ifranzki:
You have build errors in Travis:
https://travis-ci.org/github/opencryptoki/opencryptoki/builds/758374630
Ignore the 2 linux-x86 failures, those are unrelated. But all others should build cleanly.

usr/lib/common/p11util.c
```diff
@@ -622,3 +623,33 @@ void p11_attribute_trim(CK_ATTRIBUTE *attr)
         }
     }
 }
+
+/* is_keytype_symmetric() - Check whether a key type corresponds to a
+ * symmetric key.
+ * @type key type to check
+ */
+CK_BBOOL is_keytype_symmetric(CK_KEY_TYPE type)
```
jschmidb, line 631: You could also just check CK_OBJECT_CLASS for CKO_SECRET_KEY

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -6446,34 +6888,51 @@ CK_RV ep11tok_unwrap_key(STDLL_TokData_t * tokdata, SESSION * session,
...
     }
+    if (isab) {
+        rc = check_ab_kek_type(kobj->template, CKR_UNWRAPPING_KEY_TYPE_INCONSISTENT);
+        if (rc != CKR_OK)
+            goto error;
+        /* AB unwrapping is taking care of all the boolean usage flags.
+         * No need to check attributes, just need to duplicate the array. */
+        if ((rc = dup_attribute_array(attrs,attrs_len,
+                                      &new_attrs, &new_attrs_len)))
+            goto error;
```
jschmidb, line 6905: For better reading I would add ( ...  != CKR_OK) or code it analog to rc = check_ab_kek_type above with a separate rc check in the next line

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -6329,6 +6758,13 @@ CK_RV ep11tok_wrap_key(STDLL_TokData_t * tokdata, SESSION * session,
         goto done;
     }
 
+    rc = check_ab_wrap(tokdata, &sign_blob, &sign_blob_len, &sobj,
+                       key_obj, wrap_key_obj, mech);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("AB wrapping check failed.\n");
```
jschmidb, line 6764: Why not adding the rc value to the trace output ...

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -489,6 +491,394 @@ static CK_RV cleanse_attribute(TEMPLATE *template,
...
+
+static CK_RV force_ab_sensitive(CK_ATTRIBUTE_PTR *p_attrs, CK_ULONG *p_attrs_len,
+                                CK_KEY_TYPE type)
+{
+    CK_RV rc;
+    CK_BBOOL cktrue = TRUE, abvalue, sensitive;
+
+    rc = get_bool_attribute_by_type(*p_attrs, *p_attrs_len,
+                                    CKA_IBM_ATTRBOUND, &abvalue);
+    if (rc == CKR_OK && abvalue) {
```
jschmidb, line 822: We would save some brackets and indentation if we check for (rc != CKR_OK || !abvalue) and exit/goto end in this case.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -489,6 +491,394 @@ static CK_RV cleanse_attribute(TEMPLATE *template,
...
+    if (rc == CKR_TEMPLATE_INCOMPLETE)
+        abunwrapkey = FALSE;
+    else if (rc != CKR_OK)
+        return rc;
+    rc = get_bool_attribute_by_type(attrs, attrs_len, CKA_IBM_ATTRBOUND, &abkey);
+    if (rc == CKR_TEMPLATE_INCOMPLETE)
+        abkey = FALSE;
+    else if (rc != CKR_OK)
+        return rc;
+    if (mech->mechanism == CKM_IBM_ATTRIBUTEBOUND_WRAP) {
```
jschmidb, line 681: If we would check for (mech->mechanism != CKM_IBM_ATTRIBUTEBOUND_WRAP) and do the cleanup at line 715 here, we would save brackets and indentation ... just for better reading.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -489,6 +491,394 @@ static CK_RV cleanse_attribute(TEMPLATE *template,
...
+        return rc;
+    }
+    rc = template_attribute_get_bool(wrap_key_obj->template, CKA_IBM_ATTRBOUND, &abwrapkey);
+    if (rc == CKR_TEMPLATE_INCOMPLETE)
+        abwrapkey = FALSE;
+    else if (rc != CKR_OK) {
+        TRACE_ERROR("Invalid CKA_IBM_ATTRBOUND attribute on wrapping key\n");
+        return rc;
+    }
+    if (abwrapkey) {
```
jschmidb, line 610: If checking for (!abwrapkey) and exiting here, we would save an indentation

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -489,6 +491,394 @@ static CK_RV cleanse_attribute(TEMPLATE *template,
...
+}
+
+static CK_RV check_ab_attributes(CK_ATTRIBUTE *attrs,
+                                 CK_ULONG attrs_len,
+                                 CK_OBJECT_CLASS kc)
+{
+    CK_RV rc;
+    CK_BBOOL attrbound = FALSE, sensitive = FALSE;
+
+    if (kc != CKO_PUBLIC_KEY) {
```
jschmidb, line 567: When checking for (kc == CKO_PUBLIC_KEY) and exiting here, we would save an indentation

usr/lib/common/p11util.c
juergenchrist: Well, if we have a check that only symmetric keys have CKO_SECRET_KEY I could.  I have not found that check.  But if I use a wrong key with AB wrap I get an error that I want to catch before.  Therefore I decided to use the actual key type instead of the object class.

usr/lib/ep11_stdll/ep11_specific.c
juergenchrist: Done.  That was actually a copy and paste from existing code.

usr/lib/ep11_stdll/ep11_specific.c
juergenchrist: Done

usr/lib/ep11_stdll/ep11_specific.c
juergenchrist: Yes, but I do not see a benefit here.  I find conjunctions easier than disjunctions and since the code is equivalent I would prefer to not change it.

usr/lib/common/key.c
```diff
@@ -331,6 +331,16 @@ CK_RV key_object_validate_attribute(TEMPLATE *tmpl, CK_ATTRIBUTE *attr,
          */
         TRACE_ERROR("%s\n", ock_err(ERR_ATTRIBUTE_READ_ONLY));
         return CKR_ATTRIBUTE_READ_ONLY;
+    case CKA_IBM_ATTRBOUND:
```
ifranzki, line 334: You check the attribute here (great!), but you do not set the attribute in key_object_set_default_attributes(). I guess this is by intention? That way a key object won't have that attribute, unless the user specifies it explicitly at object creation. 
That's OK, but all the places where you try to get the value of CKA_IBM_ATTRBOUND should be able to handle the situation that the attribute is not there, and should probably assume FALSE as the default value.

usr/lib/common/p11util.c
ifranzki: I don't see it anywhere explicitly stated in the PKCS#11 standard, but all places that I have seen where CKO_SECRET_KEY is used is dealing with symmetric keys. 
The PKCS#11 standard specifies for each key type _CKK_xxx the object class to use. E.g. for CKK_AES, it specifies that the class is CKO_SECRET_KEY .

> 2.8.2 AES secret key objects
> AES secret key objects (object class CKO_SECRET_KEY, key type CKK_AES) hold AES keys.

I guess if you lookup all the key types that you check here, you will see that all of them have  class CKO_SECRET_KEY . If that's the case, then a check for class = CKO_SECRET_KEY is sufficient in your code.

usr/lib/ep11_stdll/ep11_specific.c
juergenchrist: That would add one more line.  I do not see an advantage here.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -489,6 +491,394 @@ static CK_RV cleanse_attribute(TEMPLATE *template,
...
+    }
+    if (abwrapkey) {
+        if (!abkey) {
+            TRACE_ERROR("Wrapping key is AB, but target key not\n");
+            return CKR_KEY_NOT_WRAPPABLE;
+        }
+        /* Here, only AB wrapping can be used.  Check mechanism and parameters. */
+        if (mech->mechanism != CKM_IBM_ATTRIBUTEBOUND_WRAP ||
+            sizeof(CK_IBM_ATTRIBUTEBOUND_WRAP_PARAMS) != mech->ulParameterLen) {
+            TRACE_ERROR("AB key wrapping attempt without CKM_IBM_ATTRIBUTEBOUND_WRAP");
```
ifranzki, line 618: ... or with invalid mech param

usr/lib/common/key.c
juergenchrist: That is done intentionally.  Since CKA_IBM_ATTRBOUND should be EP11-specific I did not want to set it in all cases.  And yes, not having the attribute set all the time causes some ugly code when dealing with non-present attribute values, but the code (after an update to this PR) will handle this correctly.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -489,6 +491,394 @@ static CK_RV cleanse_attribute(TEMPLATE *template,
...
+        return rc;
+    rc = get_bool_attribute_by_type(attrs, attrs_len, CKA_IBM_ATTRBOUND, &abkey);
+    if (rc == CKR_TEMPLATE_INCOMPLETE)
+        abkey = FALSE;
+    else if (rc != CKR_OK)
+        return rc;
+    if (mech->mechanism == CKM_IBM_ATTRIBUTEBOUND_WRAP) {
+        *isab = TRUE;
+        if (mech->ulParameterLen != sizeof(CK_IBM_ATTRIBUTEBOUND_WRAP_PARAMS)) {
+            TRACE_ERROR("Unwrapping AB key without CKM_IBM_ATTRIBUTEBOUND_WRAP\n");
```
ifranzki, line 684: Isn't this check here just for the mechanism parameter? If so then the trace text is wrong, and the return code should be CKR_MECHANISM_PARAM_INVALID.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -489,6 +491,394 @@ static CK_RV cleanse_attribute(TEMPLATE *template,
...
+            return CKR_MECHANISM_INVALID;
+        }
+        if (!abkey) {
+            TRACE_ERROR("CKM_IBM_ATTRIBUTEBOUND_WRAP with non-AB target\n");
+            return CKR_MECHANISM_INVALID;
```
ifranzki, line 689: Hm, is CKR_MECHANISM_INVALID really the right return code here? 
Maybe CKR_WRAPPING_KEY_TYPE_INCONSISTENT would fit better, although also not really an exact match. Or maybe CKR_TEMPLATE_INCONSISTENT?

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -489,6 +491,394 @@ static CK_RV cleanse_attribute(TEMPLATE *template,
...
+        }
+        if (!abunwrapkey) {
+            TRACE_ERROR("CKM_IBM_ATTRIBUTEBOUND_WRAP with non-AB unwrap key\n");
+            return CKR_KEY_FUNCTION_NOT_PERMITTED;
```
ifranzki, line 693: Same question as above, although CKR_KEY_FUNCTION_NOT_PERMITTED looks like a good choice. Maybe also use CKR_KEY_FUNCTION_NOT_PERMITTED  for above?

usr/lib/ep11_stdll/ep11_specific.c
juergenchrist: I don't see that.  There is an `else` attached to this `if`.

usr/lib/ep11_stdll/ep11_specific.c
juergenchrist: There is an `else` attached to this `if` capturing the cases where someone wants to perform AB-Wrapping without AB Keys.  So even if I invert the logic, I still need to do all the code in the else-branch here.  This sill not save any brackets or indentation.

usr/lib/ep11_stdll/ep11_specific.c
juergenchrist: Well, that would mean I should split the or since the exit code is different in the two cases.  Will do.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -489,6 +491,394 @@ static CK_RV cleanse_attribute(TEMPLATE *template,
...
+                             attrs[i].ulValueLen, &attr);
+        if (rc != CKR_OK) {
+            TRACE_ERROR("failed to build attribute (rc=0x%lx)\n", rc);
+            return rc;
+        }
+        rc = template_update_attribute(obj->template, attr);
+        if (rc != CKR_OK) {
+            TRACE_ERROR("%s template_update_attribute failed with rc=0x%lx\n",
+                        __func__, rc);
+            return rc;
```
ifranzki, line 799: If this has failed, then you should free attr. build_attribute allocates the attr, and template_update_attribute adds it to the list. But if template_update_attribute() fails, then attr has not been added, and thus must be freed.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -489,6 +491,394 @@ static CK_RV cleanse_attribute(TEMPLATE *template,
...
+    }
+    /* force sensitive */
+    rc = build_attribute(CKA_SENSITIVE, &cktrue, sizeof(CK_BBOOL), &attr);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("failed to build attribute (rc=0x%lx)\n", rc);
+        return rc;
+    }    
+    rc = template_update_attribute(obj->template, attr);
+    if (rc != CKR_OK)
+        TRACE_ERROR("%s failed to force template sensitive; rc=0x%08lx\n", __func__, rc);
```
ifranzki, line 810: free attr here as well.

usr/lib/ep11_stdll/ep11_specific.c
juergenchrist: Done.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -2709,6 +3096,26 @@ CK_RV token_specific_object_add(STDLL_TokData_t * tokdata, SESSION * sess,
...
+    }
+    rc = template_attribute_get_bool(obj->template, CKA_IBM_ATTRBOUND,
+                                    &attrbound);
+    if (rc == CKR_TEMPLATE_INCOMPLETE) {
+        attrbound = false;
+    } else if (rc != CKR_OK) {
+        TRACE_ERROR("Incomplete CKA_IBM_ATTRBOUND attribute for the key.\n");
+        return rc;
+    }
+    if (class != CKO_PUBLIC_KEY && attrbound) {
```
ifranzki, line 3115: How about CKO_SECRET_KEY? 
Is an attr bound public key the only key that can be created via C_CreateObject?

usr/lib/common/p11util.c
juergenchrist: Well, at the point of calling this function I only have a key type.  But since I have the template I can also extract the class.  At call side I basically need "Key is RSA or symmetric key".  So I based that purly on the key type.

testcases/crypto/abfunc.c
```diff
@@ -0,0 +1,1079 @@
...
+
+        rc = funcs->C_CancelFunction(hsess);
+        if (rc != CKR_FUNCTION_NOT_PARALLEL) {
+            return rc;
+        }
+    }
+
+    if (is_ep11_token(SLOT_ID)) {
+        testcase_setup(0);
+        rv = do_CheckMechanismInfo();
```
ifranzki, line 1071: So if CKM_IBM_ATTRIBUTEBOUND_WRAP is not supported, then you continue anyway? Shouldn't you skip the call to testdriver then? 
You could also omit the check for is_ep11_token then, since other tokens simply wont support CKM_IBM_ATTRIBUTEBOUND_WRAP and thus skip the tests.

usr/lib/common/p11util.c
juergenchrist: With these changes, the function becomes unused.  I will remove it from this commit.

testcases/crypto/abfunc.c
```diff
@@ -0,0 +1,1079 @@
...
+    return res;
+}
+
+void do_TestInvalidKeys(CK_SESSION_HANDLE session)
+{
+    CK_OBJECT_HANDLE pubkey = CK_INVALID_HANDLE, privkey = CK_INVALID_HANDLE;
+    CK_RV rc;
+
+    // Expected to fail with CKR_TEMPLATE_INCOMPLETE
+    testcase_begin("Create dilithium AB key");
```
ifranzki, line 570: Are you using Dilithium keys here for a specific reason? Dilithium keys are only supported by quite recent hardware. So this tests will be skipped if run on a system with older hardware.  Wouldn't it be more useful to choose a key type that is supported on all hardware, e.g. EC keys?

usr/lib/ep11_stdll/ep11_specific.c
juergenchrist: The idea was that the mechanism is invalid for wrapping this key.  In this case you want to wrap a non-AB key with CKM_IBM_ATTRIBUTEBOUND_WRAP and an AB wrapping key.  Now, which error do we want to deliver?  I went for the "this mechanism cannot be used for this target key", you go for the "this wrapping key cannot be used for the target key".  Both are true, but unfortunately we can only report one.

ifranzki: Yea, you are right, both makes sense. See lets keep CKR_MECHANISM_INVALID.

usr/lib/ep11_stdll/ep11_specific.c
juergenchrist: I see this as a different situation.  Here, we want to unwrap into an AB key with a non-AB unwrapping key and CKM_IBM_ATTRIBUTEBOUND_WRAP.  So the unwrapping key is here the problem.  I was thinking about CKR_UNWRAPPING_KEY_TYPE_INCONSISTENT but decided against it because it is not the key type, but the key function "AB unwrap" that is not permitted/supported by this key.

ifranzki: Yes, keep CKR_KEY_FUNCTION_NOT_PERMITTED it fits well.

usr/lib/ep11_stdll/ep11_specific.c
juergenchrist: Good find.  Fixed.

usr/lib/ep11_stdll/ep11_specific.c
juergenchrist: fixed

usr/lib/ep11_stdll/ep11_specific.c
juergenchrist: Yes, since it does not contain any secrets ;-)  Everything that contains secrets and should be AB needs special treatment since the secrets are store in encrypted for in AB (which we currently do not support for C_CreateObject).  This is also essential to really separate the key hierarchies.

testcases/crypto/abfunc.c
juergenchrist: Well, that test is the only test I found for a key type that is not supported for AB keys.  EC is supported, so I cannot use this.  I did not find any other key type supported by EP11 that does not support AB...

testcases/crypto/abfunc.c
juergenchrist: Well, skipping the call to testdriver was certainly my intention...  Fixed.

usr/lib/common/p11util.c
juergenchrist: Done.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -1036,6 +1038,408 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
...
+
+static CK_RV check_ab_kek_type(TEMPLATE *tmpl, CK_RV errval)
+{
+    CK_KEY_TYPE kektype;
+    CK_OBJECT_CLASS class;
+    CK_RV rc;
+
+    rc = template_attribute_get_ulong(tmpl, CKA_KEY_TYPE, &kektype);
+    if (rc != CKR_OK)
+        return rc;
```
ifranzki, line 1093: If template_attribute_get_ulong fails, it prints an error level trace entry. However, you can't tell from that trace entry where the call comes from. Because of that it might be useful to also print a trace entry here, where template_attribute_get_ulong is called.

Dependent on the way check_ab_kek_type is used it might be intended to not print additional trace entries. So feel free to leave it as is.....

ifranzki:
I guess you did retest after rebasing to current master? Joergs recent PR did move a few things around, not that this somehow interfered with your design.... Just to double check.

juergenchrist:
> I guess you did retest after rebasing to current master?

Yes, after rebasing, I did run all tests and they all passed as expected.

usr/lib/ep11_stdll/ep11_specific.c
juergenchrist: Good catch.  There was actually no other trace message printed on the whole part.  Fixed it for both KEY_TYPE and CLASS.

ifranzki (approved):
lgtm

p-steuer (approved):

PR #374, opened 2021-02-12
Setting SLOT=3 to use the Softtoken for testing of p11sak. Signed-off…

reumann-ibm:
Setting SLOT=3 to use the Softtoken for testing of p11sak. 

Signed-off: Matthias Reumann <matthias.reumann1@ibm.com>

### DIFF
--- a/testcases/misc_tests/p11sak_test.sh
+++ b/testcases/misc_tests/p11sak_test.sh
@@ -31,6 +31,12 @@ P11SAK_EC_PRE=p11sak-ec-pre.out
 P11SAK_EC_POST=p11sak-ec-post.out
 
 
+echo "** Setting SLOT=3 to the Softtoken unless otherwise set - 'p11sak_test.sh'"
+
+
+# setting SLOT=3 to the Softtoken
+
+SLOT=${SLOT:-3}
 
 echo "** Now generating keys - 'p11sak_test.sh'"
 


testcases/misc_tests/p11sak_test.sh
```diff
@@ -31,6 +31,13 @@ P11SAK_EC_PRE=p11sak-ec-pre.out
 P11SAK_EC_POST=p11sak-ec-post.out
 
 
+echo "** Setting SLOT=3 to the Softtoken - 'p11sak_test.sh'"
+
+
+# setting SLOT=3 to the Softtoken
+
+SLOT=3
```
ifranzki, line 39: Shouldn't we better set that outside of p11sak_test.sh, i.e. in the Makefile that calls p11sak_test.sh ? 
If you set it inside p11sak_test.sh, then you can not use p11sak_test.sh to test p11sak with another token than Soft Token. 

Or maybe only set in inside p11sak_test.sh if its not already set? That way once could set it outside of p11sak_test.sh and that would then override the internal setting.

That would be something like this (un-tested):

`: ${SLOT=3}`

juergenchrist: Use `SLOT=${SLOT:-3}` maybe?

reumann-ibm: I changed the code using the suggestion by @juergenchrist and tested it. Now the $SLOT can be defined as env variable using export SLOT=N where N is the token/slot number.

I just tested it and it seems to work.

ifranzki (approved):
Great, thanks!

p-steuer (approved):

juergenchrist:
Can we also merge this?  Otherwise we will have a weekend full of CI errors...

juergenchrist (approved):

PR #375, opened 2021-02-16
Added setting of USER_PIN in Makefile.am for testing of p11sak with t…

reumann-ibm:
…he p11sak_test.sh script.

Signed-off by: Matthias Reumann <matthias.reumann1@ibm.com>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -218,7 +218,7 @@ installcheck-local: all
 ci-installcheck: installcheck
 	killall -HUP pkcsslotd || true
 	@sbindir@/pkcsslotd
-	cd ${srcdir}/testcases && ./misc_tests/p11sak_test.sh | tee log-p11sak.txt
+	cd ${srcdir}/testcases && && PKCSLIB=@libdir@/opencryptoki/libopencryptoki.so && PKCS11_USER_PIN=$(PCKS11_USER_PIN) && ./misc_tests/p11sak_test.sh | tee log-p11sak.txt
 	killall -HUP pkcsslotd
 	@echo "done"
 endif


Makefile.am
```diff
@@ -218,7 +218,7 @@ installcheck-local: all
 ci-installcheck: installcheck
 	killall -HUP pkcsslotd || true
 	@sbindir@/pkcsslotd
-	cd ${srcdir}/testcases && ./misc_tests/p11sak_test.sh | tee log-p11sak.txt
+	cd ${srcdir}/testcases && PKCS11_USER_PIN=$(PCKS11_USER_PIN) && ./misc_tests/p11sak_test.sh | tee log-p11sak.txt
```
ifranzki, line 221: Looks good so far.

Just to double check: 

A few lines above the ock_tests.sh is called with the PINs as well as the PKCSLIB. However, p11sak_test.sh is called without PKCSLIB. So p11sak_test.sh uses the installed libopencryptoki.so, but ock_tests.sh uses the just built  libopencryptoki.so (without being installed). 

I assume that ci-installcheck is executed AFTER the install target, so that the just built libopencryptoki.so has already been installed when ci-installcheck is run? Can someone please confirm?

juergenchrist: It is of course executed after install.  Wouldn't make much sense to installcheck without install (that would be just check, which we do not use).  But I am not sure if/how p11sak finds the library.  Maybe, just to be safe, add that setting here, too.  At least a quick grep in the code shows that p11sak takes PKCSLIB securely from the environment.

ifranzki: @reumann-ibm do you plan to also pass PKCSLIB to the p11sak_test.sh? 
We should get this PR done soon, its failing daily in the CI.....

reumann-ibm: I added the PKCSLIB in the Makefile. I also tested it on my local system with PASS. Hopefully it will run through the CI now.

juergenchrist (approved):

ifranzki (approved):

p-steuer (approved):

PR #379, opened 2021-03-08
Protected key option for EC and IBM ED keys

jschmidb:


usr/lib/common/pkey_utils.c
```diff
@@ -602,3 +702,343 @@ CK_RV pkey_aes_cmac(OBJECT *key_obj, CK_BYTE *message,
...
+    if (template_attribute_get_non_empty(tmpl, CKA_EC_PARAMS, &attr) != CKR_OK) {
+        TRACE_ERROR("%s: This template does not have CKA_EC_PARAMS, cannot determine curve type.\n",
+                    __func__);
+        curve_type = curve_invalid;
+        goto done;
+    }
+    ec_params = (CK_BYTE *) attr->pValue;
+    ec_params_len = attr->ulValueLen;
+
+    if (memcmp(ec_params, p256, ec_params_len) == 0)
```
juergenchrist, line 774: Comparison only makes sense if `ec_params_len == sizeof(p256)`.  Otherwise we might end up reading parts of other variables (or curves).  Same for cases below.

usr/lib/common/pkey_utils.c
```diff
@@ -602,3 +702,343 @@ CK_RV pkey_aes_cmac(OBJECT *key_obj, CK_BYTE *message,
...
+        break;
+    case curve_p521:
+        fc = KDSA_ENCRYPTED_ECDSA_SIGN_P521 | 0x80;
+        if (rng_cb != NULL) {
+            rng_cb(param.P521.rand, sizeof(param.P521.rand));
+            fc = KDSA_ENCRYPTED_ECDSA_SIGN_P521;
+        }
+        break;
+    default:
+        /* Cannot occur */
```
juergenchrist, line 901: If it occurs anyway (e.g., due to a coding bug during addition of a new curve), we will not even notice that we have an unimplemented point causing a sign error.  Even worse, in this case we jump to `done` returning `ret` which is uninitialized so far.  So please initialize `ret` to some error and/or issue a trace call.

usr/lib/common/pkey_utils.c
```diff
@@ -161,6 +161,42 @@ static int s390_kmc(unsigned long func, void *param, unsigned char *dest,
...
+    register unsigned long r0 __asm__("0") = (unsigned long)func;
+    register unsigned long r1 __asm__("1") = (unsigned long)param;
+    register unsigned long r2 __asm__("2") = (unsigned long)src;
+    register unsigned long r3 __asm__("3") = (unsigned long)srclen;
+    unsigned long rc = 1;
+
+    __asm__ volatile(
+        "0: .insn   rre,%[__opc] << 16,0,%[__src]\n"
+        "   brc 1,0b\n" /* handle partial completion */
+        "   brc 7,1f\n"
```
juergenchrist, line 190: Not sure, but wouldn't `brc 6,1f\n` be sufficient here?  Due to the line before, `cc=3` (mask value position 1) is already covered and cannot be set here anymore.

usr/lib/common/pkey_utils.c
```diff
@@ -602,3 +702,343 @@ CK_RV pkey_aes_cmac(OBJECT *key_obj, CK_BYTE *message,
...
+        }
+        ret = CKR_FUNCTION_FAILED;
+        goto done;
+        break;
+    case 2:
+        TRACE_ERROR("%s rc from KDSA = 2\n", __func__);
+        ret = CKR_FUNCTION_FAILED;
+        goto done;
+        break;
+    case 3:
```
juergenchrist, line 937: How can this happen?  Partial completion is already taken care of by `s390_kdsa`.

usr/lib/common/pkey_utils.c
```diff
@@ -602,3 +702,343 @@ CK_RV pkey_aes_cmac(OBJECT *key_obj, CK_BYTE *message,
...
+    case 1:
+        /* public key invalid or signature corrupted. We are here if the value
+         * of the signature is greater than the value of the public key. */
+        ret = CKR_SIGNATURE_INVALID;
+        break;
+    case 2:
+        ret = CKR_SIGNATURE_INVALID;
+        break;
+    default:
+        /* KDSA partial completion */
```
juergenchrist, line 1214: Handled already in `s390_kdsa`.

usr/lib/common/pkey_utils.c
```diff
@@ -915,6 +964,136 @@ CK_RV pkey_ec_sign(OBJECT *privkey, CK_BYTE *hash, CK_ULONG hash_len,
...
+        ret = CKR_FUNCTION_FAILED;
+        goto done;
+        break;
+    case 2:
+        TRACE_ERROR("%s rc from KDSA = 2\n", __func__);
+        ret = CKR_FUNCTION_FAILED;
+        goto done;
+        break;
+    case 3:
+        TRACE_DEBUG("%s KDSA partial completion, retrying ...\n", __func__);
```
juergenchrist, line 1066: Handled by `s390_kdsa`

usr/lib/common/pkey_utils.c
```diff
@@ -1042,3 +1221,122 @@ struct {                          \
...
+    case 1:
+        /* public key invalid or signature corrupted. We are here if the value
+         * of the signature is greater than the value of the public key. */
+        ret = CKR_SIGNATURE_INVALID;
+        break;
+    case 2:
+        ret = CKR_SIGNATURE_INVALID;
+        break;
+    default:
+        /* KDSA partial completion */
```
juergenchrist, line 1332: Handled by `s390_kdsa`

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -6750,6 +6751,135 @@ static CK_RV h_opaque_2_blob(STDLL_TokData_t *tokdata, CK_OBJECT_HANDLE handle,
...
+    ctx->multi_init = FALSE;
+    ctx->multi = FALSE;
+    ctx->active = TRUE;
+    ctx->recover = FALSE;
+    ctx->pkey_active = FALSE;
+
+    rc = CKR_OK;
+
+done:
+    object_put(tokdata, key_obj, TRUE);
```
juergenchrist, line 6877: Why do we unlock the object here?  We just checked that everything is fine and then we allow the user to, e.g., remove CKA_SIGN and then sign with this key?  Shouldn't the unlocking be done in the _final function?

usr/lib/common/pkey_utils.c
jschmidb: The curve_type was already checked in the switch above, so we would already be at "done" with a bad rc. There are two "switch (curve_type)" statements in a row. But I will add the trace and rc here again, just for clarity.

usr/lib/common/pkey_utils.c
```diff
@@ -277,15 +335,48 @@ CK_RV pkey_update_and_save(STDLL_TokData_t *tokdata, OBJECT *key_obj,
...
+        TRACE_ERROR("%s: No CKA_ECDSA_PARAMS found in template, cannot determine curve\n",
+                    __func__);
         return CK_FALSE;
+    }
 
-    switch (mech->mechanism) {
+    ec_params = (CK_BYTE *) attr->pValue;
+    ec_params_len = attr->ulValueLen;
+
+    if (memcmp(ec_params, p256, ec_params_len) == 0 ||
```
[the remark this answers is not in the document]
jschmidb: Yes, absolutely! Done.

usr/lib/common/pkey_utils.c
jschmidb: Done.

usr/lib/common/pkey_utils.c
```diff
@@ -602,3 +702,343 @@ CK_RV pkey_aes_cmac(OBJECT *key_obj, CK_BYTE *message,
...
 }
+
+/**
+ * Determine the CPACF curve type from given template.
+ */
+cpacf_curve_type_t get_cpacf_curve_type(TEMPLATE *tmpl)
+{
+    const CK_BYTE p256[] = OCK_PRIME256V1;
+    const CK_BYTE p384[] = OCK_SECP384R1;
+    const CK_BYTE p521[] = OCK_SECP521R1;
```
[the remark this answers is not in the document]
jschmidb: Done.

usr/lib/common/pkey_utils.c
```diff
@@ -277,15 +335,48 @@ CK_RV pkey_update_and_save(STDLL_TokData_t *tokdata, OBJECT *key_obj,
...
  */
-CK_BBOOL pkey_op_supported_by_cpacf(int msa_level, CK_MECHANISM *mech)
+CK_BBOOL pkey_op_ec_curve_supported_by_cpacf(TEMPLATE *tmpl)
 {
-    if (!mech)
+    const CK_BYTE p256[] = OCK_PRIME256V1;
+    const CK_BYTE p384[] = OCK_SECP384R1;
+    const CK_BYTE p521[] = OCK_SECP521R1;
+    const CK_BYTE ed25519[] = OCK_ED25519;
+    const CK_BYTE ed448[] = OCK_ED448;
```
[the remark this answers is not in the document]
jschmidb: Done.

usr/lib/common/pkey_utils.c
jschmidb: OK, removed "case 3".

usr/lib/common/pkey_utils.c
jschmidb: see above. Done.

usr/lib/common/pkey_utils.c
jschmidb: Done.

usr/lib/common/pkey_utils.c
jschmidb: Done.

usr/lib/common/pkey_utils.c
juergenchrist: Sorry, did not see the first switch checked exactly the same thing.  Couldn't we simply get rid of the second switch?  Just set the `fc` in the first switch and remember the two parameters for `rng_cb`.  Then, the second switch boils down to
```
if (rng_cb != NULL) {
    rng_cb(rng_area, rng_area_size);
    fc &= ~0x80u;
}
```
Of course, more lines would go into the first switch...

Also maybe add a constant for the deterministic bit 0x80 in the KDSA function code.

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: This function obtained the lock with object_mgr_find_in_map1, so it must also release it. 
And no, OCK does not keep the lock in-between Init, Update, Final calls.

ifranzki:
Did you find a way to verify that the ECDSA/EdDSA signatures created by your CPACF code are 'correct', i.e. that a signature created by CPACF can be verified by EP11 and vice versa? 
As far as I see your tests just verify the the CPACF verify works with the CPACF signatures. So you test that you understand yourself.  But we should also make sure that the signatures are 'correct', and understood by all implementations.
You could for example create an EC key with CKA_IBM_PROTKEY_EXTRACTABLE attribute being TRUE, then perform a sign with that. Then set CKA_IBM_PROTKEY_EXTRACTABLE to FLASE on the public key, and perform the verify operation. The same the other way round, too, sign with CKA_IBM_PROTKEY_EXTRACTABLE=FALSE, and verify with CKA_IBM_PROTKEY_EXTRACTABLE=TRUE.

jschmidb:
>Did you find a way to verify that the ECDSA/EdDSA signatures created by your CPACF code are 'correct', i.e. that a signature created by CPACF can be verified by EP11 and vice versa?

Yes, I tested this by internally enforcing sign via pkey / verify via ep11 and vice versa. But I didn't add a specific testcase yet. Maybe I can do this ...

ifranzki:
You have travis builds errors on s390 (gcc and clang):

https://travis-ci.org/github/opencryptoki/opencryptoki/jobs/761950169

```
usr/lib/common/pkey_utils.c:52:32: error: expected ';' at end of declaration
    register unsigned long __nr asm("0") = doublewords - 1;
                               ^
                               ;
usr/lib/common/pkey_utils.c:54:5: error: use of undeclared identifier 'asm'
    asm volatile(".insn s,0xb2b00000,0(%1)" /* stfle */
    ^
usr/lib/common/pkey_utils.c:50:50: error: unused parameter 'list' [-Werror,-Wunused-parameter]
static inline int s390_stfle(unsigned long long *list, int doublewords)
                                                 ^
usr/lib/common/pkey_utils.c:50:60: error: unused parameter 'doublewords' [-Werror,-Wunused-parameter]
static inline int s390_stfle(unsigned long long *list, int doublewords)
                                                           ^
usr/lib/common/pkey_utils.c:843:49: error: arithmetic on a pointer to void is a GNU extension [-Werror,-Wpointer-arith]
        memcpy(param.P256.vp, pkey_attr->pValue + 32, 32);
                              ~~~~~~~~~~~~~~~~~ ^
```

and some more....

jschmidb:
> You could for example create an EC key with CKA_IBM_PROTKEY_EXTRACTABLE attribute being TRUE, then perform a sign with that. Then set CKA_IBM_PROTKEY_EXTRACTABLE to FLASE on the public key, and perform the verify operation. The same the other way round, too, sign with CKA_IBM_PROTKEY_EXTRACTABLE=FALSE, and verify with CKA_IBM_PROTKEY_EXTRACTABLE=TRUE.

Yes, changing CKA_IBM_PROTKEY_EXTRACTABLE from true to false is possible, but the other way round is not, because the attribute cannot be changed from false to true.

ifranzki:
> Yes, changing CKA_IBM_PROTKEY_EXTRACTABLE from true to false is possible, but the other way round is not, because the attribute cannot be changed from false to true.

Right, but for a public key, you can always extract the public key parts from an existing key, and create a new public key object with C_CreateObject with the CKA_IBM_PROTKEY_EXTRACTABLE of your choice.

ifranzki:
> Yes, changing CKA_IBM_PROTKEY_EXTRACTABLE from true to false is possible, but the other way round is not, because the attribute cannot be changed from false to true.

Or simply generate an EC key pair with both keys having CKA_IBM_PROTKEY_EXTRACTABLE=TRUE.
Then create a signature with the priv key (with CKA_IBM_PROTKEY_EXTRACTABLE=TRUE), then sent CKA_IBM_PROTKEY_EXTRACTABLE =FALSE on the priv key, and generate a second signature (keep both, the first and the second signature).  Then verify the second signature with the pub key (still having CKA_IBM_PROTKEY_EXTRACTABLE=TRUE), then set CKA_IBM_PROTKEY_EXTRACTABLE=FALSE on the pub key, and verify the first signature with it.

jschmidb:
> Or simply generate an EC key pair with both keys having CKA_IBM_PROTKEY_EXTRACTABLE=TRUE ...

Good ideas! I just added a testcase importing the known keys from ECDH_TEST_VECTOR ecdh_tv. With that, I can separately create the private key with CKA_IBM_PROTKEY_EXTRACTABLE=true and the public key with CKA_IBM_PROTKEY_EXTRACTABLE=false, and vice versa. Works both ways. I will push this together with the changes from Juergens comments.

jschmidb:
OK, the recent push contains the fixes for Juergens comments, a new testcase according to Ingos comments, and some casts to get around the Travis errors. Hope I catched them all.

testcases/crypto/ec_func.c
```diff
@@ -1887,6 +1935,191 @@ CK_RV run_TransferECCKeyPairSignVerify()
     return rc;
 }
 
+/**
+ * Tests the EP11 token protected key option. PKEY_MODE ENABLE4NONEXTR must be
```
juergenchrist, line 1939: What is the default value of PKEY_MODE?  Unfortunately there is the option "DEFAULT" which I guess is not the default value if PKEY_MODE is not specified in the config file.  Since the default config file does not specify PKEY_MODE (except for the documentation of it), I am just wondering if we will ever execute this test case in the CI or not?

jschmidb: You are right: the default for PKEY_MODE is in fact DEFAULT, so the protected key option will not be enabled in CI automatically. This is a general problem with all of our token options. Many of the internal tests depend on various token options and will always be executed using defaults.

jschmidb:
Fixed some more Travis errors

jschmidb:
I'm still struggling with Travis errors in some inline asm code.

jschmidb:
Reverted last change. Have to check how to fix the Travis s390x-clang-locks errors.

jschmidb:
The trick with coding the inline asm correctly for clang was leaving the first arg in the parentheses out:
`lrvg    %%r0,0(0,%[__src])   -->   lrvg    %%r0,0(%[__src])`
Thanks Patrick for the hint!

p-steuer:
Looks good to me otherwise.

usr/lib/common/pkey_utils.c
```diff
@@ -31,6 +31,72 @@
...
+    unsigned long long facility_bits[3];
+    int msa = 0;
+    int num = 0;
+#ifdef IMPLEMENT_GETAUXVAL
+    const unsigned long hwcap = getauxval(AT_HWCAP);
+
+    if (hwcap & HWCAP_S390_STFLE) {
+#endif
+        memset(facility_bits, 0, sizeof(facility_bits));
+        num = s390_stfle(facility_bits, 3);
```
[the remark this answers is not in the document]
jschmidb: > so in case IMPLEMENT_GETAUXVAL is undefined s390_stfle is still called ?

Yes, this was my intention. If getauxval is undefined, because we are running on an old kernel, STFLE could probably still be available, because we are running on new hardware. However, yes, if running on old hw or on an emulator not supporting STFLE, this would crash. Do we want to be defensive here?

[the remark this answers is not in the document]
p-steuer: Yes. Id say its okay to not support this feature on systems that dont have getauxval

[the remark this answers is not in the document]
jschmidb: Done.

p-steuer (approved):

PR #380, opened 2021-03-10
Fix AB for pre CEX7 cards

juergenchrist:
CI discovered that ECDH1 derivation does not work with AB keys on pre-CEX7 cards.  Fix and fail with CKR_MECHANISM_INVALID in that case.

### DIFF
--- a/testcases/crypto/abfunc.c
+++ b/testcases/crypto/abfunc.c
@@ -565,24 +565,7 @@ void do_TestInvalidKeys(CK_SESSION_HANDLE session)
     CK_OBJECT_HANDLE pubkey = CK_INVALID_HANDLE, privkey = CK_INVALID_HANDLE;
     CK_RV rc;
 
-    // Expected to fail with CKR_TEMPLATE_INCOMPLETE
-    testcase_begin("Create dilithium AB key");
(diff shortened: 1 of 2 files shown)


testcases/crypto/abfunc.c
```diff
@@ -565,24 +565,6 @@ void do_TestInvalidKeys(CK_SESSION_HANDLE session)
     CK_OBJECT_HANDLE pubkey = CK_INVALID_HANDLE, privkey = CK_INVALID_HANDLE;
     CK_RV rc;
 
-    // Expected to fail with CKR_TEMPLATE_INCOMPLETE
```
ifranzki, line 568: So this removal is not related to the bug, right? 
Do I read it correctly that Dilithium keys are supported as AB keys?

juergenchrist: Ups, this line also makes sense for the remaining test cases.  Putting it back in.  But yes, Dilithium keys are actually supported for AB keys (undocumented feature).  So if Dilithium can be used, this test case will actually fail.

ifranzki (approved):
lgtm

p-steuer (approved):

PR #381, opened 2021-03-15
Fix for slot option. Now accepting 0 as argument

reumann-ibm:
opencryptoki p11sak tool: slot option did not accept argument 0 for slot index 0

The p11sak tool now accepts argument 0 for --slot <SLOTID>

Signed-off by: Matthias Reumann <matthias.reumann1@ibm.com>

### DIFF
--- a/usr/lib/ep11_stdll/ep11_specific.c
+++ b/usr/lib/ep11_stdll/ep11_specific.c
@@ -10705,11 +10705,12 @@ static CK_RV ep11tok_relogin_session(STDLL_TokData_t * tokdata,
     return CKR_OK;
 }
 
-CK_RV ep11tok_logout_session(STDLL_TokData_t * tokdata, SESSION * session)
+CK_RV ep11tok_logout_session(STDLL_TokData_t * tokdata, SESSION * session,
+                             CK_BBOOL in_fork_initializer)
 {
     ep11_private_data_t *ep11_data = tokdata->private_data;
     ep11_session_t *ep11_session = (ep11_session_t *) session->private_data;
-    CK_RV rc, rc2;
+    CK_RV rc = CKR_OK, rc2;
     ST_SESSION_HANDLE handle = {.slotID =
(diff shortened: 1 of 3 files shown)


usr/sbin/p11sak/p11sak.c
```diff
@@ -1878,11 +1868,6 @@ static CK_RV parse_remove_key_args(char *argv[], int argc, p11sak_kt *kt,
         *label = "";
     }
 
-    if (*slot == 0) {
```
ifranzki, line 1881: So now you don't recognize anymore that no slot is specified, but silently use slot 0. This might be OK if this is intended.....

You could alternatively have a flag variable within this function that is set to true when the slot option is parsed, and left false when no slot option is found. Then you could check that flag here and issue the message if no slot was specified. 

The same applies to the other 2 places that you changed.

reumann-ibm: I added a bool to check if the slot argument was set in all three functions that check for arguments.

usr/sbin/p11sak/p11sak.c
```diff
@@ -2133,8 +2154,12 @@ static CK_RV list_ckey(CK_SESSION_HANDLE session, p11sak_kt kt, int long_print)
                 printf(" %*s | ", CELL_SIZE, keytype);
             printf("%s\n", label);
         }
+
         free(label);
+        *label = NULL;
```
ifranzki, line 2159: This must be `label = NULL` anf not `*label = NULL` ? 
Otherwise you set the first char where the just freed pointer points to to zero.....
(use after free).

Same for the other places.

p-steuer:
Hmm i think the bool is added in the wrong commit f the two.

usr/sbin/p11sak/p11sak.c
reumann-ibm: Changed. See new PR #382 and commit 7d2db1e1

PR #382, opened 2021-03-17
Fixes potential double free() to label and keytype pointers. And added option to allow for slotID = 0

reumann-ibm:


usr/sbin/p11sak/p11sak.c
```diff
@@ -1656,6 +1657,12 @@ static CK_RV parse_list_key_args(char *argv[], int argc, p11sak_kt *kt,
         } else if (strcmp(argv[i], "--slot") == 0) {
             if (i + 1 < argc) {
                 *slot = (CK_ULONG) atol(argv[i + 1]);
+                if ((*slot > 0) || (strcmp(argv[i + 1], "0") == 0)) {
```
ifranzki, line 1660: just a minor thing: If the slot is specified as '+0' or '0x0', or ' 0', then you would still treat this as an error. However, this might be acceptable.....

usr/sbin/p11sak/p11sak.c
```diff
@@ -2149,7 +2149,9 @@ static CK_RV list_ckey(CK_SESSION_HANDLE session, p11sak_kt kt, int long_print)
             printf("%s\n", label);
         }
         free(label);
+        label = NULL;
```
ifranzki, line 2173: Commit title: Its more to avoid a double free, not a memory leak. Maybe you can rephrase just the commit title (and shorten the title while you are at it, it overflows into the description).

usr/sbin/p11sak/p11sak.c
reumann-ibm: atol returns 0 if the string cannot be converted to a long int. It returns 0, too, if the argument is "0". Hence, I check for the string to be "0" or for the long to be greater than 0. The user input " 0" sets the slotID = 0. Others I have not tested.

usr/sbin/p11sak/p11sak.c
reumann-ibm: Commit message should be changed now.

usr/sbin/p11sak/p11sak.c
ifranzki: right, atoi also accepts hex format (e.g. 0x0), octal, prefixed with whitespace chars, etc. So it might return 0 also for a valid zero input test, but your strcmp with "0" might not catch that.  You might want to use strtol() instead or atoi(), to better detect parsing errors:
From https://man7.org/linux/man-pages/man3/strtol.3.html
```
           errno = 0;    /* To distinguish success/failure after call */
           val = strtol(str, &endptr, base);

           /* Check for various possible errors */

           if (errno != 0) {
               perror("strtol");
               exit(EXIT_FAILURE);
           }

           if (endptr == str) {
               fprintf(stderr, "No digits were found\n");
               exit(EXIT_FAILURE);
           }

           /* If we got here, strtol() successfully parsed a number */
```

usr/sbin/p11sak/p11sak.c
ifranzki: Yes, but now the PR contains only one commit. The second commit seems to be gone.....

usr/sbin/p11sak/p11sak.c
reumann-ibm: I changed the parsing of the slot number replacing atol() with strtol().

usr/sbin/p11sak/p11sak.c
reumann-ibm: Should be back again.

usr/sbin/p11sak/p11sak.c
```diff
@@ -1655,7 +1659,27 @@ static CK_RV parse_list_key_args(char *argv[], int argc, p11sak_kt *kt,
             /* Get options */
         } else if (strcmp(argv[i], "--slot") == 0) {
             if (i + 1 < argc) {
-                *slot = (CK_ULONG) atol(argv[i + 1]);
+                str = argv[i + 1];
+                errno = 0;
+                *slot = strtol(str, &endptr, base);
```
ifranzki, line 1664: You could move the strtol handling plus error checking into a subroutine, since you have that same code 3 times now....
But if other reviewers are OK with having the same code 3 times, I am OK, too.

ifranzki (approved):

usr/sbin/p11sak/p11sak.c
p-steuer: i dont mind in that case.

p-steuer (approved):

PR #386, opened 2021-03-19
p11sak: fix attributes help

jschmidb:
A few fixes on attribute handling in p11sak:
- The token attr (T) is always true and ignored when specified for keygen
- All attrs are only applicable for symmetric keys, a subset is applicable for asymmetric keys. This is now reflected in the key type specific help
- Msg "Unknown argument '%c'" is replaced by a more "speaking" msg


reumann-ibm:
I reviewed the code changes and approve.

p-steuer (approved):

ifranzki (approved):

PR #388, opened 2021-03-23
Fix detection of DER encoded or raw ECDH public data

ifranzki:
From PKCS#11: "pPublicData: pointer to other party’s EC public key value. A token MUST be able to accept this value encoded as a raw octet string (as per section A.5.2 of [ANSI X9.62]). A token MAY, in addition, support accepting this value as a DER-encoded ECPoint (as per section E.6 of [ANSI X9.62]) i.e. the same as a CKA_EC_POINT encoding."

It can happen that the specified  public data looks like it would be an DER encoded ECPoint, but in fact it is not. This happens when the public data starts with 0x04 (either an OCTET STRING DER tag, or a point conversion indicator byte for "uncompressed") followed by a byte who's value is 2 less than the length of the public data. That way the second byte looks like a valid single-byte length octet of a DER encoded tag.

Add an additional check if the resulting length after DER decoding is less than expected (2 times prime length of the curve plus one format conversion byte). If so, then check if the raw length would fit better, and use it as raw if the raw length is exactly the expected length.

Note that OCK also supports ECPoints without a format conversion byte, so we can not simply always check for the expected length.

### DIFF
--- a/usr/lib/ep11_stdll/ep11_specific.c
+++ b/usr/lib/ep11_stdll/ep11_specific.c
@@ -56,6 +56,7 @@
 
 #include <ica_api.h>
 #include <openssl/crypto.h>
+#include <openssl/ec.h>
 
 #include "ep11_func.h"
 #include "ep11_specific.h"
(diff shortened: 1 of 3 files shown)


p-steuer (approved):
Wish we had a better way to handle this but i cant think of any .

PR #389, opened 2021-03-25
Fix detection of DER encoded or raw ECDH public data (2)

ifranzki:
The previous commit did not completely fix the detection of the ECDH public data format.

When a raw ECPoint starts with 0x04 followed by a byte who's value is 2 less than the length of the public data, followed by a byte that is a valid point conversion indicator (e.g. 0x02 - compressed, 0x04 - uncompressed, or 0x06 - hybrid) then it would have still been detected as an DER encoded EC Point, resulting in a wrong EC Point.

The detection logic is corrected and is now moved into a separate function, and is used at all 3 places, i.e. in th soft token, the ICA token, and the EP11 token.

### DIFF
--- a/testcases/misc_tests/obj_lock.c
+++ b/testcases/misc_tests/obj_lock.c
@@ -336,7 +336,7 @@ int main(int argc, char **argv)
          * before using this test option.
          */
         if (!is_ep11_token(SLOT_ID)) {
-            testcase_notice("Slot %u doesn't support protected keys.", SLOT_ID);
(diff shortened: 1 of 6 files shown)


p-steuer (approved):
Looks good to me. I will merge it once the build errors are fixed.

ifranzki:
This was actually a compile error in a previous commit/PR that we did oversee (see https://travis-ci.org/github/opencryptoki/opencryptoki/builds/763927185).
Nevertheless, I have fixed it with an additional commit.

PR #392, opened 2021-03-29
Minor ICSF Fixes

ifranzki:
Some test cases with specific EC curves and RSA keys still fail, but I fear that this is due to ICSF not supporting those EC curves and RSA keys.

### DIFF
--- a/usr/lib/icsf_stdll/icsf.c
+++ b/usr/lib/icsf_stdll/icsf.c
@@ -2976,6 +2976,9 @@ int icsf_hash_signverify(LDAP * ld, int *reason, struct icsf_object_record *key,
         return -1;
     }
 
+    if (sig == NULL && sig_len != NULL)
+        *sig_len = 0;
+
     if (ber_printf(msg, "ooo", (clear_text) ? clear_text : "",
                    clear_text_len, (chain_data) ? chain_data : "",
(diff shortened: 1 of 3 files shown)


usr/lib/icsf_stdll/icsf.c
```diff
@@ -2976,6 +2976,9 @@ int icsf_hash_signverify(LDAP * ld, int *reason, struct icsf_object_record *key,
         return -1;
     }
 
+    if (sig == NULL && sig_len != NULL)
+        *sig_len = 0;
```
p-steuer, line 2981: shouldnt we return here directly in case of size query ?

ifranzki: No, this would not be a real size query then, as it would always return zero as size, but that's not what is desired. 

The code further down (i.e. icsf_call) will do the real size query, if the (ber encoded) signature length is zero (or smaller than required for the signature) in the request. The result returned from icsf_call will then contain the required size, and *sig_len is set from that. 

Setting *sig_len to zero is required, because *sig_len could contain any value on entry. So by setting *sig_len to zero first (if sig is NULL) we ensure that a size query is done, regardless what *sig_len was on entry.

p-steuer (approved):

PR #399, opened 2021-04-16
Fix heap-out-of-bounds access

ifranzki:


jschmidb:
Looks reasonable. To make it even more clear, we could do it like:
```
CK_ULONG ec_params_len;
...
ec_params = (CK_BYTE *) attr->pValue;
ec_params_len = attr->ulValueLen;
...
if ((ec_params_len == sizeof(p256) &&
         memcmp(ec_params, p256, sizeof(p256)) == 0) ||
...
```

But the additional variable is not really necessary.

p-steuer (approved):

PR #400, opened 2021-04-16
Fix potential deadlock situation with double read-locks

ifranzki:
Do not get and read-lock an object twice within the same thread via function object_mgr_find_in_map1(), as this would read-lock the object twice.

This could cause a deadlock situation, when in-between the first and the second call to object_mgr_find_in_map1() the token object is modified by another process. The second object_mgr_find_in_map1() would detect that the object has been modified (object_mgr_check_shm()), and would try to re-load the object from the disk. For re-loading, the object is unlocked once, and a write-lock is acquired instead. However, if the current thread has read-locked the object twice, but releases only one read-lock, then it will never get the write lock, because it still owns the read lock itself.

To avoid this situation, release the read-lock before calling another function that also acquires the read lock of the object. That way, only one read-lock is held by the current thread, and re-loading the object will not cause a deadlock.

### DIFF
--- a/usr/lib/common/decr_mgr.c
+++ b/usr/lib/common/decr_mgr.c
@@ -540,6 +540,10 @@ CK_RV decr_mgr_init(STDLL_TokData_t *tokdata,
         }
         memset(ctx->context, 0x0, sizeof(AES_GCM_CONTEXT));
 
+        /* Release obj lock, token specific aes-gcm may re-acquire the lock */
+        object_put(tokdata, key_obj, TRUE);
+        key_obj = NULL;
+
         rc = aes_gcm_init(tokdata, sess, ctx, mech, key_handle, 0);
         if (rc) {
             TRACE_ERROR("Could not initialize AES_GCM parms.\n");
--- a/usr/lib/common/encr_mgr.c
+++ b/usr/lib/common/encr_mgr.c
@@ -537,6 +537,10 @@ CK_RV encr_mgr_init(STDLL_TokData_t *tokdata,
         }
         memset(ctx->context, 0x0, sizeof(AES_GCM_CONTEXT));
 
+        /* Release obj lock, token specific aes-gcm may re-acquire the lock */
+        object_put(tokdata, key_obj, TRUE);
+        key_obj = NULL;
+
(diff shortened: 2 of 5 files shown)


jschmidb:
Yes, indeed. For AES-GCM, only the icatoken is affected. Here, token_specific_aes_gcm_init in ica_specific.c does a object_mgr_find_in_map1(tokdata, key, &key_obj, READ_LOCK), so that the object is read-locked twice. For RSA-OAEP, ccatoken, icatoken, and softtoken are affected. For PSS-SIGN, cca, ep11, ica, and soft tokens are affected. 
Hm, the question is, are these all places? I searched a bit for object_mgr_find_in_map1 and found this path also:
ep11tok_sign -> obtains a READ_LOCK on the obj via h_opaque_2_blob -> then calls sign_mgr_sign -> calls rsa_pkcs_sign, which again obtains a READ_LOCK via object_mgr_find_in_map1.

jschmidb:
Another path would be:
ep11tok_encrypt -> obtains READ_LOCK via h_opaque_2_blob -> then calls encr_mgr_encrypt -> calls rsa_x509_encrypt, which again obtains a READ_LOCK via object_mgr_find_in_map

ifranzki:
Hm, yea good finding. I did a search, too, but obviously looked too short.

Theoretically we should unlock the object at all places, where a token specific function is called and only the object handle is passed (but not the object pointer). Or even better would be to call those token specific functions always with the object pointer (instead of the handle), and keep the locking strictly in common code. But that would be trillions of places to change....

ifranzki:
I think I should now have covered all the places in ep11_specific.c. In most cases we can simply change the order to first check if pkey is active and only if  it is not active, then obtain the lock.

p-steuer:
> Theoretically we should unlock the object at all places, where a token specific function is called and only the object handle is passed (but not the object pointer). Or even better would be to call those token specific functions always with the object pointer (instead of the handle), and keep the locking strictly in common code. But that would be trillions of places to change....

Yes, i dont see why the token specific stuff would need to lock objects. Approved anyway, as a fix for the deadlock problem.

p-steuer (approved):

PR #402, opened 2021-05-05
SOFT: Check the EC Key on C_CreateObject and C_DeriveKey

ifranzki:
When constructing an OpenSSL EC public or private key from PKCS#11 attributes or ECDH public data, check that the key is valid, i.e. that the point is on the curve.

This prevents one from creating an EC key object via C_CreateObject with invalid key data. It also prevents C_DeriveKey to derive a secret using ECDH with an EC public key (public data) that uses a different curve
or is invalid by other means.

### DIFF
--- a/usr/lib/soft_stdll/soft_specific.c
+++ b/usr/lib/soft_stdll/soft_specific.c
@@ -4365,6 +4365,12 @@ static CK_RV fill_ec_key_from_pubkey(EC_KEY *ec_key, const CK_BYTE *data,
         goto out;
     }
 
+    if (!EC_KEY_check_key(ec_key)) {
+        TRACE_ERROR("EC_KEY_check_key failed\n");
+        rc = CKR_PUBLIC_KEY_INVALID;
+        goto out;
+    }
+
 out:
     if (allocated && ecpoint != NULL)
         free(ecpoint);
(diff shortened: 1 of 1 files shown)


juergenchrist (approved):
lgtm

usr/lib/soft_stdll/soft_specific.c
```diff
@@ -4365,6 +4365,19 @@ static CK_RV fill_ec_key_from_pubkey(EC_KEY *ec_key, const CK_BYTE *data,
         goto out;
     }
 
+    if (EC_POINT_is_on_curve(EC_KEY_get0_group(ec_key),
+                             EC_KEY_get0_public_key(ec_key), NULL) != 1) {
```
juergenchrist, line 4369: This check is also done by `EC_KEY_check_type` as far as I can see in the code.  However, since this is not documented (or I just could not find it), I would keep this check here and below.

ifranzki: Right, that was also my thinking when I coded that. `EC_KEY_check_key` should actually also check if the public key point is on the curve, but the documentation does not state that explicitly. So I also thought that its better to leave it in, even if the same thing is checked twice.

ifranzki:
According to @p-steuer `EC_KEY_check_key()` is sufficient, and we should not do an additional `EC_POINT_is_on_curve()` check, since there is a cost associated with doing checking this twice.

I have changed the code accordingly.

p-steuer (approved):

PR #403, opened 2021-05-07
Change PIDFile to use a non-deprecated directory

markkp:
The usage of /var/run is deprecated. Use /run for the PIDFile instead.

### DIFF
--- a/misc/pkcsslotd.service.in
+++ b/misc/pkcsslotd.service.in
@@ -4,7 +4,7 @@ After=local-fs.target
 
 [Service]
 Type=forking
-PIDFile=/var/run/pkcsslotd.pid
+PIDFile=/run/pkcsslotd.pid
 ExecStart=@sbindir@/pkcsslotd
 
 [Install]


juergenchrist (changes requested):
This change only changes the place the pid-file is used in the systemd service file.  It, however, introduces an inconsistency in that we use `/run/pkcsslotd.pid` in systemd but create `/var/run/pkcsslotd.pid` in the slot daemon.  Even though this should not be a problem due to symlinking we should keep at least these two places consistent.

ifranzki:
Agree. And if we change it, we should probably also change the defines for the sockets in slotmgr.h like:
```
#define PROC_SOCKET_FILE_PATH "/var/run/pkcsslotd.socket"
#define ADMIN_SOCKET_FILE_PATH "/var/run/pkcsslotd.admin.socket"
```
to
```
#define PROC_SOCKET_FILE_PATH "/run/pkcsslotd.socket"
#define ADMIN_SOCKET_FILE_PATH "/run/pkcsslotd.admin.socket"
```

ifranzki:
I have opened https://github.com/opencryptoki/opencryptoki/pull/508 to change this for all places.
I am closing this one for now.
Thanks for suggesting the change!

Issue #404, opened 2021-05-17
failures in ec_tests from upstream testsuite


ngothan:
Description of problem:

5 failures are reported when running ec_tests from upstream test suite. They are present when running tests for ICA token on s390x.

Token #1 Info:
	Label: icatok                          
	Manufacturer: IBM                             
	Model: ICA             
	Serial Number:                 
	Flags: 0x880045 (RNG|LOGIN_REQUIRED|CLOCK_ON_TOKEN|USER_PIN_TO_BE_CHANGED|SO_PIN_TO_BE_CHANGED|)
	Sessions: 0/[effectively infinite]
	R/W Sessions: [information unavailable]/[effectively infinite]
	PIN Length: 4-8
	Public Memory: [information unavailable]/[information unavailable]
	Private Memory: [information unavailable]/[information unavailable]
	Hardware Version: 0.0
	Firmware Version: 0.0
	Time: 2021042902021900
```
./ec_tests  -slot 1

------
* TESTCASE run_GenerateECCKeyPairSignVerify BEGIN Starting ECC generate key pair with pkey=0 ...
* TESTCASE run_GenerateECCKeyPairSignVerify FAIL (testcases/crypto/ec_func.c:1395) C_GenerateKeyPair with valid input failed at i=0 (brainpoolP160r1), rc=CKR_FUNCTION_FAILED
------
* TESTCASE run_GenerateSignVerifyECC BEGIN Starting with mechtype='CKM_ECDSA', inputlen=20 parts=0, pkey=0
* TESTCASE run_GenerateSignVerifyECC ERROR (testcases/crypto/ec_func.c:1206)) C_Sign rc=CKR_FUNCTION_FAILED
* TESTCASE run_ImportECCKeyPairSignVerify FAIL (testcases/crypto/ec_func.c:1577) run_GenerateSignVerifyECC failed index=0.
------
* TESTCASE run_GenerateSignVerifyECC BEGIN Starting with mechtype='CKM_ECDSA', inputlen=20 parts=0, pkey=0
* TESTCASE run_GenerateSignVerifyECC ERROR (testcases/crypto/ec_func.c:1206)) C_Sign rc=CKR_FUNCTION_FAILED
* TESTCASE run_TransferECCKeyPairSignVerify FAIL (testcases/crypto/ec_func.c:1889) run_GenerateSignVerifyECC failed index=0.
------
[...]

PR #405, opened 2021-05-17
ICA: handle unsupported curves correctly for openssl fallbacks

jschmidb:
This pr fixes issue #404. 
When libica has no crypto cards, the set of supported EC curves depends on openssl.
If openssl is e.g. in FIPS mode, very few curves are supported. This was not handled
correctly in icatoken and the EC tests. 
As libica itself does not provide sign update, I added the check for unsupported curves
only for sign and sign final to keep changes small. We may discuss this.

### DIFF
--- a/testcases/crypto/ec_func.c
+++ b/testcases/crypto/ec_func.c
@@ -911,7 +911,11 @@ CK_RV run_DeriveECDHKeyKAT()
                                 priv_keyA, derive_tmpl,
                                 derive_tmpl_len, &secret_keyA);
         if (rc != CKR_OK) {
-            if (is_ep11_token(SLOT_ID) &&
+            if (rc == CKR_CURVE_NOT_SUPPORTED) {
+                testcase_skip("Slot %u doesn't support this curve: %s",
+                              (unsigned int) SLOT_ID, ecdh_tv[i].name);
+                goto testcase_next;
+            } else if (is_ep11_token(SLOT_ID) &&
                 rc == CKR_MECHANISM_PARAM_INVALID &&
                 (ecdh_tv[i].kdf != CKD_NULL ||
                  ecdh_tv[i].shared_data_len > 0)) {
--- a/usr/lib/ica_s390_stdll/ica_specific.c
+++ b/usr/lib/ica_s390_stdll/ica_specific.c
@@ -4261,8 +4261,13 @@ CK_RV token_specific_ec_generate_keypair(STDLL_TokData_t *tokdata,
     /* Generate key data for this key object */
     rc = p_ica_ec_key_generate(ica_data->adapter_handle, eckey);
     if (rc != 0) {
-        TRACE_ERROR("ica_ec_key_generate() failed with rc=%d.\n", rc);
-        ret = CKR_FUNCTION_FAILED;
+        if (rc == EINVAL) {
+            TRACE_ERROR("ica_ec_key_generate() failed with rc=EINVAL, probably curve not supported by openssl.\n");
+            ret = CKR_CURVE_NOT_SUPPORTED;
+        } else {
+            TRACE_ERROR("ica_ec_key_generate() failed with rc=%d.\n", rc);
+            ret = CKR_FUNCTION_FAILED;
+        }
         goto end;
     }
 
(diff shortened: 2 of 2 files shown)


usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -4261,8 +4261,13 @@ CK_RV token_specific_ec_generate_keypair(STDLL_TokData_t *tokdata,
     /* Generate key data for this key object */
     rc = p_ica_ec_key_generate(ica_data->adapter_handle, eckey);
     if (rc != 0) {
-        TRACE_ERROR("ica_ec_key_generate() failed with rc=%d.\n", rc);
-        ret = CKR_FUNCTION_FAILED;
+        if (rc == EINVAL) {
```
ifranzki, line 4264: I guess libica may return EINVAL also for other reasons? With this we treat every EINVAL return as CKR_CURVE_NOT_SUPPORTED. However, without changing libica to return a more precise error indication in case of an unsupported curve, I don't see a better option to handle this.

testcases/crypto/ec_func.c
```diff
@@ -1196,15 +1200,29 @@ CK_RV run_GenerateSignVerifyECC(CK_SESSION_HANDLE session,
     if (parts > 0) {
         rc = funcs->C_SignFinal(session, signature, &signaturelen);
         if (rc != CKR_OK) {
-            testcase_error("C_SignFinal rc=%s", p11_get_ckr(rc));
-            goto testcase_cleanup;
+            if (rc == CKR_CURVE_NOT_SUPPORTED) {
```
ifranzki, line 1203: I guess there is no way to detect the invalid curve already a C_SignInit? That would probably be the better place, but if the ICA token can't find out if it supports the curve at C_SignInit, then this place is OK, too.

usr/lib/ica_s390_stdll/ica_specific.c
jschmidb: Yes, right. We would have to introduce a specific error code for 'curve not supported by openssl'. In case there is no card, libica returns ENODEV, but only if fallbacks are disabled. With fallbacks enabled, we get either 0 (=ok), EINVAL (=curve not supported by openssl), EACCES (fips error), or EIO (any other error). 
We could introduce EPERM, like 'operation not permitted by openssl' for unsupported curves, but would break the libica API with that. Currently, the libica book is not very specific here, but existing customer applications might check this rc.

testcases/crypto/ec_func.c
jschmidb: Yep, there is no way. But as an aside: I saw that in ica_ec_key_new(), libica checks for supported curves via curve_supported(nid), which is just a switch over some known curves to be supported. There is another routine used in ecdsa_sign_sw: is_supported_openssl_curve(), which really tries to use a given curve in openssl. Maybe we can make a better check in ica_ec_key_new / ica_ec_key_init / ica_ec_key_generate so that an unsupported curve shows up when creating/initializing the key. But this would be a separate libica item.

usr/lib/ica_s390_stdll/ica_specific.c
juergenchrist: I guess this is a problem we will tackle once moving all sw-fallbacks from libica to the ica-token?

usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -4261,8 +4261,13 @@ CK_RV token_specific_ec_generate_keypair(STDLL_TokData_t *tokdata,
...
+            TRACE_ERROR("ica_ec_key_generate() failed with rc=%d.\n", rc);
```
juergenchrist, line 4265: Maybe we could add `EINVAL` to the error message here?  Makes it quicker to find what is going wrong...  Similarly at other places.

juergenchrist (approved):

usr/lib/ica_s390_stdll/ica_specific.c
jschmidb: Yes, exactly.

usr/lib/ica_s390_stdll/ica_specific.c
jschmidb: Yes, makes sense. Done.

usr/lib/ica_s390_stdll/ica_specific.c
p-steuer: @ifranzki :
> I guess libica may return EINVAL also for other reasons? 

@jschmidb 
> Yes, right. [...] EINVAL (=curve not supported by openssl) [...]

Sounds to me like EINVAL = curve not supported by openssl.. in what other cases EINVAL is returned ?

Anyway, if you need to change the return codes, please update the book and the comments in the code.

p-steuer:
I dont agree with the error code mapping in case of sign and verify:

> If an attempt to create, generate, derive or unwrap an EC key of an unsupported curve is made, the attempt should fail with the error code CKR_CURVE_NOT_SUPPORTED.

[pkcs11-curr-v3.0]

Also see [pkcs11-base-v3.0] which lists the curve not sup error with C_CreateObject, C_GenerateKey(Pair),  C_DeriveKey,  and UnwrapKey,

usr/lib/ica_s390_stdll/ica_specific.c
jschmidb: > in what other cases EINVAL is returned ?

The usual parm checking, e.g. if any parm is null

jschmidb:
> I dont agree with the error code mapping in case of sign and verify:
> If an attempt to create, generate, derive or unwrap an EC key of an unsupported curve is made, the    attempt should fail with the error code CKR_CURVE_NOT_SUPPORTED.

Yes, agreed. I removed the changes for sign/verify. With that, the ec_tests will still show 2 failed tests in a scenario without a CCA card and openssl-fips. Another fix is necessary in libica to check for supported curves when importing a key.

p-steuer (approved):

PR #409, opened 2021-06-08
ICA: Fix libica return code handling for unsupported curves

jschmidb:
With commit "Fix checking if curve supported in key generate, init, derive",
libica returns EPERM, if a key is generated, initialized, or derived, and
the EC curve is not supported.


p-steuer:
So with older libica versions that do not return EPERM in that case, it would fail with the generic "ica_ecdh_derive_secret() failed with rc =..." ? Im ok with that.

p-steuer (approved):

ifranzki (approved):

PR #410, opened 2021-06-10
Improve CI coverage

juergenchrist:
Change our CI code such that we test more paths.  For every token we now test
multi tenant support by instantiating at least two tokens.  New token naming
scheme for the CI is:

Token numbers: <typenum><count>
where <typename> is
ICA:  1
CCA:  2
SW:   3
EP11: 4

For all except EP11 we have as tokname the lowercase type followed by a
0-based count (similar to token numbers).  For EP11 the token is called "ep"
followed by the 0-based count.  Furthermore, for EP11 the token configuration
file differs between the tokens:

0: Default with only APQN_ANY
1: FORCE_SENSITIVE and APQN_ANY
2: STRICT_MODE, VHSM_MODE, and APQN_ANY
3: OPTIMIZE_SINGLE_OPERATIONS and APQN_ANY
4: DIGEST_LIBICA OFF and APQN_ANY
5: PKEY_MODE DISABLED and APQN_ANY
6: AQPN_WHITELIST selecting the latest CEXP as defined in testcaes/ciconfig.sh
variable LATESTCEXP

Note the count 6 for EP11 is only generated if a corresponding card is found
on the machine.  Otherwise this slot will be skipped.

Preparation for the CI is built into the makefile.  The current implementation
does not allow for parallel make since then preparation and the installcheck
would run in parallel.  But this is not a problem since the installcheck
itself is already highly parallelized (one process per slot to test running in
parallel).

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -201,6 +201,14 @@ endif
 if ENABLE_TESTCASES
 PKCS11_SO_PIN ?= 76543210
 PKCS11_USER_PIN ?= 01234567
+PKCS11_VHSM_PIN ?= 1234567890
+
+ci-prepare:
+	killall -HUP pkcsslotd || true
+	${srcdir}/testcases/ciconfig.sh "$(sysconfdir)/opencryptoki" "$(sysconfdir)/opencryptoki"
+	@sbindir@/pkcsslotd
+	cd ${srcdir}/testcases && ./init_token.sh 42 && ./init_vhsm.exp 42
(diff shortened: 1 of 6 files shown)


juergenchrist:
Runtime (at least locally) seemed okay.  Yes, it increased mostly due to the latest CEXP slot which runs many tests that the other slots do not run on my machine.  I would anyway like to give it a try and adjust if it takes too long.

ifranzki (approved):
lgtm

p-steuer (approved):
thanks !

PR #411, opened 2021-06-15
[TEST] Fix setup and cleanup of slots

juergenchrist:
With the addition of lots of slots to our CI the initialization of slots ran
into timeouts in the expect scripts.  Unfortunately these errors were ignored
(expect standard action) which caused uninitialized slots to be tested.
Remove the 5 second timeout in preference of the default 10 seconds and treat
timeout errors in the expect script.

Also provide a cleanup target for the CI that correctly removes all token data
such that the next run starts from a clean state.

Do not attempt to reinitialize slot 42 if it is already initialized.

Switch slot 45 to use PKEYs since the default and disabled setting currently
cover the same code paths.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -207,7 +207,8 @@ ci-prepare:
 	killall -HUP pkcsslotd || true
 	${srcdir}/testcases/ciconfig.sh "$(sysconfdir)/opencryptoki" "$(sysconfdir)/opencryptoki"
 	@sbindir@/pkcsslotd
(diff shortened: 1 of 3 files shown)


juergenchrist:
This PR requires an additional change to our CI that can only be done once this is merged (usage of `make ci-uninstall` instead of `make uninstall`).  I will do this once this PR is merged.

p-steuer (approved):

PR #412, opened 2021-06-21
Serialize token initialization

juergenchrist:
Initializing 12 tokens in parallel causes some random bugs in our CI.  To
prevent these, initialize the tokens in sequence such that in the log output
the errors and commands can be identified without interleaving of other
commands and stray newline characters.


ifranzki (approved):
lgtm

p-steuer (approved):

PR #413, opened 2021-06-22
A couple of minor p11sak fixes

ifranzki:
Things that catched my eyes while looking at the code for a bugzilla....


usr/sbin/p11sak/p11sak.c
```diff
@@ -2566,11 +2567,17 @@ int main(int argc, char *argv[])
     if (!pin) {
         printf("Please enter user PIN:");
         rc = get_pin(&pin, &pinlen);
+        pin_alloced = ckb_true;
```
jschmidb, line 2594: Is 'alloced' an English word?  I would change the variable to e.g. pin_allocated. The code looks good!

usr/sbin/p11sak/p11sak.c
```diff
@@ -2566,11 +2567,17 @@ int main(int argc, char *argv[])
...
 
-        if (strlen(pin) == 0) {
+        if (pin == NULL || strlen(pin) == 0) {
             char *s = getenv("PKCS11_USER_PIN");
             if (s) {
-                strcpy((char*) pin, s);
+                free(pin);
+                pin = strdup(s);
+                if (!pin) {
+                    rc = -ENOMEM;
```
jschmidb, line 2602: Use CKR_HOST_MEMORY ?

usr/lib/common/trace.c
```diff
@@ -131,7 +131,7 @@ void set_trace(struct trace_handle_t t_handle)
 
 void trace_finalize(void)
 {
-    if (trace.fd)
+    if (trace.fd >= 0)
```
juergenchrist, line 134: It is cleaner this way, but since we anyway ignore the error from `close`, we could even skip the `if` statement altogether.

usr/sbin/p11sak/p11sak.c
```diff
@@ -2264,9 +2264,10 @@ static CK_BBOOL user_input_ok(char *input)
 static CK_RV confirm_destroy(char **user_input, char* label)
 {
     int nread;
-    size_t buflen;
+    size_t buflen = 0;
     CK_RV rc = CKR_OK;
 
+    *user_input = NULL;
     printf("Are you sure you want to destroy object %s [y/n]? ", label);
     while (1){
```
jschmidb, line 2294: Would be a cosmetic change also: while (1) {   with a space before the {

usr/sbin/p11sak/p11sak.c
```diff
@@ -103,23 +103,14 @@ static CK_RV get_pin(char **pin, size_t *pinlen)
...
-     * Replace with terminating NULL.
-     */
-    *pin = (char*) malloc(nread);
-    if (*pin == NULL) {
-        rc = -ENOMEM;
-        goto done;
-    }
-
     /* strip the carriage return since not part of pin. */
     user_input[nread - 1] = '\0';
```
juergenchrist, line 107: Could we make this conditional while your are at it?  `getline` also returns on EOF-terminated lines.  Only if the line is empty and EOF is received you get `-1`.

usr/sbin/p11sak/p11sak.c
```diff
@@ -2561,13 +2589,19 @@ int main(int argc, char *argv[])
 
     /* Prompt for PIN if not already set via option */
     if (!pin) {
-        printf("Please enter user PIN:");
+        printf("Please enter user PIN: ");
```
juergenchrist, line 2592: Since `get_pin` requires an `expect` script (due to two-three missing `isatty` calls), this might break some scripts.  Are we sure we want to do that?

usr/sbin/p11sak/p11sak.c
```diff
@@ -2561,13 +2589,19 @@ int main(int argc, char *argv[])
...
         rc = get_pin(&pin, &pinlen);
+        pin_alloced = ckb_true;
```
juergenchrist, line 2594: That seems to be wrong.  Only if `rc != -1` did we actually allocate the pin and have to free it.  But seeing that we initialize `pin` with `NULL`, the error case might not be that interesting.

juergenchrist (approved):

usr/lib/common/trace.c
ifranzki: Well, valgrind issued a warning when close was called with -1, so thats why I wanted to avoid doing that.

usr/sbin/p11sak/p11sak.c
ifranzki: Yea, get_pin does not return CKR_ errors but -1 or negative errnos, so that lead me to use negative errnos. But its certainly better to use CKR_ errors here. Will fix this.

usr/sbin/p11sak/p11sak.c
ifranzki: fixed.

usr/sbin/p11sak/p11sak.c
ifranzki: Hm, good point!  Our tetscases for p11sak don't use expect, but pass the pin as argument via --pin. 
But still there might be other scripts that would be affected.

Opinions? 

Its just 'unusual' to have no space after the colon.....

usr/sbin/p11sak/p11sak.c
ifranzki: Will do.

usr/sbin/p11sak/p11sak.c
ifranzki: I have translate the rc to CKR_ errors anyway, so I set pin_alloced = ckb_true only for rc = 0.

usr/sbin/p11sak/p11sak.c
juergenchrist: I completely agree with you.  That is why I asked the question.  I think the change should be okay since most `expect` users probably anyway only use something like `"Please enter user PIN:"` without anchoring and then it would still match.

ifranzki: So do we want to go with my change? 

If so, then I think I have addressed all review findings so far.

juergenchrist: I think we want this change.  The `get_pin` function is anyway something we have in multiple variants and should one day consolidate.

p-steuer (approved):

PR #414, opened 2021-06-28
configure: Add check for C++ compiler

ifranzki:
AC_PROG_CXX will return "g++" even if no c++ compiler is installed, but does not fail. Add a check to detect if the C++ compiler is missing and fail the configure script if so.


juergenchrist (approved):
lgtm

p-steuer (approved):
seems to be a known issue

PR #419, opened 2021-07-08
pkcstok_migrate: Don't remove 'tokversion = x.y' during migration

ifranzki:
When migrating a slot the opencryptoki.conf file is modified. If it contains slots that already contain the 'tokversion = x.y' keyword, this is accidentally removed when migrating another slot.

### DIFF
--- a/usr/sbin/pkcstok_migrate/pkcstok_migrate.c
+++ b/usr/sbin/pkcstok_migrate/pkcstok_migrate.c
@@ -2060,7 +2060,7 @@ done:
  */
 static int parseupdate_ockversion(void *private, const char *version)
 {
-	struct parseupdate *u = (struct parseupdate *)private;
+    struct parseupdate *u = (struct parseupdate *)private;
 
     fprintf(u->f, "version %s", version);
     return 0;
@@ -2075,14 +2075,14 @@ static void parseupdate_disab_event_supp(void *private)
 
 static void parseupdate_eol(void *private)
 {
-	struct parseupdate *u = (struct parseupdate *)private;
+    struct parseupdate *u = (struct parseupdate *)private;
 
     fputc('\n', u->f);
 }
 
 static int parseupdate_begin_slot(void *private, int slot, int nl_before_begin)
 {
-	struct parseupdate *u = (struct parseupdate *)private;
+    struct parseupdate *u = (struct parseupdate *)private;
 
     u->activeslot = (slot == u->slotnum);
     if (nl_before_begin)
@@ -2094,7 +2094,7 @@ static int parseupdate_begin_slot(void *private, int slot, int nl_before_begin)
 
 static int parseupdate_end_slot(void *private)
 {
-	struct parseupdate *u = (struct parseupdate *)private;
+    struct parseupdate *u = (struct parseupdate *)private;
 
     if (u->activeslot)
         fprintf(u->f, "  tokversion = 3.12\n");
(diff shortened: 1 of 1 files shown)


juergenchrist (approved):

ifranzki:
Added one more commit to fix issue https://github.com/opencryptoki/opencryptoki/issues/420.
Please re-review.

usr/sbin/pkcstok_migrate/pkcstok_migrate.c
```diff
@@ -2491,27 +2491,19 @@ static CK_BBOOL pkcsslotd_running(void)
         /* if endptr is not a null character, the directory is not
          * entirely numeric, so ignore it */
         long lpid = strtol(ent->d_name, &endptr, 10);
-        if (*endptr != '\0') {
+        if (*endptr != '\0')
             continue;
-        }
```
juergenchrist, line 2496: Not sure, but this might lead to merge conflicts for backports...

usr/sbin/pkcstok_migrate/pkcstok_migrate.c
```diff
@@ -2491,27 +2491,19 @@ static CK_BBOOL pkcsslotd_running(void)
...
 
-        /* try to open the cmdline file */
+        /* try to open the cmdline file. Ignore it if open fails */
         snprintf(fname, sizeof(fname), "/proc/%ld/cmdline", lpid);
         fp = fopen(fname, "r");
-        if (!fp) {
-            warnx("fopen(%s) failed, errno=%s", fname, strerror(errno));
-            return CK_TRUE;
-        }
+        if (!fp)
```
juergenchrist, line 2500: I guess this is mostly the case with SELINUX or other LSMs active.  But in general, couldn't we simpllify this whole function a lot?  I mean, we do have a PID file for a running pkcsslotd.  So we could just use that file:
1. If it does not exist => pkcsslotd is not running (yes, a user might have deleted it while pkcsslotd is still running, but then we do not care...)
2. If the PID mentioned in the file does not exist => pkcsslotd is not running (PID file is stale)
3. If /proc/<PID>/cmdline cannot be opened => Assume pkcsslotd is running
4. If /proc/<PID>/cmdline contains the pkcsslotd command => pkcsslotd is running
5. else pkcsslotd is not running.

Sounds easier than iterating through the whole /proc directory skipping all the files in there that are not directories for running processes.

juergenchrist (approved):

usr/sbin/pkcstok_migrate/pkcstok_migrate.c
ifranzki: Changed the code to use the pid file. Please take a took again.

juergenchrist (approved):
lgtm

juergenchrist:
Unrelated, but I thought again about the changes in `parseupdate_key_str` and the TODO of adding a proper string mode into the lexer and parser.  The logic when to add quotation marks is still flawed.  For example the line
```
description = "slot"
```
would be transformed into
```
description = slot
```
since it does not contain a space.  However, the transformed line would not parse.  Maybe we should just go the really save way and quote every string instead of using bare words.  The parser currently (and in the future) does not distinguish between them anyway.

ifranzki:
> However, the transformed line would not parse

Why not ? Both cases, with quotes and without quotes are treated as STRING if I read that correctly.

juergenchrist:
> Both cases, with quotes and without quotes are treated as STRING if I read that correctly.

No.  Since `slot` is a reserved keyword, it is not recognized as STRING, but as SLOT.  The same is true for the other reserved keywords `version` and `disable-event-support` or something like `{idecidedtoputJSONherewithoutanyspaces}` which would fail on `{`.  That was/is the reason I think we need a proper string lexing/parsing mode.  But, as recently already stated, moving to a more common format like, e.g., JSON would simplify a lot.  This whole mess we have here would be easy with json: Read the object into memory, manipulate the object in memory, dump the in-memory object to a tmp file, exchange files.  No need for a custom parser with an ad-hoc grammar with lots of pitfalls...

ifranzki:
> No. Since `slot` is a reserved keyword, it is not recognized as STRING, but as SLOT

OK, yes that's true. 

And I agree using JSON would be beneficial.

The only downside with read-in->modify->write-out is that if you have user comments within the file, they might be lost afterwards (dependent if/how JSON supports comments at all, XML for example supports comments that would be retained.)

But yes, thats unrelated to this PR.

ifranzki:
Added another commit to rework the string quoting in opencryptoki.conf migration processing.

p-steuer (approved):

PR #421, opened 2021-07-13
EP11: Update EP11 specific attributes from key blob

ifranzki:
After a key generation, key import, derive, or unwrap operation, retrieve some EP11 specific attributes from the key blob, and update/set these attributes in the Opencryptoki object as well.

The EP11 specific attributes can be specified by an application in the template for object creation, import, derive, or unwrap, and they will then be contained in the EP11 key blob as well. With this change, EP11 specific attributes that have not been specified and got a default value by EP11, are now also available
as PKCS#11 attributes.

### DIFF
--- a/usr/lib/ep11_stdll/ep11_specific.c
+++ b/usr/lib/ep11_stdll/ep11_specific.c
@@ -531,6 +531,8 @@ static CK_RV refresh_target_info(STDLL_TokData_t *tokdata);
 static CK_RV get_ep11_target_for_apqn(uint_32 adapter, uint_32 domain,
                                       target_t *target, uint64_t flags);
 static void free_ep11_target_for_apqn(target_t target);
+static CK_RV update_ep11_attrs_from_blob(STDLL_TokData_t *tokdata,
+                                         TEMPLATE *tmpl);
 
 
 /* defined in the makefile, ep11 library can run standalone (without HW card),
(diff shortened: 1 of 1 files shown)


usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -11984,3 +12019,87 @@ static void put_target_info(STDLL_TokData_t *tokdata,
...
+    ep11_target_info_t* target_info;
+    CK_RV rc = CKR_OK;
+    CK_ULONG i;
+
+    CK_ATTRIBUTE ibm_attrs[] = {
+        { CKA_IBM_RESTRICTABLE, &restr, sizeof(restr) },
+     /* Skip CKA_IBM_NEVER_MODIFIABLE for now, it causes CKR_ARGUMENTS_BAD
+        { CKA_IBM_NEVER_MODIFIABLE, &never_mod, sizeof(never_mod) }, */
+        { CKA_IBM_USE_AS_DATA, &useasdata, sizeof(useasdata) },
+        { CKA_IBM_ATTRBOUND, &attrb, sizeof(attrb) },
```
juergenchrist, line 12045: With this change, `CKA_IBM_ATTRBOUND` is attached to every EP11 key.  This makes some `if` statements in the AB implementation trivial.  I did not add the attribute everywhere since attribute lookup in OCK is linear in the number of elements and adding an element is prepending (i.e., the commonly used attributes like key type, modulus, prime, ... are moved even further away).

juergenchrist (approved):

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: Without adding these attribute to the OCK object, an application can't retrieve them (unless they were specified at object creation in the template). Since the EP11 key blob does contain these attributes in any case, I think they should also be visible in the OCK object in any case. 

Yas this increases the linear search time, but I think searching in about 20 attributes does not really cause much trouble.

p-steuer (approved):

PR #426, opened 2021-07-20
[COMMON,ICSF] Fix permission of token SHM segment.

juergenchrist:
The shm token segments were generated with permission 0666 for the user that
created the segment and the corresponding primary group.  This is potentially
a security problem since especially the segmant was word-rw.

This patch changes the permission to 0660 and changes the ownership of the
segment on creation to group "pkcs11".  Since every user that should be
allowed to use OCK has to be in that group, we should not have any usage
regressions.  If the segments already exist, a reboot might be necessary (or
any other way that cleans up the segments).

### DIFF
--- a/usr/lib/common/shared_memory.c
+++ b/usr/lib/common/shared_memory.c
@@ -29,6 +29,7 @@
 #include <stddef.h>
 #include <string.h>
 #include <syslog.h>
+#include <grp.h>
 
 /* For logging functions: */
 #include "pkcs11types.h"
@@ -159,6 +160,7 @@ int sm_open(const char *sm_name, int mode, void **p_addr, size_t len, int force)
     struct shm_context *ctx = NULL;
     size_t real_len = sizeof(*ctx) + len;
     int created = 0;
+    struct group *grp;
 
     /*
      * This is used for portability purpose. Please check `shm_open`
@@ -169,6 +171,14 @@ int sm_open(const char *sm_name, int mode, void **p_addr, size_t len, int force)
         goto done;
     }
 
+    grp = getgrnam("pkcs11");
+    if (!grp) {
+        rc = -errno;
+        SYS_ERROR(errno, "getgrname(\"pkcs11\"): %s\n",
+                strerror(errno));
+        goto done;
+    }
+    
     /* try and open first... */
     fd = shm_open(name, O_RDWR, mode);
     if (fd < 0) {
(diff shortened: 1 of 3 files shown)


usr/lib/common/shared_memory.c
```diff
@@ -202,6 +218,18 @@ int sm_open(const char *sm_name, int mode, void **p_addr, size_t len, int force)
         goto done;
     }
 
+    /*
+     * If the shared memory segment does not belong to the pkcs11 group or does
+     * not have correct permissions, do not use it.
+     */
+    if (stat_buf.st_gid != grp->gr_gid ||
+        (stat_buf.st_mode & ~S_IFMT) != (unsigned int)mode) {
+        TRACE_ERROR("SHM segment has wrong gid/mode combination (expected: %u/0%o; got: %u/0%o)\n",
```
ifranzki, line 227: Should this message maybe also go to the Syslog? How would a user know what's wrong when this hits? 

This might be especially important for migration cases, i.e. when one migrates from a version without this change to a version with this change. As you mention in the commit message, a reboot might be required to force the re-creation of the segments.

ifranzki:
The pkcsslotd also has its own shared memory segment, although create differently. Can you please double check if the permissions and group ownership needs to be fixed there, too (sbin/pkcslotd/shmem.c)?

usr/lib/common/shared_memory.c
juergenchrist: This is in the user program, not the pkcsslotd.  So I thought we should not put this in syslog but use the regular tracing.  At least i used it when my `pkcsconf` showed 0 slots (just remove the `& ~S_IFMT` and you will get the error).  Do we really want this in the syslog?

juergenchrist:
> The pkcsslotd also has its own shared memory segment, although create differently.

Just to mix things up, here we use SysV IPC...  It sets 0660 with group "pkcs11" already (or fails if it cannot set this group).  So this PR just adjusts the POSIX shm segments to the SysV IPC segments...

usr/lib/common/shared_memory.c
ifranzki: Well, I would say that failures due to setup problems should probably go to syslog. We do this for other situations with setup-type errors, too. 
Example: api_interface.c:
```
        OCK_SYSLOG(LOG_ERR, "C_Initialize: Module failed to attach to "
                   "shared memory. Verify that the slot management "
                   "daemon is running, errno=%d\n", errno);
```
Other, regular, non-setup related errors should of course go to trace only.

juergenchrist: Changed the `TRACE_ERROR` to `OCK_SYSLOG`.

ifranzki: Thanks. Just one minor thing: We usually have both, TRACE **and** SYSLOG, so that you  also see in the trace which messages were written to syslog (i.e. if you only have the trace output, but not the syslog).

juergenchrist: Fixed.

ifranzki (approved):

p-steuer (approved):

PR #428, opened 2021-07-26
CI: Perform cleanup of left over EP11 sessions

ifranzki:
Allow to cleanup left over EP11 sessions for CI runs via 'make ci-cleanup'. This is to be called by the CI after 'make ci-installcheck', and before pkcsslotd is terminated.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -231,6 +231,9 @@ ci-installcheck: ci-prepare installcheck
 	killall -HUP pkcsslotd
 	@echo "done"
 
+ci-cleanup:
+	cd ${srcdir}/testcases && PKCS11_USER_PIN=$(PKCS11_USER_PIN) PKCSLIB=@libdir@/opencryptoki/libopencryptoki.so ./cleanup_vhsm.exp 42
+
 ci-uninstall: uninstall
 	rm -f $(sysconfdir)/opencryptoki/ep11tok*.conf
 	rm -rf $(localstatedir)/lib/opencryptoki/*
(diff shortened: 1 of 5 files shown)


juergenchrist (approved):
lgtm

ifranzki:
Well, I guess it still needs some rework:
It should use the PIN ('01234567') that was set up by init_token.sh also for the ci-cleanup step, or, if PKCS11_USER_PIN is set, use that one.

init_vhsm.exp also should use the pin fromPKCS11_USER_PIN  if it is set, I guess.

ifranzki:
> It should use the PIN ('01234567') that was set up by init_token.sh also for the ci-cleanup step, or, if PKCS11_USER_PIN is set, use that one.

Well, that's alrady done by:
`PKCS11_USER_PIN ?= 01234567`

> init_vhsm.exp also should use the pin fromPKCS11_USER_PIN if it is set, I guess.

Will add tis in a second commit, to pass PKCS11_USER_PIN and PKCS11_VHSM_PIN to init_vhsm.exp

juergenchrist:
> Will add tis in a second commit, to pass PKCS11_USER_PIN and PKCS11_VHSM_PIN to init_vhsm.exp

Maybe while you are at it, you could pass PKCS11_USER_PIN also to init_token.sh and use it there?  It seems to have _all_ values hardcoded...

juergenchrist (approved):

p-steuer (approved):

PR #434, opened 2021-07-30
travis: adapt ppc64le targets

p-steuer:
Signed-off-by: Patrick Steuer <patrick.steuer@de.ibm.com>


ifranzki (approved):
OK, it now this PR gets packages from http://ports.ubuntu.com/ubuntu-ports focal-updates/main so from focal. Lets hope that the same then applies to the OpenSSL PR.

PR #440, opened 2021-08-16
TESTCASE: DSA: Destroy objects created during testing

ifranzki:
Leaving token object hanging around after the testcase may cause trouble with testcases executed afterwards, e.g. the obj_mgmt_tests does not expect to find any other token objects except those it created itself. 

FYI: @chrisRund


juergenchrist (approved):
lgtm

p-steuer (approved):

PR #442, opened 2021-08-17
CI: Temporarily disable slot 42

ifranzki:
To be reverted once we have corrected the EP11 session issue in the CI.


p-steuer (approved):
s390x build seems not to be working. all others okay.

PR #443, opened 2021-08-18
EP11: m_GetAttributeValue() may fail on a CEX5 after m_DeriveKey()

ifranzki:
For whatever reason, the m_GetAttributeValue() call to obtain attributes from an EP11 key blob may fail when using a CEX5 crypto card when using a key blob that was just derived via m_DeriveKey() on the same card.

In such error case, extract the boolean blob attributes from the blob directly. The first part of the EP11 key blob is not encrypted, and it contains a copy of the boolean attributes at offset 48. See EP11 structure document section "3. Key objects (blobs)" for the full structure of the key blob.


ifranzki:
Hm, it seems that the key blob is really wrong. It can't be used for crypto operations, too.  So this fix is not the right way to fix this.

PR #444, opened 2021-08-20
Testcases: EP11 can't derive keys <= 8 bytes on a CEX5

ifranzki:
The CEX5P seems to have a firmware bug that hinders it from deriving a valid EP11 key blob for a derived key size <= 8. Enhance the testcases to identify situations where invalid/unusable key blobs are produced, but also skip tests the derive keys <= 8 bytes for the EP11 token until the firmware bug is resolved.

Note: For EP11, the smallest really usable CKK_GENERIC_SECRET key is 10 bytes anyway. AES or DES/3DES keys are larger. Keys of type CKK_GENERIC_SECRET can be used for HMAC mechanism only, and HMAC usually requires a minimum key size of digest-size / 2. For HMAC-SHA-1 the minimum key is therefore 20 / 2 = 10 bytes. This key size restriction an be lifted via control point XCP_CPB_KEYSZ_HMAC_ANY for EP11.

### DIFF
--- a/testcases/common/common.c
+++ b/testcases/common/common.c
@@ -102,6 +102,10 @@ int check_supp_keysize(CK_SLOT_ID slot_id, CK_ULONG mechanism, CK_ULONG keylen)
     if (rc != CKR_OK)
         return FALSE;
 
+    /* min and max being zero indicate no key size limitation */
+    if (mech_info.ulMinKeySize == 0 && mech_info.ulMaxKeySize == 0)
+        return TRUE;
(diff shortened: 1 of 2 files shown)


juergenchrist (approved):
lgtm.  Did we report this to the EP11 guys?

ifranzki:
> Did we report this to the EP11 guys?
Yes Jup has been opened for this.

p-steuer (approved):

PR #446, opened 2021-08-24
SOFT: Detect unsupported EC curves with OpenSSL 3.0

ifranzki:
OpenSSL 3.0 behaves different in reporting an error when an unsupported EC curve is used to generate an EC key. OpenSSL 1.1.1 returns an error at EVP_PKEY_CTX_set_ec_paramgen_curve_nid() already, but OpenSSL 3.0 returns
an error only at EVP_PKEY_keygen().


juergenchrist (approved):
lgtm

p-steuer (approved):

PR #447, opened 2021-08-24
COMMON: Remove unused code

ifranzki:
Functions encrypt_data() and decrypt_data() would only be used for tokens that are no secure-key tokens, and use an encryption algorithm for the MK_USER/MK_SO file other than CKM_DES3_CBC (for the old token format only).

The only token that uses anything other than CKM_DES3_CBC is the TPM token, but it is flagged as secure key token. The ICSF token specifies 0 as mechanism, but it does not use this code path at all.

In general, there should be no reason why the token's own encryption and decryption functions should be used instead of Opencryptoki's software implementation (based on OpenSSL). Also the removed functions do not provide any additional mechanisms than CKM_DES3_CBC and CKM_AES_CBC, which are also covered by the general software implementation.

Besides that, there was a memory leak inside the removed functions. It never really leaked any memory because it was never used.

### DIFF
--- a/usr/lib/common/loadsave.c
+++ b/usr/lib/common/loadsave.c
@@ -390,74 +390,6 @@ done:
     return initial_vector;
 }
 
-static CK_RV encrypt_data(STDLL_TokData_t *tokdata, CK_BYTE *key,
-                          CK_ULONG keylen, const CK_BYTE *iv,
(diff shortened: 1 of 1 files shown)


juergenchrist (approved):
lgtm

p-steuer (approved):

PR #448, opened 2021-08-25
API: Retain the OpenSSL error stack state on API calls

ifranzki:
Internal use of OpenSSL calls may cause additional errors to be pushed to the current thread's OpenSSL error stack. This may interfere with the calling application's use of the OpenSSL error stack, and may leak Opencryptoki internal error information to the application.

On exit of all Opencryptoki API calls, make sure to restore the current thread's error stack to the state it was on entry to the Opencryptoki API call.

### DIFF
--- a/usr/lib/api/api_interface.c
+++ b/usr/lib/api/api_interface.c
@@ -35,15 +35,19 @@
 #include "trace.h"
 #include "ock_syslog.h"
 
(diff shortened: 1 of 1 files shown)


juergenchrist (approved):
lgtm

p-steuer (approved):
would make sense to add that to libica as well

PR #449, opened 2021-08-26
Add SW fallbacks to ICA token and rework of OpenSSL based soft token implementation

ifranzki:
Move OpenSSL based SW implementations from soft token to a new common code source file. Use them in the soft token, but also in the ICA token as SW fallbacks, in case libica does not provide an implementation. Also unify certain mechanism implementations where the code was almost equal and use a single generic implementation for all of those.

While this PR looks like lost of code changes, it really just moves lots of code around (from soft_specific.c to openssl_mech.c). In a later commit it then also unifies some of the moved functions to a single generic function, thus throwing away several of the previously moved lines. So it looks like more code changes that it really is :-)


juergenchrist (approved):
lgtm now

usr/lib/common/mech_openssl.c
```diff
@@ -0,0 +1,4617 @@
...
+    EVP_PKEY_CTX *ctx = NULL;
+
+    rc = template_attribute_get_ulong(publ_tmpl, CKA_MODULUS_BITS, &mod_bits);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("%s\n", ock_err(ERR_TEMPLATE_INCOMPLETE));
+        return CKR_TEMPLATE_INCOMPLETE; // should never happen
+    }
+
+    // we don't support less than 512 bit keys in the sw
+    if (mod_bits < 512 || mod_bits > 4096) {
```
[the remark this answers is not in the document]
ifranzki: The soft token itself still uses 4096 as max key size supported in the mechanisms info. I could actually use OPENSSL_RSA_MAX_MODULUS_BITS for the max key size for the RSA mechs in the soft token, too. If that's wanted, then I guess this would better be a separate/new commit. 
Opinions?

ifranzki:
Added one more commit to check in the ICA token if libica supports the curve when importing an EC key object via C_CreateObject

usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -4846,3 +4914,60 @@ CK_RV token_specific_ecdh_pkcs_derive(STDLL_TokData_t *tokdata,
...
+CK_RV token_specific_object_add(STDLL_TokData_t *tokdata, SESSION *sess,
+                                OBJECT *obj)
+{
+    ica_private_data_t *ica_data = (ica_private_data_t *)tokdata->private_data;
+    CK_OBJECT_CLASS class;
+    CK_KEY_TYPE keytype;
+#ifndef NO_EC
+    ICA_EC_KEY *eckey = NULL;
+    unsigned int privlen;
+    EVP_PKEY *ec_key = NULL;
```
juergenchrist, line 4927: I really don't like this naming.  why not `ica_eckey` and `ossl_eckey`?  Sure, it is more typing, but it better captures what the variables are about (and not only differs in a meaningless underscore).

ifranzki: you are right. Changed it to the suggested names.

ifranzki:
Corrected MAX_RSA_KEYLEN in defs.h to be (OPENSSL_RSA_MAX_MODULUS_BITS / 8) and not hardcoded 1920 (which would have be for RSA keys up to 15360 bits only), as mech_openssl.c uses this definition for buffer sizes.

p-steuer (approved):

PR #450, opened 2021-08-30
Configlib

juergenchrist:
Add generic config library.  The config library is able to parse generic configurations and
opencryptoki.conf.  It is able to track comments and newlines to support
configuration migration tools.


jschmidb:
I saw that in icsf_specific.c there is a call to an old routine parse_config_file() with different parms than the newly introduced parse_config_file routine in cfgparse.y. Let's check that we don't get any name conflicts.

jschmidb:
The config parser operates on an open config file. There are lots of file operation like fputs, but no fopen and fclose. So it's assumed that the config file is opened and closed by the token init routine when using the new parser?

juergenchrist:
> So it's assumed that the config file is opened and closed by the token init routine when using the new parser?

Safety checks and error messages on missing files or files that cannot be opened for reading are up to the caller.  The parser only returns its parsing state, i.e., if the file could be parsed or if a parse error terminated the run.  When you `fopen`/`fclose` the file is up to you.

juergenchrist:
> also here: n always != null? (and some more places below .... )

Which line?  In general, all these functions are public API of the configuration AST.  So they should better be NULL-safe or explicitly state that NULL arguments are not allowed.  I decided to make them NULL-safe even though when they are called from other functions of the config lib, they pointer should not be null (except for `confignode_deepfree` which serves as the main free function).

juergenchrist:
> I saw that in icsf_specific.c there is a call to an old routine parse_config_file() with different parms than the newly introduced parse_config_file routine in cfgparse.y. Let's check that we don't get any name conflicts.

Good finding.  I totally missed that.  I think we should rename the parse function in the config lib.  How about `parse_configlib_file`?

jschmidb:
> How about parse_configlib_file?

Yes, why not ...

jschmidb:
> also here: n always != null? (and some more places below .... )
> Which line? 

Hm, I added the comment to the code, don't know why it doesn't display here ... anyway: in routines like confignode_dumpstruct, confignode_dumpidxstruct, etc. n is often dereferenced without a null check. So the question was, are we sure n is always != null there ....

juergenchrist:
> Can you check the description of this commit: should "but the separate build tree" read "but in the separate build tree" ?

Fixed.

juergenchrist:
> > How about parse_configlib_file?
> 
> Yes, why not ...

Fixed.

ifranzki:
May be not related, but worth a check: usr/sbin/pkcsicsf/pkcsicsf.c also seems to read in the opencryptoki config file and modifies it (adds an entry). I guess this all open coded, no parser used so far. Can you check if this could be changed to use the parser the way you implemented it now somewhere in the future?

juergenchrist:
> Can you check if this could be changed to use the parser the way you implemented it now somewhere in the future?

As far as I can see this code is appending new slots to the opencryptoki.conf and creating separate configuration files for these tokens.  In the current version, we would have to read the whole configuration into memory (with comment tracking enabled), modify the AST in memory, and dump the modified version to a new file.  One could think about appending some partial configuration, but personally I do not like this idea since it is hard to predict if the result will be a valid configuration.  Just imagine a configuration that does not end in a new line and you simply append.  In that case you have to add a new line to the end of the file or you might end up with an invalid file.  So better dump the whole configuration (and ensure the dumper is correct).

ifranzki:
> ... and dump the modified version to a new file

I guess this can be done using confignode_dump(), right? 

> One could think about appending some partial configuration, but personally I do not like this idea ....

Right, I don't like this either. So read the existing config in, modify AST and then dump it would be the right way to do this. The same procedure should then be used for pkcstok_migrate that also modifies the config.

juergenchrist:
> > ... and dump the modified version to a new file
> 
> I guess this can be done using confignode_dump(), right?

Yes, that is the only purpose of this function.  But of course it is sensitive to EOL-nodes (which are the only nodes introducing a new line).  So every modification should (to be on the save side) start with an EOL line (maybe first one without comment and then one with a comment of the form "Automatically added/migrated") and then the real content.  Well, it would be simpler for migration tools to require an EOL after every configuration element.  We anyway need to take care of this when creating ASTs programmatically and dumping them to a file...

juergenchrist:
> So every modification should (to be on the save side) start with an EOL line

Just pushed a fix for this problem to at least add one space between the two configuration elements.  Maybe we should also add some convenience API for the main nodes that not just allocates this node but directly appends an EOC-node to it (such that it actually produces a new line during the dump)?  Otherwise it is easy to forget it and end up with an ugly configuration.  An alternative is to go the UNIX conf way and treat everything up to the end of the line as a configuration value.  Then we would also force an End-of-Line terminator for every config element.

hfreude (approved):
This config lib looks totally nice to me.
I tried to read and understand the lex and yacc stuff - the lexer stuff is fine to read and understand and looks like it is doing what's needed - nice with the state for MAIN and INITIAL. I also tried to read the yacc stuff - but I have to give up on this grammar rules. Instead I tried to check the assignments on the action rules.  My recommendation for this rules may be to add an example at each rule line so that an innocent may get a clue on what this rule may cover.
Btw. in kernel code a macro container_of() is used very often which does the very same as your offsetof lines to in configuration.h. This would general make sence to add it to ock. However, the kernel macro is:
#define container_of(ptr, type, member) ({				\
	void *__mptr = (void *)(ptr);					\
	BUILD_BUG_ON_MSG(!__same_type(*(ptr), ((type *)0)->member) &&	\
			 !__same_type(*(ptr), void),			\
			 "pointer type mismatch in container_of()");	\
	((type *)(__mptr - offsetof(type, member))); })

juergenchrist:
Added convenience functions for programmatic AST manipulation (they automatically add an EOC on allocation with an optional comment), commented and simplified grammar, and updated our definition of bare words.

jschmidb (approved):
lgtm

ifranzki (approved):
lgtm

p-steuer (approved):

PR #453, opened 2021-09-03
Fix pkcstok_migrate and pkcsslotd to handle version keywords correctly

ifranzki:


ifranzki:
@juergenchrist Does your new parser also allow the version keywords to be either quoted or un-quoted?

juergenchrist:
> Does your new parser also allow the version keywords to be either quoted or un-quoted?

No, only unquoted versions are allowed.  Honestly I do not see a benefit in including newline characters in the version (which the current parser would do for quoted strings).

juergenchrist:
> > Does your new parser also allow the version keywords to be either quoted or un-quoted?
> 
> No, only unquoted versions are allowed. Honestly I do not see a benefit in including newline characters in the version (which the current parser would do for quoted strings).

Sorry, misunderstood what you mean.  The new parser would create different nodes:

juergenchrist (approved):
This version stuff was a mess from the beginning.  I cannot remember correctly, but I think I did not change the type and originally the token version was a version while the hardware and firmware version were strings.  In general I am not fond of the idea to have versions as strings.  At least I have never seen the need to use newlines in a version...

juergenchrist:
> The new parser would create different nodes

We could simply add a function `confignode_get_version` which returns the version if we have a `ConfigNodeVersionVal` node, parses the version with something like `sscanf(value, "%u.%u", major, minor)` if we have a `ConfigNodeStringVal` node, or returns an error if we have any other node.  I already have such a convenience function for strings where I switch between `ConfigNodeStringVal` and `ConfigNodeBareVal` nodes (i.e. between quoted and unquoted string values).

ifranzki:
> This version stuff was a mess from the beginning.

I totally agree!

Unfortunately we currently have the situation, that the value of the `hwversion` and `firmwareversion` keywords MUST be quoted (without the fix in this PR), but the value of the `tokversion` keyword MUST NOT be quoted....

We can't really change this to all un-qouted or all quoted without breaking compatibility. So therefore I would suggest we should allow both, quoted and un-quoted for all version keywords (this PR achieves this). Still kind of a mess, but at least a somewhat more consistent mess from the user's point of view.  Yes, needs more code to support that (also with your new parser in the future - your `confignode_get_version` function would definitely help here), but I don't really see a better way....

juergenchrist:
> also with your new parser in the future - your `confignode_get_version` function would definitely help here

Added the function (at the end of configuration.h as an inline function).  See configlib PR.  I also added some Macros to unpack the version from an `unsigned int` into a major and a minor component.

p-steuer (approved):

PR #456, opened 2021-09-16
Introduce p11sak config file for list-key --long

finn-callies:
This pull request provides an extension for the long listing output of the p11sak utility. It introduces a config file to specify attributes that are printed in addition to the currently supported PKCS#11 attributes when using the p11sak long output format.


finn-callies:
Closing as discussed, new commit structure will be provided

PR #461, opened 2021-09-29
pkcsep11_session: Display max and available EP11 sessions per APQN

ifranzki:
Enhance the pkcsep11_session tool to provide a new sub-command 'status' that shows the maximum and currently available number of EP11 sessions for each available (and online) EP11 APQN.

This is inspired by @juergenchrist and his `ep11sessions` tool.


ifranzki:
close/open to trigger travis... but unfortunately travis was still not trigged?

juergenchrist (approved):
Looks familiar...

usr/sbin/pkcsep11_session/pkcsep11_session.c
```diff
@@ -972,6 +980,86 @@ static CK_RV set_vhsmpin(CK_SESSION_HANDLE session)
     return CKR_OK;
 }
 
+static CK_RV status_handler(uint_32 adapter, uint_32 domain,
+                            void *handler_data)
+{
+    ep11_target_t target_list;
+    struct XCP_Module module;
+    target_t target = XCP_TGT_INIT;
+    uint32_t res[16 * 2];
```
juergenchrist, line 989: Btw. this is just what I extracted from the current document.  `m_get_xcp_info` follows PKCS output conventions and we could do a size query and allocate a properly-sized output buffer.  Just in case future cards have different output.

ifranzki: I tried with `int32_t res[CK_IBM_XCPXQ_MAXIDX * 2];` (where CK_IBM_XCPXQ_MAXIDX is currently 13), but I got CKR_BUFFER_TOO_SMALL.... So I decided to go with 16....

juergenchrist: I meant do a first call with `NULL` buffer to get the correct length and then allocate the buffer.  Then it should always be proper.  But yes, I know that problem...

ifranzki: Hm, tried that, but gut reslen=0 after the size query....
```
1021        rc = dll_m_get_xcp_info(NULL, &reslen, CK_IBM_XCPQ_EXT_CAPLIST, 0, target);
1: reslen = 0
2: rc = 0
```
I guess I need to do a CK_IBM_XCPQ_EXT_CAPS query to get the capabilities count and allocate the buffer based on that.

ifranzki: Hmm CK_IBM_XCPQ_EXT_CAPS also returns zero (If I am doing it right):
```
1022        rc = dll_m_get_xcp_info(&caps, &reslen, CK_IBM_XCPQ_EXT_CAPS, 0, target);
1: reslen = 8
2: caps = 0
3: rc = 0
```

ifranzki: OK, got CK_IBM_XCPQ_EXT_CAPS to work, not sure why it did not work initially....

PR #463, opened 2021-10-01
Config: Add BARE constants and some bugfixes

juergenchrist:
The inital configlib commit did not include the possibility to use bare word
    constants.  This is needed, e.g., for the "disable-event-support" keyword of
    opencryptoki.conf.
    
    To support this, add ConfigNodeBareConst to keep the constant in the key
    without any further values in the node.  Also restructure the grammar:
    - Every configelem now takes care of the positions of comments _within_ and
    _after_ the element.
    - Since "after element i" is "before element (i+1)", the config element cannot
    take care of comments before the element (shift/reduce conflict since the
    parser cannot know if the comment is after element i or before element i+1).
    To alleviate this situation and preserve comments at the beginning of a file
    or a structure content, add a commentedconfigelemstar that only takes care of
    comments before the first configelem.  That should cover all comment
    positions.

   Fix incorrect version masks in dump of ConfigVersionValNode.

    confignode_deepfree did not free elements in BARELISTs causing memory leaks.

    BUILT_SOURCES should not contain the generated headers.  They are generated
    and updated via ylwrap which does a content comparison for all but the
    generated parser/lexer C sources.  If the inputs to bison or flex change but
    still produce the same header file, make install will regenerate the parser
    and lexer, recompile them, and relink all binaries that use the config lib.
    By using the generated C files in the BUILT_SOURCES variable we solve this
    problem since the C files are always updated by ylwrap (even if they did not
    change).

### DIFF
--- a/usr/lib/config/cfgparse.y
+++ b/usr/lib/config/cfgparse.y
@@ -33,10 +33,10 @@ typedef void (*error_hook_f)(int line, int col, const char *msg);
 
  
   static inline void configerror(CONFIGLTYPE *lloc, configscan_t scanner,
-  	      	                 struct ConfigBaseNode **res,
-		                 error_hook_f error_hook,
-				 int trackComments,
-			         const char *msg)
+                                 struct ConfigBaseNode **res,
+                                 error_hook_f error_hook,
+                                 int trackComments,
+                                 const char *msg)
   {
     (void)scanner;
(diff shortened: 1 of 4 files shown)


ifranzki (approved):
lgtm

p-steuer (approved):

PR #464, opened 2021-10-01
ICA: Adjust min/max key sizes for RSA mechanisms according to libica's support

ifranzki:
libica may support RSA keys > 1024 only if running in FIPS mode. Make sure the ICA token adjusts its mon/max key sizes accordingly.
Also make sure testcases only use RSA key sizes supported by the token.

### DIFF
--- a/usr/lib/ica_s390_stdll/ica_specific.c
+++ b/usr/lib/ica_s390_stdll/ica_specific.c
@@ -1928,6 +1928,11 @@ static CK_RV ica_specific_rsa_keygen(STDLL_TokData_t *tokdata,
         rc = CKR_FUNCTION_NOT_SUPPORTED;
         goto privkey_cleanup;
         break;
+    case EPERM:
+        TRACE_ERROR("%s\n", ock_err(ERR_KEY_SIZE_RANGE));
+        rc = CKR_KEY_SIZE_RANGE;
+        goto privkey_cleanup;
+        break;
     default:
         TRACE_ERROR("%s\n", ock_err(ERR_FUNCTION_FAILED));
         rc = CKR_FUNCTION_FAILED;
@@ -2204,6 +2209,10 @@ static CK_RV ica_specific_rsa_encrypt(STDLL_TokData_t *tokdata,
         TRACE_ERROR("%s\n", ock_err(ERR_FUNCTION_NOT_SUPPORTED));
         rc = CKR_FUNCTION_NOT_SUPPORTED;
         break;
+    case EPERM:
+        TRACE_ERROR("%s\n", ock_err(ERR_KEY_SIZE_RANGE));
+        rc = CKR_KEY_SIZE_RANGE;
+        break;
     default:
         TRACE_ERROR("%s\n", ock_err(ERR_FUNCTION_FAILED));
         rc = CKR_FUNCTION_FAILED;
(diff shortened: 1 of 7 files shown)


usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -3589,8 +3589,45 @@ static CK_BBOOL isMechanismHW(STDLL_TokData_t *tokdata, CK_ULONG mechanism)
...
+        min = 512;
+    else if (mask & KEY_SIZE_1024)
+        min = 1024;
+    else if (mask & KEY_SIZE_2048)
+        min = 2048;
+    else if (mask & KEY_SIZE_4096)
+        min = 4096;
+
+    max = min;
+    if ((mask & KEY_SIZE_4096) && min <= 4096)
```
juergenchrist, line 3614: `min <= 4096` is trivially true

usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -3589,8 +3589,45 @@ static CK_BBOOL isMechanismHW(STDLL_TokData_t *tokdata, CK_ULONG mechanism)
...
+        max = 4096;
+    else if ((mask & KEY_SIZE_2048) && min <= 2048)
```
juergenchrist, line 3616: `min <= 2048` is trivially true (although not that easy to see).

usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -3589,8 +3589,45 @@ static CK_BBOOL isMechanismHW(STDLL_TokData_t *tokdata, CK_ULONG mechanism)
...
+        max = 2048;
+    else if ((mask & KEY_SIZE_1024) && min <= 1024)
```
juergenchrist, line 3618: `min <= 1024` is trivially true (although not that easy to see).

usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -3589,8 +3589,45 @@ static CK_BBOOL isMechanismHW(STDLL_TokData_t *tokdata, CK_ULONG mechanism)
...
+        max = 1024;
+    else if ((mask & KEY_SIZE_512) && min <= 512)
```
juergenchrist, line 3620: `min < 512` is trivially true.

usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -3589,8 +3589,45 @@ static CK_BBOOL isMechanismHW(STDLL_TokData_t *tokdata, CK_ULONG mechanism)
...
+#define KEY_SIZE_4096  0x08
+
+static void adjust_key_sizes(unsigned int mask, CK_MECHANISM_INFO *mech_info)
+{
+    CK_ULONG min = 0, max = 0;
+
+    if (mask == 0)
+        return;
+
+    if (mask & KEY_SIZE_512)
```
juergenchrist, line 3604: simpler code would be `min = ((int)mask & -(int)mask) * 512`, but only if the "isolate right-most one" pattern is known and accepted.

usr/lib/ica_s390_stdll/ica_specific.c
ifranzki: ah yes, right ;-)

usr/lib/ica_s390_stdll/ica_specific.c
ifranzki: Hm well, since there is no simple "isolate left-most one" pattern, I would tend to leave both as is....

jschmidb:
> libica may support RSA keys > 1024 only if running in FIPS mode.

libica supports RSA keys > 1024 also when not running in fips mode. But when running in fips mode AND using openssl 3.0, then RSA keys <= 1024 are not supported.

ifranzki:
I still wonder why Travis does not build anymore....
 @p-steuer can you please take a look?

PR #466, opened 2021-10-06
Rework part of token initialization

juergenchrist:
Do not check user/group permissions in every token on token initialization.
The check can easily be done in API library once for all tokens.

Also allocate the TokData structure in the API library.  This allows the API
library to pass other objects to the tokens without changing the signature of
the initialization function.


usr/lib/api/apiutil.c
```diff
@@ -566,6 +567,26 @@ int DL_Load_and_Init(API_Slot_t *sltp, CK_SLOT_ID slotID)
...
+    if (!sltp->TokData) {
+        TRACE_ERROR("Allocating host memory failed.\n");
+        return FALSE;
+    }
+    sltp->TokData->ro_session_count = 0;
+    sltp->TokData->global_login_state = CKS_RO_PUBLIC_SESSION;
+#ifdef ENABLE_LOCKS
+    pthread_rwlock_init(&sltp->TokData->sess_list_rwlock, NULL);
+#endif
+    pthread_mutex_init(&sltp->TokData->login_mutex, NULL);
```
ifranzki, line 588: Its a bit strange that these 2 locks are initialized here (by the API), but are to be destroyed by the tokens itself (in SC_Finalize). 

The same applies to freeing tokdata, this also is done by SC_Finalize...

I guess there even is now some kind of lock-leak in certain error cases here, as well as a possible double free of tokdata:  DL_UnLoad does not destroy the locks but frees the tokdata - tokdata have already been freed by SC_Finalize in some cases.

I don't see a reason why the destroying of these locks and freeing tokdata could not also be moved to the API layer. 

That way things allocated by the API are also freed by the API, and only things allocated by the tokes are to be freed by the tokens.

juergenchrist: `SC_Finalize` does not free the tokdata anymore.  I will move the lock destroy to the `DL_Unload` function before the free of the tokdata.

juergenchrist: Done.

p-steuer (approved):

PR #476, opened 2021-10-08
Release version 3.17.0

p-steuer:


ifranzki:
Can you please change `SWToken - The software token uses OpenSSL version 1.0.2 or higher.` to mention OpenSSL **1.1.1** ? And actually, its not only a requirement of the SW token, but the whole OCK code.

PR #484, opened 2021-10-14
Policies

juergenchrist:
Implement strength and policy parsing and enforce policies in all tokens.  The
main object of the policy is the struct objstrength which carries the strength
of a key along with a flag specifying if the EC is allowed, and a (maximal)
signature length if this key is used to sign or verify and the length of the
signature depends on the key.

A valid strength configuration has to be found.  If no policy configuration is
found, policies are simply deactivated and keys are only classified according
to the strength definition, but can be used without restrictions.  If both, a
strength and a policy configuration, are available, both have to be valid for
OpenCryptoki to start.  If either of these two configurations fails to parse
or validate, OpenCryptoki will fail with a general error since the system is
not in a valid state.

The configuration files have to be owned by root:pkcs11 and have to have modes
0640.  This is done for security reasons to prevent anybody except root from
altering the policy or strength definitions.

This PR also introduces a global mechanism table with generated accessors for the string or the numeric representation of a mechanism as well as an exploitation of that table to remove a number of switches and lookup tables over the mechanisms.


jschmidb:
> Can you tell which testcases fail due to to key sizes?

I had the fails with the aes_tests.

ifranzki:
> I had the fails with the aes_tests.

OK, they don't seem to check the key sizes yet. Up to now it was save to assume that when AES is supported all AES key sizes were supported. We should probably add a key size check to these testcases and skip those key sizes that are not supported.

Can you confirm that `pkcsconf -m` shows the right key sizes for the AES mechanisms (i.e. `Key Size: 32-32`)?

jschmidb:
Yes, this looks good:
```
Mechanism #0
	Mechanism: 0x1080 (CKM_AES_KEY_GEN)
	Key Size: 32-32
	Flags: 0x8000 (CKF_GENERATE)
Mechanism #1
	Mechanism: 0x1081 (CKM_AES_ECB)
	Key Size: 32-32
	Flags: 0x300 (CKF_ENCRYPT|CKF_DECRYPT)
Mechanism #2
	Mechanism: 0x1082 (CKM_AES_CBC)
	Key Size: 32-32
	Flags: 0x60300 (CKF_ENCRYPT|CKF_DECRYPT|CKF_WRAP|CKF_UNWRAP)
Mechanism #3
	Mechanism: 0x1085 (CKM_AES_CBC_PAD)
	Key Size: 32-32
	Flags: 0x60300 (CKF_ENCRYPT|CKF_DECRYPT|CKF_WRAP|CKF_UNWRAP)
```

ifranzki:
> Just saw I'm also getting errors with the ec_tests ...

Probably because it does not support all curves due to policy (curve size) restrictions. Again, the min/max key sizes of the mechanism should somehow tell you which curve sizes are supported.

jschmidb:
Agreed

ifranzki:
> Lets collect those kind of testcase enhancements (maybe in a new github issue?) and then we can work on it independent of this PR. 

This should now be fixed in the testcases with commit https://github.com/opencryptoki/opencryptoki/pull/484/commits/5a56c6d1bf06a5dfd642de66cb5ab2c5053b2110 also included in this PR.

usr/lib/api/policy.c
```diff
@@ -0,0 +1,1728 @@
...
+        case CKM_IBM_CMAC:
+        case CKM_SHA1_KEY_DERIVATION:
+        case CKM_SHA224_KEY_DERIVATION:
+        case CKM_SHA256_KEY_DERIVATION:
+        case CKM_SHA384_KEY_DERIVATION:
+        case CKM_SHA512_KEY_DERIVATION:
+        case CKM_SSL3_MASTER_KEY_DERIVE:
+        case CKM_SSL3_PRE_MASTER_KEY_GEN:
+            if (policy_update_symmetric(pp, info, CK_TRUE) != CKR_OK) {
+                TRACE_DEVEL("Mechanism 0x%lx blocked by policy!\n", mech);
```
ifranzki, line 849: Could the mechanism name be printed here?

usr/lib/api/policy.c
```diff
@@ -0,0 +1,1728 @@
...
+            /* For these mechanisms, the standard does not specify a
+               value for info.  So we only pass them to the generic
+               check for signature size. */
+            break;
+        default:
+            TRACE_ERROR("Mechanism 0x%lx unknown to policy!\n", mech);
+            return CKR_MECHANISM_INVALID;
+        }        
+        if (policy_check_signature_size(pp, mech, info) != CKR_OK) {
+            TRACE_DEVEL("Policy removes SIGN/VERIFY from 0x%lx\n", mech);
```
ifranzki, line 999: print mechanism name

usr/lib/api/policy.c
```diff
@@ -0,0 +1,1728 @@
...
+            info->flags &= ~(CKF_SIGN | CKF_SIGN_RECOVER |
+                             CKF_VERIFY | CKF_VERIFY_RECOVER);
+            if ((info->flags &
+                    ~(CKF_HW|CKF_EC_F_P|CKF_EC_NAMEDCURVE|CKF_EC_UNCOMPRESS)) == 0) {
+                TRACE_DEVEL("Mechanism 0x%lx does not provide any feature after policy adjustment!\n",
```
ifranzki, line 1004: dito

usr/lib/api/policy.c
```diff
@@ -0,0 +1,1728 @@
...
+    } else {
+        s->strength = POLICY_STRENGTH_IDX_0;
+        s->siglen = 0;
+        s->allowed = CK_TRUE;
+    }
+    return rv;
+}
+
+static CK_RV policy_is_mech_allowed(policy_t p, CK_MECHANISM_PTR mech,
+                                    struct objstrength *s,int check,
```
ifranzki, line 726: blank missing after `*s,`

usr/lib/api/policy.c
```diff
@@ -0,0 +1,1728 @@
...
+    TRACE_WARNING("POLICY VIOLATION: prf not allowed: 0x%lx\n", val);
+    return CKR_FUNCTION_FAILED;
+}
+
+static CK_RV policy_is_key_allowed_i(struct policy_private *pp,
+                                     struct objstrength *s)
+{
+    /* 1. Check strength */
+    if (s->strength > pp->minstrengthidx) {
+        TRACE_WARNING("POLICY VIOLATION: Key does not satisfy minimal strength constraint\n");
```
ifranzki, line 441: Its hard to know for which key / object such trace message is issued:

Example from my token init output:
```
11/03/2021 10:01:36 3125 [usr/lib/api/policy.c:441 api] WARN: POLICY VIOLATION: Key does not satisfy minimal strength constraint
11/03/2021 10:01:36 3125 [usr/lib/api/policy.c:441 api] WARN: POLICY VIOLATION: Key does not satisfy minimal strength constraint
11/03/2021 10:01:36 3125 [usr/lib/api/policy.c:441 api] WARN: POLICY VIOLATION: Key does not satisfy minimal strength constraint
11/03/2021 10:01:36 3125 [usr/lib/api/policy.c:441 api] WARN: POLICY VIOLATION: Key does not satisfy minimal strength constraint
11/03/2021 10:01:36 3125 [usr/lib/api/policy.c:441 api] WARN: POLICY VIOLATION: Key does not satisfy minimal strength constraint
11/03/2021 10:01:36 3125 [usr/lib/api/policy.c:441 api] WARN: POLICY VIOLATION: Key does not satisfy minimal strength constraint
```
However, I also don't know how to make this better.... So you may just leave it as is.

usr/lib/api/policy.c
```diff
@@ -0,0 +1,1728 @@
...
+    strength->flags = 1;
+    reqstrength = confignode_to_intval(strength)->value;
+    for (i = 0; i < NUM_SUPPORTED_STRENGTHS; ++i) {
+        if (reqstrength == supportedstrengths[i]) {
+            TRACE_DEVEL("POLICY: Using strength %lu\n", supportedstrengths[i]);
+            break;
+        } else if (reqstrength > supportedstrengths[i]) {
+            if (i > 0)
+                i -= 1;
+            TRACE_DEVEL("POLICY: Using strength %lu\n", supportedstrengths[i]);
```
ifranzki, line 1535: so you round up the specified strength value to the next higher supported strength..... That's fine, just the man page does not mention that.....

usr/lib/api/policy.c
```diff
@@ -0,0 +1,1728 @@
...
+            break;
+        }
+    }
+    if (i == NUM_SUPPORTED_STRENGTHS && reqstrength > 0) {
+        i -=1;
+        TRACE_DEVEL("POLICY: Using strength %lu\n", supportedstrengths[i]);
```
ifranzki, line 1541: ... and if its higher than the highest supported strength, you round down.....

usr/lib/api/policy.c
juergenchrist: done

usr/lib/api/policy.c
juergenchrist: done

usr/lib/api/policy.c
juergenchrist: done

usr/lib/api/policy.c
juergenchrist: done

usr/lib/api/policy.c
juergenchrist: Since at that point I have no other information, the caller would have to provide additional information.  I cannot do anything here (except maybe print the strength that was determined, but this does not really help).

usr/lib/api/policy.c
juergenchrist: done

usr/lib/api/policy.c
juergenchrist: done

ifranzki (approved):
Great work!

p-steuer (approved):

Issue #481, opened 2021-10-13
opencryptoki.conf parser: version values quoted with single quotes are parsed incorrectly


ifranzki:
A configuration like this
```
slot 2
{
tokversion = '3.12'
stdll = libpkcs11_cca.so
}
```
parses error free, but results in tokversion = 0.0.

The problem is that single quoted strings are treated as string tokens and include the single quotes in the string. Furthermore the do_vers() function in slotmgr.c does not check for number parsing errors:

```
static int do_vers(struct parse_data *d,
                   CK_VERSION *slotinfo, int kw, const char *val)
{
	char *dot;

	if (!val || !*val) {
		snprintf(d->errbuf, sizeof(d->errbuf), "%s has no value\n",
                 keyword_token_to_str(kw));
		return -1 ;
	}

    dot = strchr(val, '.');
    slotinfo->major = strtol(val, NULL, 10);    <----- no parsing error checking
    slotinfo->minor = dot ? strtol(dot + 1, NULL, 10) : 0; <----- no parsing error checking
	return 0;
}
```

ifranzki:
@juergenchrist How does your new parser handle single quoted strings?

PR #487, opened 2021-11-03
Change pkcsslotd to use the generic config file parser

ifranzki:
Change pkcsslotd (and pkcstok_migrate) to use the new generic config file parser and get rid of the old one.
Also updated pkcsicsf and the ICSF token to use the generic parser, and no longer its own.


ifranzki:
This will also fix issue https://github.com/opencryptoki/opencryptoki/issues/481

usr/sbin/pkcsslotd/slotmgr.c
```diff
@@ -329,182 +327,213 @@ static int create_pid_file(pid_t pid)
...
-        else
-            d->sinfo_struct.version = vers.major << 16 | vers.minor;
-        break;
-    default:
-        snprintf(d->errbuf, sizeof(d->errbuf),
-                 "Unknown string-valued keyword detected: \"%s\"",
-                 keyword_token_to_str(tok));
-        res = 1;
-        break;
+    ret = parse_configlib_file(file, &config, config_parse_error, 0);
```
juergenchrist, line 482: Not sure if we want to do that, but seeing that the configuration contains paths to libraries that we dlopen, should we check file permissions on the config file?  If it is world-writable, we should probably not use it.

usr/sbin/pkcsslotd/slotmgr.c
```diff
@@ -329,182 +327,213 @@ static int create_pid_file(pid_t pid)
...
+            DbgLog(DL0, "Config file version: '%s'\n",
+                   confignode_to_fileversion(c)->base.key);
+            continue;
+        }
 
-static void slotmgr_disab_event_supp(void *private)
-{
-    UNUSED(private);
+        if (confignode_hastype(c, CT_BARECONST)) {
+            if (strcmp(confignode_to_bareconst(c)->base.key,
```
juergenchrist, line 500: You could simply use `c->key` here since `&(confignode_to_bareconst(c)->base) == c`.

usr/sbin/pkcsslotd/slotmgr.c
```diff
@@ -329,182 +327,213 @@ static int create_pid_file(pid_t pid)
...
+            continue;
+        }
+
+        if (strcmp(c->key, "tokversion") == 0 &&
+            confignode_getversion(c, &sinfo[slot_no].version) == 0)
+            continue;
+
+        ErrLog("Error parsing config file '%s': unexpected token '%s' "
+               "at line %d: \n", config_file, c->key, c->line);
+        return 1;
```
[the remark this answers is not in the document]
ifranzki: Maybe I don't understand your point, but I would state that my code also finds any unused node with just one pass.  If it finds a known node, it processes it and then continues the loop. Only if we reach the end, then we must have an unknown node and report it as error. So how would my code ignore any nodes? 

The only difference I see is that my version reports only one unknown node and then stops, while your version reports all unknown nodes at once.

usr/sbin/pkcsslotd/slotmgr.c
ifranzki: well, we didn't check that when using the old parser either....
Nevertheless, it might certainly be a good ting to add such a check.  Could it have any (unwanted) negative impact on existing installations?

usr/sbin/pkcsslotd/slotmgr.c
ifranzki: yes I know, but to be totally implementation independent, I decided to do it that way :-)

usr/sbin/pkcsslotd/slotmgr.c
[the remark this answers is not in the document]
juergenchrist: My bad.  Too quick in going over the (poorly diffed) code.

usr/sbin/pkcsslotd/slotmgr.c
juergenchrist: Depending on the modes we enforce, yes.  Assume we enforce that only root has write permissions but the distribution does not install it that way.

usr/sbin/pkcstok_migrate/pkcstok_migrate.c
```diff
@@ -2180,16 +2042,62 @@ static struct parsefuncs parseupdatefuncs = {
...
+                    src_file);
+        ret = CKR_FUNCTION_FAILED;
+        goto done;
+    }
+
+    c = confignode_find(slot->value, "tokversion");
+    if (c != NULL) {
+        /* modify existing tokversion */
+        if (confignode_hastype(c, CT_VERSIONVAL)) {
+            confignode_to_versionval(c)->value = 0x3000c;
```
jschmidb, line 2077: There is already a constant for this value 0x000c: TOKVERSION_312

usr/sbin/pkcsslotd/slotmgr.c
ifranzki: I have added a 4th commit to check if the file is world writable:
pkcsslotd: 
- pkcsslotd: Check for opencryptoki.conf file permissions

It does not check for specific owner or group, as we don't know how this file is installed by distributors.

usr/sbin/pkcstok_migrate/pkcstok_migrate.c
ifranzki: ah yes, sure. Will change it to use this.

usr/lib/icsf_stdll/icsf_specific.c
```diff
@@ -382,6 +384,179 @@ CK_RV icsftok_init(STDLL_TokData_t * tokdata, CK_SLOT_ID slot_id,
...
+        TRACE_DEVEL("Config node: '%s' type: %u line: %u\n",
+                    c->key, c->type, c->line);
+
+        str = confignode_getstr(c);
+        if (str != NULL) {
+            for (k = 0; k < refs_len; k++) {
+                if (!strcasecmp(refs[k].key, c->key)) {
+                    strncpy(refs[k].addr, str, refs[i].len);
+                    refs[k].addr[refs[k].len - 1] = '\0';
+                    goto found;
```
juergenchrist, line 452: Why not simply labeling the `confignode_foreach` and `continue <label>`?

usr/sbin/pkcsicsf/pkcsicsf.c
```diff
@@ -136,46 +138,27 @@ int get_pin(char **pin, size_t *pinlen)
...
 
-    /* bump up the slot number, since this will be next new slot entry.
-     * if it was an empty file or a file with no slot entries,
-     * then next new slot entry will be 0.
-     */
-
-    fclose(fp);
-
-    return ++found;
+    return ++slot_id;
```
juergenchrist, line 158: I still think this is dangerous.  Even if I only have one slot, but this slot has the maximum number, then this tool will produce an invalid configuration.  Personally, I think we should go for a two-pass here: first pass mark in a 1024-byte array all slots that are actually used; second pass search in the array for an unmarked slot.

usr/sbin/pkcsicsf/pkcsicsf.c
```diff
@@ -191,84 +174,149 @@ int remove_file(char *filename)
...
 {
+    struct group *grp;
+
+    // Set absolute permissions or rw-rw----
+    fchmod(file, S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP);
+
+    grp = getgrnam("pkcs11"); // Obtain the group id
+    if (grp) {
+        // set ownership to root, and pkcs11 group
+        if (fchown(file, getuid(), grp->gr_gid) != 0) {
```
juergenchrist, line 187: Why `getuid` and not `0`?

usr/sbin/pkcstok_migrate/pkcstok_migrate.c
```diff
@@ -2180,16 +2043,62 @@ static struct parsefuncs parseupdatefuncs = {
...
+    }
+
+    c = confignode_find(slot->value, "tokversion");
+    if (c != NULL) {
+        /* modify existing tokversion */
+        if (confignode_hastype(c, CT_VERSIONVAL)) {
+            confignode_to_versionval(c)->value = TOKVERSION_312;
+        } else if (confignode_hastype(c, CT_STRINGVAL)) {
+            free(confignode_to_stringval(c)->value);
+            confignode_to_stringval(c)->value = strdup(TOKVERSION_312_STRING);
```
juergenchrist, line 2081: Maybe we should check the result of `strdup`?

juergenchrist (approved):

usr/lib/icsf_stdll/icsf_specific.c
ifranzki: ah yes... sometimes things can be so much easier :-)

usr/sbin/pkcsicsf/pkcsicsf.c
ifranzki: well, I guess the comment is wrong, should be '_set ownership to **current user**, and pkcs11 group_'.
This is actually a copy from pkcs_utils.c, but it would make too much trouble to include that module directly due to other unresolved externals, thus the copied routine.....

usr/sbin/pkcsicsf/pkcsicsf.c
ifranzki: Well, the problem is that the calling routine config_add_slotinfo() uses the value returned by get_start_slot() as a start value, and may increase it for further slots it adds. So even if we find a free slot id, we can't guarantee that subsequent slots are unused. So I guess the logic must be changed to let config_add_slotinfo() call get_start_slot() every time to get the next free slot-id (instead of simply incrementing it). I will then rename get_start_slot() to get_free_slot()....

juergenchrist: In that case I would suggest to compute the bitmap in this function and write a function to search the first unset bit in the map for every slot that gets added.  Yes, quite a change to the old version, but it is the closest thing to correct I can think of.

usr/lib/icsf_stdll/icsf_specific.c
ifranzki: ha, well, not really so easy....
C does not have a labeled loop/continue.
So I guess I have to leave it as is.

usr/sbin/pkcstok_migrate/pkcstok_migrate.c
ifranzki: fixed

usr/sbin/pkcsicsf/pkcsicsf.c
ifranzki: no not necessarily, but the user is part of the pkce11 group. 

However, it is certainly a good question who should be allowed to modify opencryptoki.conf.  Currently its this tool (pkcsicsf) and pkcstok_migrate that programmatically modifies opencryptoki.conf. Currently neither of it is limited to be run by just root. 
It also depends on how the distributions install opencryptoki.conf....

Any idea how to handle this better?

ifranzki: OK, changed pkcsicsf as well as pkcstok_migrate to only be usable as root. Also when updating opencryptoki.conf it is then owned by root:root with file mode 644.

p-steuer (approved):

PR #488, opened 2021-11-03
EP11: Add support for CKM_GENERIC_SECRET_KEY_GEN

ifranzki:
For whatever reason, CKM_GENERIC_SECRET_KEY_GEN was excluded from the list of supported mechanisms of the EP11 token.

### DIFF
--- a/usr/lib/ep11_stdll/ep11_specific.c
+++ b/usr/lib/ep11_stdll/ep11_specific.c
@@ -1547,9 +1547,9 @@ static CK_RV check_key_attributes(STDLL_TokData_t * tokdata,
     CK_ULONG check_types_sec_sensitive[] =
         { CKA_ENCRYPT, CKA_DECRYPT, CKA_WRAP, CKA_UNWRAP, CKA_SENSITIVE };
     CK_ULONG check_types_gen_sec[] =
-        { CKA_SIGN, CKA_VERIFY, CKA_ENCRYPT, CKA_DECRYPT };
+        { CKA_SIGN, CKA_VERIFY };
     CK_ULONG check_types_gen_sec_sensitive[] =
-        { CKA_SIGN, CKA_VERIFY, CKA_ENCRYPT, CKA_DECRYPT, CKA_SENSITIVE };
+        { CKA_SIGN, CKA_VERIFY, CKA_SENSITIVE };
     CK_ULONG check_types_derive[] = { CKA_DERIVE };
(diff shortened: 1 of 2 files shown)


usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -1547,9 +1547,9 @@ static CK_RV check_key_attributes(STDLL_TokData_t * tokdata,
     CK_ULONG check_types_sec_sensitive[] =
         { CKA_ENCRYPT, CKA_DECRYPT, CKA_WRAP, CKA_UNWRAP, CKA_SENSITIVE };
     CK_ULONG check_types_gen_sec[] =
-        { CKA_SIGN, CKA_VERIFY, CKA_ENCRYPT, CKA_DECRYPT };
```
juergenchrist, line 1550: Why don't we need CKA_ENCRYPT and CKA_DECRYPT here?  Shouldn't they be in here since a generic secret key should have them?

ifranzki: No, generic secret keys can not be used for encrypt/decrypt. EP11 seemed to ignore CKA_ENCRYPT and CKA_DECRYPT for derive and unwrap operations, but it fails if it is specified for key generation. And it does not make sense to set CKA_ENCRYPT and CKA_DECRYPT to TRUE, because generic secret keys can not be used for encryption/decryption.
PKCS#11 Standard: 
_2.6.2 Generic secret key objects
Generic secret key objects (object class CKO_SECRET_KEY, key type CKK_GENERIC_SECRET) hold generic secret keys. **These keys do not support encryption or decryption;** however, other keys can be derived from them and they can be used in HMAC operations._

juergenchrist: In that case, do we need to ensure they are not present at key gen?

juergenchrist (approved):

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: Well, I would just let it fail in the EP11 host library. If we would check that those are not present, or if present that they are FALSE, and return an error if they are present and TRUE, what would be the benefit? At the end it fails that way or the other way.... 

I don't read the PKCS#11 standard in a way that it forbids generic secret keys to have those attributes set to TRUE, its just that there is no mechanism that uses generic secret keys to perform encrypt/decrypt operations.

PR #492, opened 2021-11-17
TESTCASE: tok2tok_transport: Enhance error handling

ifranzki:
The same that was fixed with commit 022f04d9902d2d621158f1acefc0c7268b95ed6b for EC keys can also happen with RSA keys. Lets skip the testcase for RSA keys the same way as for EC keys.

FYI: @chrisRund


juergenchrist (approved):
lgtm

p-steuer (approved):

PR #496, opened 2021-12-07
libica version 4 support and a minor fix

ifranzki:
Now that libica version 4 is available, we need to adjust the ICA and EP11 tokens to support libica version 4 (libica.so.4), but fallback to version 3 (libica.so.3) if version 4 is not available. That way we don't have a requirement on a specific libica version.

### DIFF
--- a/usr/lib/ep11_stdll/ep11_specific.c
+++ b/usr/lib/ep11_stdll/ep11_specific.c
@@ -68,7 +68,8 @@
 #define EP11SHAREDLIB_V2 "libep11.so.2"
 #define EP11SHAREDLIB_V1 "libep11.so.1"
 #define EP11SHAREDLIB "libep11.so"
-#define ICASHAREDLIB  "libica.so.3"
+#define ICASHAREDLIB_V4  "libica.so.4"
+#define ICASHAREDLIB_V3  "libica.so.3"
 
 CK_RV ep11tok_get_mechanism_list(STDLL_TokData_t * tokdata,
                                  CK_MECHANISM_TYPE_PTR mlist,
@@ -2044,9 +2045,9 @@ static CK_RV make_wrapblob(STDLL_TokData_t * tokdata, CK_ATTRIBUTE * tmpl_in,
 }
 
 #ifdef EP11_HSMSIM
-#define DLOPEN_FLAGS        RTLD_GLOBAL | RTLD_NOW | RTLD_DEEPBIND
+#define DLOPEN_FLAGS        RTLD_NOW | RTLD_DEEPBIND
 #else
-#define DLOPEN_FLAGS        RTLD_GLOBAL | RTLD_NOW
+#define DLOPEN_FLAGS        RTLD_NOW
 #endif
 
 static void *ep11_load_host_lib()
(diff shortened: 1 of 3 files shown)


usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -221,9 +222,13 @@ static CK_RV load_libica(void)
...
     /* Load libica */
-    ibmca_dso = dlopen(LIBICA_SHARED_LIB, RTLD_NOW);
+    ibmca_dso = dlopen(LIBICA_SHARED_LIB_V4, RTLD_NOW);
+    if (ibmca_dso == NULL)
+        ibmca_dso = dlopen(LIBICA_SHARED_LIB_V3, RTLD_NOW);
+
     if (ibmca_dso == NULL) {
-        TRACE_ERROR("%s: dlopen(%s) failed\n", __func__, LIBICA_SHARED_LIB);
+        TRACE_ERROR("%s: dlopen(%s or %s) failed\n", __func__,
+                    LIBICA_SHARED_LIB_V4, LIBICA_SHARED_LIB_V3);
```
jschmidb, line 231: Looks good. However, when we are here, both, loading libica 3 and 4 failed, so should we write: 
`TRACE_ERROR("%s: dlopen(%s and %s) failed\n", __func__,`   ?

ifranzki: Well, 'and' sounds like we need to be able to load both.... But one of it is enough, that's why I chose 'or'.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -2209,12 +2210,19 @@ static CK_RV ep11tok_load_libica(STDLL_TokData_t *tokdata)
         return CKR_OK;
 
     if (strcmp(ep11_data->digest_libica_path, "") == 0) {
-        strcpy(ep11_data->digest_libica_path, ICASHAREDLIB);
+        strcpy(ep11_data->digest_libica_path, ICASHAREDLIB_V4);
         default_libica = 1;
-    }
-
+        libica->library = dlopen(ep11_data->digest_libica_path,
+                                 RTLD_GLOBAL | RTLD_NOW);
```
juergenchrist, line 2216: Why here `RTLD_GLOBAL` (in EP11), but not later in ICA?  Is this required/on purpose?

usr/lib/ica_s390_stdll/ica_specific.c
juergenchrist, line 231: Maybe mingle the output of `dlerror()` somewhere in there?  Might help to diagnose what is actually wrong.

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: No I guess that's not on purpose. Lets remove RTLD_GLOBAL from all dlopen's then....

usr/lib/ica_s390_stdll/ica_specific.c
ifranzki: good point. fixed.

juergenchrist (approved):
lgtm

Issue #498, opened 2021-12-14
p11sak errors out if /etc/opencryptoki/p11sak_defined_attrs.conf is missing


kkaarreell:
opencryptoki-3.17.0-3.el9

p11sak errors out if /etc/opencryptoki/p11sak_defined_attrs.conf is missing.
We are shipping the config with in the package but since p11sak works fine even with this config file being empty it seems that its presence should be optional and p11sak tool should better handle its absence.

Steps to reproduce:
1. rm /etc/opencryptoki/p11sak_defined_attrs.conf
2. initialize sw token
3. p11sak generate-key aes 128 --slot 3 --pin 01234567 --label aes_key --attr ED
Failed to load /etc/opencryptoki/p11sak_defined_attrs.conf: No such file or directory
4. p11sak list-key aes --slot 3 --pin 01234567
Failed to load /etc/opencryptoki/p11sak_defined_attrs.conf: No such file or directory

Now do
# touch /etc/opencryptoki/p11sak_defined_attrs.conf
and repeate steps above.. it works well.

There is also another thing I have noticed regarding /etc/opencryptoki/p11sak_defined_attrs.conf.
If it contains one definition twice then also the respective attribute is listed twice.
E.g.
```
attribute {
   name = CKA_PRIME_1
   id = 0x00000124
   type = CK_BYTE
}

attribute {
   name = CKA_PRIME_1
   id = 0x00000124
   type = CK_BYTE
}
```
While this may be expected I assume that it would be better to ignore the redundant definition and just print a warning.

PR #499, opened 2021-12-14
Fix p11sak config file handling

jschmidb:
Fixes issue #498

### DIFF
--- a/usr/sbin/p11sak/p11sak.c
+++ b/usr/sbin/p11sak/p11sak.c
@@ -19,6 +19,7 @@
 #include "cfgparser.h"
 #include "configuration.h"
 #include <ctype.h>
+#include <linux/limits.h>
 
 #include <unistd.h>
 #include <sys/types.h>
@@ -2773,34 +2774,31 @@ static CK_RV close_session(CK_SESSION_HANDLE session)
 }
 
 static CK_RV parse_file() {
-    FILE *fp;
+    FILE *fp = NULL;
     char *file_loc = getenv("P11SAK_DEFAULT_CONF_FILE");
-    char *buf;
+    char pathname[PATH_MAX];
     struct passwd *pw;
 
     // open and parse file
     if (file_loc != NULL) {
         if ((fp = fopen(file_loc, "r")) == NULL) {
-            fprintf(stderr, "Failed to load %s: %s\n", file_loc, strerror(errno));
+            fprintf(stderr, "Cannot load config file from env variable P11SAK_DEFAULT_CONF_FILE(%s): %s\n",
+                    file_loc, strerror(errno));
+            fprintf(stderr, "Printing of custom attributes not available.\n");
             return CKR_ARGUMENTS_BAD;
         }
     } else {
         pw = getpwuid(getuid());
-        if (pw == NULL) {
-            perror("Could not find home directory of current user!");
-            return CKR_CANCEL;
+        if (pw != NULL) {
+            snprintf(pathname, sizeof(pathname), "%s/.p11sak_defined_attrs.conf", pw->pw_dir);
+            fp = fopen(pathname, "r");
         }
-        if (asprintf(&buf, "%s/.p11sak_defined_attrs.conf", pw->pw_dir) == -1) {
-            perror("Could not allocate memory for home directory path!");
-            return CKR_HOST_MEMORY;
-        }
-        fp = fopen(buf, "r");
-        free(buf);
-
         if (!fp) {
             file_loc = P11SAK_DEFAULT_CONF_FILE;
             if ((fp = fopen(file_loc, "r")) == NULL) {
-                fprintf(stderr, "Failed to load %s: %s\n", file_loc, strerror(errno));
+                fprintf(stderr, "Cannot load config file from default location %s: %s\n",
+                        file_loc, strerror(errno));
+                fprintf(stderr, "Printing of custom attributes not available.\n");
                 return CKR_ARGUMENTS_BAD;
             }
         } 
@@ -2892,7 +2890,7 @@ int main(int argc, char *argv[])
 
     /* Parse p11sak_defined_attrs.conf */
     rc = parse_file();
-    if (rc != CKR_OK) 
+    if (rc == CKR_DATA_INVALID)
         goto done;
 
     /* Execute command */
--- a/usr/sbin/p11sak/p11sak.c
+++ b/usr/sbin/p11sak/p11sak.c
@@ -1143,7 +1143,7 @@ static CK_ATTRIBUTE_TYPE col2type(int col)
 /**
  *  Print in p11sak_defined_attrs.conf defined attributes in long format
  */
-static CK_RV print_vendor(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE hkey, int long_print) 
+static CK_RV print_custom_attrs(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE hkey, int long_print) 
 {
     CK_RV rc = CKR_OK;
     int f;
@@ -1376,7 +1376,7 @@ static CK_RV sec_key_print_attributes(CK_SESSION_HANDLE session,
                         CK_BBOOL2a(*(CK_BBOOL*) bool_tmplt[i].pValue));
             }
         }
-        rc = print_vendor(session, hkey, long_print);
+        rc = print_custom_attrs(session, hkey, long_print);
         if (rc != CKR_OK) {
             fprintf(stderr, "Attribute retrieval failed (error code 0x%lX: %s)\n", rc,
                 p11_get_ckr(rc));
@@ -1386,7 +1386,7 @@ static CK_RV sec_key_print_attributes(CK_SESSION_HANDLE session,
         printf(" |");
         for (i = 2; i < KEY_MAX_BOOL_ATTR_COUNT; i++)
             short_print(i, bool_tmplt, kt_SECRET);
-        rc = print_vendor(session, hkey, long_print);
+        rc = print_custom_attrs(session, hkey, long_print);
         if (rc != CKR_OK) {
             fprintf(stderr, "Attribute retrieval failed (error code 0x%lX: %s)\n", rc,
                 p11_get_ckr(rc));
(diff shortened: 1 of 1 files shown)


usr/sbin/p11sak/p11sak.c
```diff
@@ -2781,7 +2781,8 @@ static CK_RV parse_file() {
     // open and parse file
     if (file_loc != NULL) {
         if ((fp = fopen(file_loc, "r")) == NULL) {
-            fprintf(stderr, "Failed to load %s: %s\n", file_loc, strerror(errno));
+            fprintf(stderr, "Cannot load config file from env variable %s: %s\n", file_loc, strerror(errno));
+            fprintf(stderr, "Printing of vendor attributes not available.\n");
```
ifranzki, line 2856: I would not call this 'vendor' attributes, but maybe 'custom' attributes? Attributes like CKA_MODULUS_BITS and CKA_MODULUS are in fact no vendor attributes, just attributes that are normally not printed....

usr/sbin/p11sak/p11sak.c
```diff
@@ -2800,7 +2801,8 @@ static CK_RV parse_file() {
         if (!fp) {
             file_loc = P11SAK_DEFAULT_CONF_FILE;
             if ((fp = fopen(file_loc, "r")) == NULL) {
-                fprintf(stderr, "Failed to load %s: %s\n", file_loc, strerror(errno));
+                fprintf(stderr, "Cannot load config file from default location %s: %s\n", file_loc, strerror(errno));
+                fprintf(stderr, "Printing of vendor attributes not available.\n");
```
ifranzki, line 2876: same here (and also in commit text)

usr/sbin/p11sak/p11sak.c
```diff
@@ -1185,101 +1237,120 @@ static CK_RV print_vendor(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE hkey, int
...
     }
     if (!long_print)
         printf(" %s ", def_attr ? "0" : "-");
     
-    return CKR_OK;
+    rc = CKR_OK;
+
+done:
+    if (duplicate_attrs_in_conf)
+        fprintf(stderr, "Warning: There are duplicate vendor attr definitions in config file, please check.\n");
```
ifranzki, line 1348: same here. Please don't call this 'vendor' attributes. Maybe also rename the functions you added to something else than 'vendor'.

ifranzki:
Can you please change the commit titles from `[P11SAK] ... ` to `P11SAK: ....` ?
I guess Juergen accidentally introduced the [...] style once, but I would like to stick with the original style.

usr/sbin/p11sak/p11sak.c
jschmidb: OK, done.

usr/sbin/p11sak/p11sak.c
jschmidb: Done.

usr/sbin/p11sak/p11sak.c
jschmidb: Done.

jschmidb:
> Can you please change the commit titles from [P11SAK] ... to P11SAK: .... ?

Done.

usr/sbin/p11sak/p11sak.c
```diff
@@ -1182,104 +1234,123 @@ static CK_RV print_vendor(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE hkey, int
...
     if (!long_print)
         printf(" %s ", def_attr ? "0" : "-");
-    
-    return CKR_OK;
+
+    rc = CKR_OK;
+
+done:
+    if (duplicate_attrs_in_conf)
+        fprintf(stderr, "Warning: There are duplicate custom attr definitions in config file, please check.\n");
```
juergenchrist, line 1348: Why do we care about this?  The config file is user configuration.  If the user wants that attribute twice we should just silently print it...

usr/sbin/p11sak/p11sak.c
```diff
@@ -1182,104 +1234,123 @@ static CK_RV print_vendor(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE hkey, int
...
-                    rc = funcs->C_GetAttributeValue(session, hkey, &temp, 1);
-                    if (rc == CKR_OK) {
-                        if (temp.ulValueLen != sizeof(CK_ULONG)) {
-                            fprintf(stderr, " Error in retrieving Attribute %s: %lu\n",
-                                    confignode_to_bareval(name)->value, temp.ulValueLen);
-                        } else {
-                            if (long_print) {
-                                printf("          %s: %lu (0x%lX)\n", confignode_to_bareval(name)->value, 
-                                        *(CK_ULONG*) temp.pValue, *(CK_ULONG*) temp.pValue);    
+                if (!attr_already_printed(hex, attrs, attrs_len)) {
```
juergenchrist, line 1240: So what do I have to do to get the same attribute printed twice?  I do not see any statement that I can only print the value once.  Especially since I can use different names for the same hex value I doubt this check is a good idea.

usr/sbin/p11sak/p11sak.c
```diff
@@ -2781,7 +2852,8 @@ static CK_RV parse_file() {
     // open and parse file
     if (file_loc != NULL) {
         if ((fp = fopen(file_loc, "r")) == NULL) {
-            fprintf(stderr, "Failed to load %s: %s\n", file_loc, strerror(errno));
+            fprintf(stderr, "Cannot load config file from env variable %s: %s\n", file_loc, strerror(errno));
```
juergenchrist, line 2855: Very misleading error message: "Cannot load config file from env variable /does/not/exist: No such file or directory".  Did you mean `P11SAK_DEFAULT_CONF_FILE` instead of the first `%s` (or even better before the `%s` and add the value in brackets).

usr/sbin/p11sak/p11sak.c
```diff
@@ -2892,8 +2965,14 @@ int main(int argc, char *argv[])
 
     /* Parse p11sak_defined_attrs.conf */
     rc = parse_file();
-    if (rc != CKR_OK) 
+    switch (rc) {
```
juergenchrist, line 2968: Why not simply `if (rc == CKR_DATA_INVALID)`?  That is the only remaining exit code that leads to a failure.

usr/sbin/p11sak/p11sak.c
jschmidb: This was a suggestion in issue #498

usr/sbin/p11sak/p11sak.c
jschmidb: Hmm ... different names for the same hex value. Why would a customer do this? With the old code I can reproduce duplicate printing of the same attribute (hex value) and I agree with Karel that this doesn't look good. However, we could leave it up to the customer to fix the config file instead of ignoring duplicate attrs in our code.

usr/sbin/p11sak/p11sak.c
jschmidb: Argh ... good point: yes I should print the name of the env variable instead of the first %s, and as you suggest, also print the actual value

usr/sbin/p11sak/p11sak.c
jschmidb: parse_file() may also return CKR_CANCEL or CKR_HOST_MEMORY

usr/sbin/p11sak/p11sak.c
jschmidb: Done.

usr/sbin/p11sak/p11sak.c
juergenchrist: Point is: this is a user configuration file.  We do not know why a user added the attribute twice.  Just because it "doesn't look good" does not mean that it is wrong.  If there is no real reason why a user should not be allowed to do it, why add that undocumented constraint?

usr/sbin/p11sak/p11sak.c
juergenchrist: Yeah, I would get rid of these two.  The `CKR_CANCEL` case is another example where this tool cannot be used inside sandboxes or user namespaces, and the `CKR_HOST_MEMORY` can simply be replaced by a buffer local to this function of some size (and a bailout if the path does not fit within this buffer).

jschmidb: OK, now using a local buffer of PATH_MAX bytes and snprintf. CKR_CANCEL does not lead to an error anymore.

usr/sbin/p11sak/p11sak.c
jschmidb: OK, here I would leave the decision up to Ingo: I can completely remove the commit that causes duplicate conf entries to be multiply written. Then we would just fix the config file handling.

finn-callies: Maybe it would make sense to print the id the user defines in the config file too and let him use it multiple times. When it is on purpose he can do it and if it was an accident, because he didn't know it was already defined, he can tell by the printed id, even when he defined different names for the same id... to identify an id which was used multiple times maybe it would be good to sort the attributes in the long output by the id, so that the id's with multiple definitions are listed directly under each other. This would not be easy though with the current implementation, because an attribute is printed instantly after parsing, without knowing which attribute follows....

ifranzki: Actually I also asked myself if rejecting a duplicate definition is really a good idea. I have not finally may up my mind though ....

Yes, it looks bad if you see duplicate definitions, but well, its a user config file, if the user does not like the result of what he configured, then he is free to change it :-) 

For simplicity, I guess I would tend to just not check for duplicate definitions.

usr/sbin/p11sak/p11sak.c
juergenchrist: That is standard Linux procedure.  Take bashrc as an example.  You have `/etc/bashrc` and `~/.bashrc` since you do not want to see all these user-config files per default when issuing `ls`.

usr/sbin/p11sak/p11sak.c
jschmidb: OK agreed: so we decide to just fix the config file handling. I removed the duplicate entries commit together with the whitespace fixes. We may do some more general housekeeping on p11sak later.

usr/sbin/p11sak/p11sak.c
```diff
@@ -2790,17 +2793,14 @@ static CK_RV parse_file() {
             perror("Could not find home directory of current user!");
             return CKR_CANCEL;
```
juergenchrist, line 2794: This should not be an error.  Just fall back    to the default location.

jschmidb: OK, just pushed the change ... let me know if this is what you think ...

juergenchrist (approved):
lgtm

ifranzki (approved):

Issue #467, opened 2021-10-07
arbitrary code execution warning


james-callahan:
In the man page (https://github.com/opencryptoki/opencryptoki/blame/025022f5b89302891272c0178cf9e5314a5741be/man/man7/opencryptoki.7.in#L19-L26) there is a warning that members of the `pkcs11` group will be able to run arbitrary code as each other.

How does this become the case? As mentioned in the same commit that introduced the warning (https://github.com/opencryptoki/opencryptoki/commit/7744b6224e80848596ac80a07745c7a588eef2a0), there is a shared memory segment with `pkcsslotd`: but how does that result in arbitrary code execution rather than e.g. just being able to corrupt another person's token?

Furthermore, what can be done to fix or mitigate the security hole?
  - Could there be unique shared memory segments per user/client?
  - Is it too hard to replace it with communicating over a unix socket?
  - Would proxying through `p11-kit` make it safe?

p-steuer:
Sorry for the late reply, @james-callahan . Investigating this one took us some time.

Before the 3686bd27f890e95f47bc23d4961a1de4d87013da patch series from 2013, the location of some shared libraries to be loaded were stored in the shared memory segment. Users in the pkcs11 group could write to the shared memory and thus replace those paths by their own shared libraries, which enabled arbitrary code execution e.g., by placing malicious code it in the library constructor.

Since the above commit, the location of those shared libraries is not read from the shared memory segment anymore since that was replaced by socket communication.

Therefore, we will change the wording of the warning you mentioned.

However, all users in the pkcs11 group may still write and read whatever is stored in the shared memory. In case this is still a problem in your setup, you would have to think about providing each user its own opencryptoki instance (e.g. in a container) or, as you suggested, try to prooxy via p11-kit. In the latter case, the shared memory segment resides on the server side and clients/users wont be able to access it.

In the long run, our plans are to get rid of the whole shared memory segment. The most likely solution would be to use unix socket communication instead.

james-callahan:
Thanks for your reply.

> However, all users in the pkcs11 group may still write and read whatever is stored in the shared memory.

In particular I would hope that any future solution would require the unlock PIN to gain useful access to the shared memory segment. If I (as a unix user) am in the pkcs11 group, I wouldn't want every program I run to have access to the token.

[...]

PR #503, opened 2022-01-03
MAN: Update security note in opencryptoki man page

ifranzki:
The note about arbitrary code execution is no longer true since commit 3686bd27f890e95f47bc23d4961a1de4d87013da from 2013. This commit was included since Opencryptoki version 3.0.

Before that commit, the location of some shared libraries to be loaded were stored in the shared memory segment. Users in the pkcs11 group could write to the shared memory and thus replace those paths by their own shared libraries, which enabled arbitrary code execution e.g., by placing malicious code it in the library constructor.

Since the above commit, the location of those shared libraries is not read from the shared memory segment anymore since that was replaced by socket communication.

This fixes https://github.com/opencryptoki/opencryptoki/issues/467


juergenchrist (approved):
lgtm

hfreude (approved):
Fine with me.
Long term we could also rework the sentence "Group members can block other openCryptoki users from accessing PKCS#11 tokens". I assume this means a denial of service, but the sentence is very fuzzy and should be more specific.

PR #504, opened 2022-01-13
P11SAK: CKA_PRIVATE now set by OCK

finn-callies:
This is a small fix addressing a problem, where p11sak sets the CKA_PRIVATE attribute for every generated key to true. CKA_PRIVATE can now be set, alongside the other standard attributes, with the --attr option and the letter P.

### DIFF
--- a/man/man1/p11sak.1.in
+++ b/man/man1/p11sak.1.in
@@ -84,7 +84,7 @@ option will show the arguments and options available.
 .B \-\-label
 .IR LABEL
 .B \-\-attr
-.IR [M R L S E D G V W U A X N T]
+.IR [P M R L S E D G V W U A X N T]
 .B \-\-help | \-h
 .PP
 Use the
@@ -105,7 +105,7 @@ option allows the user to set the
 .IR LABEL
 attribute of the key and
 .B \-\-attr
-.IR [M R L S E D G V W U A X N T]
+.IR [P M R L S E D G V W U A X N T]
 can be used to set the binary attributes of the key (see below for detailed description of the attributes).
 .
 .PP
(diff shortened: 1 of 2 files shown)


jschmidb:
The code looks good to me. Just one thought about defaults: up to now the default for CKA_PRIVATE was true. Now, as p11sak just omits the attribute, it is left up to opencryptoki whether to make a key private or not. I found nothing in the standard about a specific default for CKA_PRIVATE. When looking at opencryptoki, it looks like the attr is also just left out as well. But I saw that e.g. z/OS ICSF PKCS#11 C API sets CKA_PRIVATE to true by default (https://www.ibm.com/docs/en/zos/2.5.0?topic=api-objects-attributes-supported). At the end I would tend to just leave the attr out and let the user explicitly decide whether a key object shall be private. As far as I remember, the GnuTLS p11tool wants a login to display private objects, but can display public objects without a login. In contrast, the p11sak tool always wants the user PIN, so there is no difference.

ifranzki:
> When looking at opencryptoki, it looks like the attr is also just left out as well.

In `template_set_default_common_attributes()` it sets the default to FALSE.

jschmidb:
Good ... I oversaw this.
p.s.: now that not all key objects will be private, should we allow p11sak to display public objects without a login, similar to p11tool?

ifranzki:
> should we allow p11sak to display public objects without a login

Hm, not a bad idea in general, but how would that work? If no pin is specified then list only public objects? But currently, if no pin is specified via option it prompts for the pin....  How to know to not prompt? Maybe an additional option like --public-only ?

Anyway, lets keep this idea on our list (or open a feature request issue for it), but that's not something got tis PR.

jschmidb:
> How to know to not prompt? Maybe an additional option like --public-only ?

The p11tool has a parm --login. If this parm is specified, the tool prompts for a login and can then display private objects.

juergenchrist (approved):

ifranzki (approved):

PR #506, opened 2022-01-14
ICA: Check fur successful pthread_mutex_lock()

ifranzki:
Return an error if we can't obtain the lock for the ICA RNG.


juergenchrist (approved):

jschmidb (approved):
lgtm

PR #510, opened 2022-01-21
ICSF: replace PBKDF implementation by libcrypto call

ifranzki:
The PKBDF2 implementation of the ICSF token is not correct, and thus can not be simply replaced by the OpenSSL implementation of PBKDF2.

Introduce a version field in ICCF's MK_SO and MK_USER files to be able to detect old version files. Old version files do not have a version field, but start with the total-length field right away. The length contained in the total-length field can be distinguished from a valid version value. That way we can still read old version files, but can also transparently support the new version files with the correct PKBDF2 implementation.

When writing ICSF MK_SO or MK_USER files, we always use the new version format. This allows a smooth transition to the new version. Just set new pins for SO and USER and the MK files are updated with the new version format.

Based on the work from Patrick Steuer: https://github.com/opencryptoki/opencryptoki/pull/318

### DIFF
--- a/usr/lib/icsf_stdll/pbkdf.c
+++ b/usr/lib/icsf_stdll/pbkdf.c
@@ -174,7 +174,7 @@ CK_RV get_masterkey(STDLL_TokData_t *tokdata,
 {
     struct stat statbuf;
     FILE *fp;
-    CK_ULONG_32 totallen, datasize, readsize;
+    CK_ULONG_32 totallen, datasize, readsize, version;
     int dkeysize;
     CK_BYTE salt[SALTSIZE];
     CK_BYTE dkey[AES_KEY_SIZE_256];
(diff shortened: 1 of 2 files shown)


juergenchrist (approved):
lgtm

jschmidb:
Two typos in the commit text: change PKBDF2 to PBKDF2

usr/lib/icsf_stdll/pbkdf.c
```diff
@@ -195,13 +195,32 @@ CK_RV get_masterkey(STDLL_TokData_t *tokdata,
...
         fclose(fp);
         TRACE_ERROR("fread failed.\n");
         return CKR_FUNCTION_FAILED;
     }
 
+    /*
+     * Version 1 files did not have the version field, thus the first field read
+     * is the total length. If the first field is not a known version number,
+     * we assume it is an old version file. For an old version file, the total
+     * length is 64 bytes, this is it is much higher than the version number.
```
jschmidb, line 209: this is much higher than the version number ?

usr/lib/icsf_stdll/pbkdf.h
```diff
@@ -23,9 +23,17 @@
 #define PIN_SIZE 80      // samedefine in pkcsconf
```
jschmidb, line 23: same define

jschmidb (approved):
lgtm

PR #513, opened 2022-01-28
EP11: Trace EP11 blob attributes on Keygen, Unwrap, Derive

ifranzki:
To help debugging when the EP11 host lib does not like certain attribute settings.


jschmidb (approved):
lgtm

PR #515, opened 2022-02-21
API: Avoid deadlock when stopping event thread

ifranzki:
Avoid that the event thread writes trace messages while it is
enabled for thread cancellation. This might leave the trace mutex in
the locked state and cause subsequent trace calls to lock forever
(e.g in stop_event_thread() right after canceling the thread).

Disable cancellation right at the beginning of the thread function,
and disable it before calling a trace function or leaving the loop.

Also make sure that the cleanup handler is registered and the
cancellation type is set before initially enabling cancellation.

Signed-off-by: Ingo Franzki <ifranzki@linux.ibm.com>

### DIFF
--- a/usr/lib/api/socket_client.c
+++ b/usr/lib/api/socket_client.c
@@ -284,6 +284,8 @@ static void *event_thread(void *arg)
 
     UNUSED(arg);
 
+    pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, &oldstate);
+
     TRACE_DEVEL("Event thread %lu running\n", pthread_self());
 
     if (anchor->socketfd < 0) {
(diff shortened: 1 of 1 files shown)


juergenchrist (approved):
lgtm

usr/lib/api/socket_client.c
```diff
@@ -303,13 +305,13 @@ static void *event_thread(void *arg)
...
     /* Enable cancellation */
-    pthread_setcancelstate(PTHREAD_CANCEL_ENABLE, &oldstate);
-    pthread_setcanceltype(PTHREAD_CANCEL_DEFERRED, &oldtype);
     cleanup.anchor = anchor;
 #if OPENSSL_VERSION_PREREQ(3, 0)
     cleanup.prev_libctx = prev_libctx;
 #endif
     pthread_cleanup_push(event_thread_cleanup, &cleanup);
+    pthread_setcanceltype(PTHREAD_CANCEL_DEFERRED, &oldtype);
+    pthread_setcancelstate(PTHREAD_CANCEL_ENABLE, &oldstate);
```
jschmidb, line 314: Don't know if this makes a difference: when looking for pthread examples on the Internet, they always first set the cancelstate and then the canceltype. Like it was before in removed lines 306+307. Did you change the order intentionally?

ifranzki: Right, I changed the order intentionally because it also sounds more logical to me to first tell how to cancel it and second allowing to cancel it. I originally had it the other way round because I probably simply copied it from the examples.... If it does not really make a difference, then thats even better.

jschmidb (approved):
lgtm

PR #516, opened 2022-02-25
Validate object name when loading token objects

ifranzki:
The object name within a token object payload must match the file name it was loaded from. For private token objects (CKA_PRIVATE=TRUE) the object name is part of the encrypted payload.

A mismatch may indicate that the token object file has been tampered, for example due to a file rename or replacement. Don't load such objects, since this is most likely not the expected object that is being loaded.


juergenchrist (approved):

jschmidb (approved):
lgtm

PR #517, opened 2022-02-25
Use mkstemp for unique token object name generation

ifranzki:
Do no longer use the pre-calculated next object name in the token's shared memory segment, but use a unique name generated by mkstemp instead. This guarantees unique file names, and the next object name to be used can not be tampered in the shared memory segment.

Existing token object files keep their names, only new files are created via mkstemp. There is no update/migration problem because files with old names and new names can live together without a problem.

The field used to contain the next token object name is kept in struct TOKEN_DATA but renamed to 'unused'. Removing this field would cause update/migration problems.

### DIFF
--- a/usr/lib/common/h_extern.h
+++ b/usr/lib/common/h_extern.h
@@ -417,8 +417,6 @@ CK_RV communicate(CK_ULONG cmd_id,
                   CK_BYTE_PTR pOut, CK_ULONG out_len,
                   CK_BYTE_PTR pIn, CK_ULONG in_len);
 
-CK_RV compute_next_token_obj_name(CK_BYTE *current, CK_BYTE *next);
-
 CK_RV save_token_object(STDLL_TokData_t *tokdata, OBJECT *obj);
 CK_RV save_private_token_object(STDLL_TokData_t *tokdata, OBJECT *obj);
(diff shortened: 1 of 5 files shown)


juergenchrist (approved):

jschmidb (approved):

ifranzki:
With the names generated by mkstemp the names are no longer generated in increasing order. The list of token objects in the SHM were supposed to be sorted (for whatever reason). Functions `object_mgr_sort_priv_shm` and `object_mgr_sort_publ_shm` are a NOP, because the names were always generated in sorted order anyway.

```
CK_RV object_mgr_sort_priv_shm(void)
{
    // for now, we assume the list is sorted by design. this is not unreasonable
    // since new object handles are assigned in increasing order. problems
    // will arise after 36^8 token objects have been created...
    //
    return CKR_OK;
}
```

I currently do not see why the list of objects in the SHM would be required to be sorted at all. 

Function `object_mgr_search_shm_for_obj` does a linear search, so  the order should be irrelevant.  If it would use a binary search, then a sorted list might be required, but I don't see that this is the case anywhere.

Maybe I should clean up the code and remove the NOP-functions `object_mgr_sort_priv_shm` and `object_mgr_sort_publ_shm` to make clear that this is not a sorted list.

PR #518, opened 2022-03-02
Miscellaneous fixes

ifranzki:


usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -8731,6 +8731,7 @@ static const CK_MECHANISM_TYPE ep11_supported_mech_list[] = {
     CKM_SHA_1_HMAC,
     CKM_GENERIC_SECRET_KEY_GEN,
     CKM_IBM_ATTRIBUTEBOUND_WRAP,
+    CKM_GENERIC_SECRET_KEY_GEN,
```
jschmidb, line 8762: It was there, two lines above

ifranzki: Arg, you are right. It was already added by https://github.com/opencryptoki/opencryptoki/commit/945a46aecd3006387e551a76d8d94ea622bddc4a
I guess this was a git-rebase error to add it again. 
SO I'll drop this commit entirely.

usr/lib/common/asn1.c
```diff
@@ -4215,12 +4215,28 @@ CK_RV ber_decode_IBM_DilithiumPrivateKey(CK_BYTE *data,
...
+            TRACE_DEVEL("ber_decode_CHOICE returned invalid option %ld\n",
+                        option);
+            goto cleanup;
+        }
+
+        offset += field_len - len;
+
+        rc = ber_decode_BIT_STRING(buf + offset, &tmp, &len, &field_len);
+        if (rc != CKR_OK) {
+            TRACE_DEVEL("ber_decode_BIT_STRING of (t1) failed\n");
```
jschmidb, line 4254: What happens if option == 0, but there is no (t1) value given? Will field_len be just 0 and the build_attribute below creates a "dummy" attr with a length of 0? Should we probably don't add the attr at all if (t1) is not given?

ifranzki: Well, the same question applies to all the key components, even those that are not optional by design. If their length would be zero we would add a zero length attribute (build_attribute can deal with len = 0).

That would anyway be an invalid encoding of an Dilithium key. 
So: garbage in, garbage out :-)

jschmidb:
OK .... other commits lgtm.

jschmidb (approved):

PR #519, opened 2022-03-08
CCA: Reject the use of twisted brainpool curves

ifranzki:
CCA does not support twisted brainpool curves, so reject them with CKR_CURVE_NOT_SUPPORTED for key generation and key import.

@vkanjur thanks for finding this!


juergenchrist (approved):

vkanjur (approved):

vkanjur:
Approved the changes.

PR #520, opened 2022-03-08
EP11: Dilithium and CPACF-Wrap can be used with hostlib version >= 3.0

ifranzki:
@jschmidb Can you please double check? I would say that both mechs are available with the EP11 host lib version 3.0 already (e.g. 3.0.1), correct?


jschmidb:
3.0.1 definitely yes. That's what I have used when implementing the support. Unfortunately I don't have a 3.0.0 anymore to check. But as far as I know it's in fact even 3.0.0.

ifranzki:
I changed my mind and encode the release and mod-level into the one byte minor field of the CK_VERSION (4 bits release, 4 bits mod-level). I hope that the host lib never has more than 15 releases and/or mod-levels.....

That way, the version checks for Dilithium and CPACF-wrap can stay as is, and will allow them only for host libs version 3.0.1 or later (i.e. > 3.0.0.). 

@jschmidb please take a look again.

jschmidb:
I think this is ok, but why don't we introduce an additional field "patch" in the CK_VERSION struct? This would make the code more readable. What do we exactly get from dll_m_get_xcp_info?

ifranzki:
CK_VERSION is a structure defined by the PKCS#11 standard, we can't change it.
And I didn't want to use a different (own defined) struct here, since this would have hits at many places (I will probably have to backport this fix to several distros....).

ifranzki:
> What do we exactly get from dll_m_get_xcp_info

0x00VVRRMM

jschmidb:
> And I didn't want to use a different (own defined) struct here, since this would have hits at many places

Yes, I was afraid of that .... ok, so we will use CK_VERSION = 0xVVRM ... ok, as you say: the max is now 15 release or patch levels, but I think they will not reach this.

jschmidb (approved):

PR #521, opened 2022-03-09
Remove AUTHORS file

ifranzki:
The AUTHORS file is heavily outdated. So we should either update it or get rid of it.

I don't really see much sense of having the AUTHORS file, since you can always get a list of contributors using `git shortlog --summary --email`.

We use `foreign` in `AM_INIT_AUTOMAKE` in `configure.ac`, so it won't complain about the missing AUTHORS file (GNU projects are required to have it, but `foreign` projects don't).

Alternatively we could produce a proper AUTHORS file using `git shortlog --summary --email`, but would have to keep it up to date more frequently.

Any opinions?


juergenchrist (approved):

vkanjur (approved):

jschmidb (approved):
Yep, makes sense

hfreude (approved):
fine with me

PR #522, opened 2022-03-15
Support pid and user namespaces 

ifranzki:
Ensure that an OCK application running under pid and user namespaces behave correctly. The pkcsslotd must still run in the root-namespace.

Note: I have not been able to successfully run an OCK application under a user namespace so far, at least not with the sample program userns_child_exec from the user_namespaces man page. Permission checks inside C_Initialize always hinders the program from connecting to the pkcsslotd due to incorrect user or group settings of either the sockets, shared memory segments, config files, etc., or of the the user running the program inside the namespace. I wasn't able to find a user and group mapping that would allow it to run, neither when running userns_child_exec as root, nor as regular user....

The second commit corrects some places where getuid (vs geteuid) was used.


ifranzki:
By temporarily commenting out all the permission checks I was able to run an OCK application in a user namespace (plus pid namespace), and it now behaves correctly in regards of the statistics shared memory segment name.

All these checks are not really required from a security point of view, its more like a cross-check detecting inconsistencies before actually attempting to access a resource. At the end the resource's permissions will hinder one from accessing them if a user does not have valid permissions anyway. 

So we could think about removing those cross-checks completely, or maybe only when running in a user namespace (real_uid != geteuid()) ? 

I think the cross-checks are useful (in a non-namespace environment) to easily find out why a user is not allowed to access opencryptoki. I.e. we get useful error messages on syslog (and trace).

Any opinions?

jschmidb (approved):

PR #524, opened 2022-03-21
Testcases cleanup and fix counting

ifranzki:
testcase_new_assertion() increments the number of ran testcases. 
Total number of testcases is ran + skipped. 
So do not call testcase_new_assertion()  before it is sure that the testcase can actually be run, and is not skipped later on. Make sure that all the skip possibilities are handed before testcase_new_assertion(). Once a test is run, it can either fail of pass, but no longer skipped.

### DIFF
--- a/testcases/crypto/abfunc.c
+++ b/testcases/crypto/abfunc.c
@@ -1180,7 +1180,7 @@ int main(int argc, char **argv)
     }
 
     if (is_ep11_token(SLOT_ID)) {
-        testcase_setup(0);
+        testcase_setup();
         rc = do_CheckMechanismInfo();
         if (rc != CKR_OK) {
             // Skip, but don't crash the test executor
--- a/testcases/crypto/ec_func.c
+++ b/testcases/crypto/ec_func.c
@@ -101,8 +101,6 @@
  *      06032B6571
  */
 
-CK_ULONG total_assertions = 65;
-
 typedef struct ec_struct {
     void const *curve;
     CK_ULONG size;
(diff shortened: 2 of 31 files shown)


testcases/crypto/abfunc.c
```diff
@@ -1180,7 +1180,7 @@ int main(int argc, char **argv)
     }
 
     if (is_ep11_token(SLOT_ID)) {
-        testcase_setup(0);
+        testcase_setup();
         rc = do_CheckMechanismInfo();
```
jschmidb, line 1184: Yes, that's true. Are you also removing the arg from the macro definition in regress.h?

testcases/crypto/ec_func.c
```diff
@@ -2392,13 +2388,15 @@ int main(int argc, char **argv)
...
 
-    pkey = CK_TRUE;
-    rv = run_GenerateECCKeyPairSignVerify();
-    rv += run_ImportECCKeyPairSignVerify();
-    rv += run_TransferECCKeyPairSignVerify();
-    rv += run_DeriveECDHKey();
-    rv += run_DeriveECDHKeyKAT();
-    rv += run_ImportSignVerify_Pkey();
+    if (is_ep11_token(SLOT_ID)) {
+        pkey = CK_TRUE;
```
jschmidb, line 2392: Yes, that make sense. But if we check is_ep11_token() here, we can remove the is_ep11_token() check inside the various routines.

testcases/crypto/abfunc.c
ifranzki: sure

testcases/crypto/ec_func.c
ifranzki: yea, I thought about this, but I guess it does not hurt to keep the check there as well. Just in case someone calls these functions with pkey set to true.

jschmidb (approved):
ok lgtm

PR #525, opened 2022-03-22
Support CKA_DERIVE_TEMPLATE attribute

ifranzki:
CKA_DERIVE_TEMPLATE: For deriving keys. The attribute template to match against any keys derived using this derivation key. Any user supplied template is applied after this template as if the object has already been created. The number of attributes in the array is the ulValueLen component of the attribute divided by the size of CK_ATTRIBUTE.


usr/lib/common/key_mgr.c
```diff
@@ -1483,41 +1490,102 @@ CK_RV key_mgr_derive_key(STDLL_TokData_t *tokdata,
...
+                                          base_key_obj->session);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("POLICY VIOLATION: derive key\n");
+        goto done;
+    }
+
+    if (!key_object_is_mechanism_allowed(base_key_obj->template,
+                                         mech->mechanism)) {
+        TRACE_ERROR("Mechanism not allwed per CKA_ALLOWED_MECHANISMS.\n");
+        rc = CKR_MECHANISM_INVALID;
```
jschmidb, line 1514: allowed

jschmidb:
OK, as far as I see, this is the first time CKA_DERIVE_TEMPLATE is used in opencryptoki. Makes sense for EC, but could we use it also for AB and DH?

usr/lib/common/key_mgr.c
ifranzki: fixed. 

Same typo at several other places, fixed those in an additional commit :-)

jschmidb (approved):

PR #526, opened 2022-03-23
TESTCASE: EC: Clarify on EP11 not deriving keys < 80 bits

ifranzki:
EP11 can not derive keys less than 80 bits (10 bytes). This was formerly dependent on control point XCP_CPB_KEYSZ_BELOW80BIT, but this control point is now always OFF.


jschmidb (approved):
lgtm

PR #528, opened 2022-03-28
MAKEFILE: Also uninstall strength.conf and p11sak_defined_attrs.conf

ifranzki:
'make uninstall' should also remove those config files, like the others.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -202,7 +202,11 @@ if ENABLE_SYSTEMD
 		rm -f $(DESTDIR)/usr/lib/tmpfiles.d/opencryptoki.conf; fi
 endif
 endif
+if ENABLE_P11SAK
+	rm -f $(DESTDIR)$(sysconfdir)/opencryptoki/p11sak_defined_attrs.conf || true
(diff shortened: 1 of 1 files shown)


juergenchrist (approved):

jschmidb:
Hmm, yes, this would be consistent with the other config files. But it's as always: uninstall/reinstall will remove any config changes by the user.

ifranzki:
> uninstall/reinstall will remove any config changes by the user.

That right. 

We could certainly decide to NOT remove config files, but if we decide to do so we have to do it consistently for all config files. 

I would assume that 'normal' users get the package from the distro, and thus do not build it on their own and install it via `make install`. So they won't/can't use 'make uninstall' either. So the 'make uninstall' case is mostly for developers. Not sure if they would expect that the config files would stay there...?

jschmidb:
> We could certainly decide to NOT remove config files, but if we decide to do so we have to do it consistently for all config files.

I agree.

jschmidb (approved):

PR #529, opened 2022-03-31
Cleanup OpenSSL context in OpenSSL exit handler

ifranzki:
Usually opencryptoki's own library context is freed during C_Finalize(). However, when an application does not call C_Finalize() before it exits, the cleanup is performed in the library destructor when the library is unloaded (i.e. during exit handlers).

OpenSSL also performs its own cleanup in exit handlers, and it may happen that OpenSSL cleanup is performed before the library destructors are called. This may cause crashes when Opencryptoki's library context has already been freed by OpenSSL cleanup, but the library destructor tries to free it a second time. This causes a double free, and very likely a crash.

Register an OpenSSL cleanup handler to clean up the library context before OpenSSL performs its own cleanup.

While we are at it, make sure pkcsconf calls C_Finalize() at the end of processing, just to be nice :-)

### DIFF
--- a/usr/lib/api/api_interface.c
+++ b/usr/lib/api/api_interface.c
@@ -3008,6 +3008,21 @@ CK_RV C_Initialize(CK_VOID_PTR pVoid)
      * usage breaks this loop.
      */
     ERR_set_mark();
+
+    /* Perform Opencryptoki cleanup when OpenSSL cleanup is run.
+     * Otherwise it might happen that the library destructor is called
+     * after OpenSSL cleanup has already been performed, and this will
+     * cause crashes when trying to free our own OpenSSL library context,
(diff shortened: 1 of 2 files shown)


usr/lib/api/api_interface.c
```diff
@@ -3008,6 +3008,21 @@ CK_RV C_Initialize(CK_VOID_PTR pVoid)
      * usage breaks this loop.
      */
     ERR_set_mark();
+
+    /* Perform Opencryptoki cleanup when OpenSSL cleanup is run.
+     * Otherwise it might happen that the library constructor is called
+     * after OpenSSL cleanup has already been performed, and this will
```
juergenchrist, line 3014: s/constructor/destructor

juergenchrist (approved):

jschmidb (approved):

PR #531, opened 2022-04-08
ICA/EP11: Call ica_cleanup() if available

ifranzki:
libica recently added function ica_cleanup to be called to free internal
OpenSSL 3.0 resources.


jschmidb:
lgtm. 
Just one question: we are now introducing dlclose() for the libica module in token_specific_final. Why do both, CCA and EP11 token, not do a dlclose() on their loaded host libs?

ifranzki:
> Why do both, CCA and EP11 token, not do a dlclose() on their loaded host libs?

good question.... I also wondered why they do not do a dlclose for libica when I worked on this patch....
I guess I will add the dlclose for the host libraries in a second commit.

jschmidb (approved):

juergenchrist (approved):

PR #532, opened 2022-04-11
COMMON: Deep copy the AES-GCM mechanism parameter

ifranzki:
During C_EncryptInit and C_DecryptInit the AES-GCM mechanism parameter CK_GCM_PARAMS must be deep-copied into the operation context. The areas pointed to by fields pIv and pAAD may only be valid during C_EncryptInit/C_DecryptInit, but may be no longer valid for C_EncryptUpdate/C_DecryptUpdate and C_EncryptFinal/C_DecryptFinal. When the context is freed, the GCM parameter is also be deep-freed.

As far as I can tell AES-GCM is the only mechanism that needs this special handling at the moment. Mechanisms that have complex mechanism parameter structures but are used in single shot calls only (e.g. C_DeriveKey, C_WrapKey/C_UnwrapKey do not need a special handling, since its mechanism parameter is not copied into the context. Only mechanisms that are used with Init-Update-Final calls are subject to special handling.

### DIFF
--- a/testcases/crypto/aes_func.c
+++ b/testcases/crypto/aes_func.c
@@ -406,6 +406,48 @@ testcase_cleanup:
     return rc;
 }
 
+CK_RV alloc_gcm_param(CK_GCM_PARAMS *gcm_param, CK_BYTE *pIV, CK_ULONG ulIVLen,
+                      CK_BYTE *pAAD, CK_ULONG ulAADLen)
+{
+    gcm_param->pIv = malloc(ulIVLen);
+    if (gcm_param->pIv == NULL)
+        return CKR_HOST_MEMORY;
+    gcm_param->ulIvLen = ulIVLen;
+    memcpy(gcm_param->pIv, pIV, ulIVLen);
+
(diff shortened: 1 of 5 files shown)


jschmidb (approved):

ifranzki:
Good point about the testcases. I was able to produce the error with modified testcases, when I zeroise the IV and AAD buffers after the C_EncryptInit/C_DecryptInit calls (needed to allocate and copy them from the const test-vector to be able to modify it) . With that I get testcase failures (due to wrong AAD tag) with the ICA token. The ICA token requires at least the AAD to be available during Final calls, and thus fails if it is not deep-copied.

With the Soft token you never see this error, because the Soft token passes the IV and ADD to OpenSSL during the Init call, and OpenSSL duplicates that data already. So for the Soft token it does not matter if the IV and AAD is no longer valid during Update or Final calls. 

The other tokens don't support AES-GCM at all.

jschmidb (approved):
lgtm

juergenchrist (approved):

PR #533, opened 2022-04-13
CCA/EP11/ICA: Do not use dlclose when in atfork handler

ifranzki:
Calling dlclose() in a atfork handler may cause a deadlock. dlclose() may itself modify the atfork handler table to remove any fork handlers that the to be unloaded library has registered. Since the atfork handler table is currently locked when we are in an atfork handler, this would produce a deadlock.


juergenchrist (approved):

jschmidb (approved):
lgtm

ifranzki:
minor update: also don't call ica_cleanup if we don't call dlclose for libica.

PR #536, opened 2022-04-19
policy: Correct config file names in comments: .conf instead of .cfg

ifranzki:
I would want to include this in the upcoming release.


juergenchrist (approved):

jschmidb (approved):
yep, I saw this one also this afternoon ;-)

PR #537, opened 2022-04-20
EP11: Make CKM_IBM_CPACF_WRAP known by ep11_get_ckm()

ifranzki:
CKM_IBM_CPACF_WRAP is only supported/known inside the EP11 token code, but not externally, thus it is not in the mechtable. Add a special case for this mechanism in ep11_get_ckm() to avoid a warning trace message about that mechanism being unsupported.


jschmidb (approved):
lgtm

PR #540, opened 2022-04-20
Fix compile errors with older libica and one minor fix in pkcsconf

ifranzki:


hfreude (approved):
I only checked the #ifdefs formal but it looks ok.
And the core dump fix is so obviously ...

jschmidb (approved):
Yes, looks good

PR #541, opened 2022-04-25
Corrections/Enhancements on libica loading for ICA and EP11 token

ifranzki:


jschmidb (approved):
Yes, same construction as in libica fips.c ... makes sense. lgtm.

ifranzki:
trigger travis build

PR #542, opened 2022-04-27
pkcsslotd: Add support for AP queue 'change' udev events

ifranzki:
The AP device driver may send an 'change' udev events for AP cards and queues whenever a card or queue is configured offline or online, or set soft-offline or soft-online via chzcrypt.

Treat such events as APQN ADD or REMOVE events and pass it to the applications, similar to bind and unbind udev events.

### DIFF
--- a/usr/sbin/pkcsslotd/socket_server.c
+++ b/usr/sbin/pkcsslotd/socket_server.c
@@ -50,8 +50,11 @@
 #define UDEV_SUBSYSTEM_AP           "ap"
 #define UDEV_ACTION_BIND            "bind"
 #define UDEV_ACTION_UNBIND          "unbind"
+#define UDEV_ACTION_CHANGE          "change"
 #define UDEV_ACTION_DEVTYPE_APQN    "ap_queue"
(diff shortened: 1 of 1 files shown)


usr/sbin/pkcsslotd/socket_server.c
```diff
@@ -1521,18 +1525,23 @@ static int udev_mon_handle_device(struct udev_mon *udev_mon,
     devpath = udev_device_get_devpath(dev);
     devtype = udev_device_get_devtype(dev);
     dev_type_prop = udev_device_get_property_value(dev, UDEV_PROERTY_DEVTYPE);
+    config_prop = udev_device_get_property_value(dev, UDEV_PROERTY_CONFIG);
+    online_prop = udev_device_get_property_value(dev, UDEV_PROERTY_ONLINE);
```
hfreude, line 1529: are you sure this is PROERTY and not PROPERTY ?????

hfreude (approved):
Check for this type PROERTY and maybe solve this, but the rest looks sane.

jschmidb (approved):

PR #543, opened 2022-04-28
config: Add support for bare numbers

ifranzki:
Similar to bare words, a bare number can appear outside of a list and represents its own configuration element.

With that the EP11 config file could also be parsed with the generic config file parser, instead of parsing it 'by hand'.

@juergenchrist Do you see any 'hidden' problems with adding this?


juergenchrist (approved):
I think we just did not think about this requirement when we implemented the first version of the config lib.  It should not produce any conflicts.

PR #546, opened 2022-05-04
TESTCASE: Use CKM_SHA256 for C_GetSessionState tests

ifranzki:
The sess_mgmt testcase uses CKM_MD5 as digest to test the session state. States of CKM_MD5 are not savable since code was changed to no longer use the deprecated OpenSSL MD5 API, but use OpenSSL's EVP API for MD5. The EVP API do no longer allow to export the digest state.

Use CKM_SHA256 instead, and also check for CKR_STATE_UNSAVEABLE on C_GetSessionState calls and skip the testcase if the state is (still) not savable.

The OpenSSL implementation of CKM_SHA256 does support savable states, unless built with OpenSSL >= 3.0.

### DIFF
--- a/testcases/pkcs11/sess_mgmt.c
+++ b/testcases/pkcs11/sess_mgmt.c
@@ -577,6 +577,13 @@ CK_RV do_OperationState1(void)
     // save session #1's operation state
     rc = funcs->C_GetOperationState(session1, NULL, &op_state_len);
     if (rc != CKR_OK) {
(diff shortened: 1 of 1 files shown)


jschmidb:
Looks good. Just saw that we have even more MD5 related code parts in OCK, e.g. in ssl3_func.c. On the one hand it's still legacy code, on the other hand could we try to remove such code, as MD5 is now really outdated.

jschmidb (approved):

PR #547, opened 2022-05-05
Config: support bare string constants, and other fixes

ifranzki:
Add support for a bare quoted string in the generic config parser. This is similar to a bare constant, but allows blanks within the value. 

@juergenchrist Same question again: do you think this could cause problems with the parser? From what I can tell it should not get in the way of other config elements.....

Also fix some things that I have missed with some previous commits.

### DIFF
--- a/testcases/unit/configdump.c
+++ b/testcases/unit/configdump.c
@@ -17,6 +17,11 @@ static const char test6[] = "foo (\n      bar,\n      buzz\n    )\n";
 static const char test7[] = "foo\n(\n  bar,\n  buzz\n)\n";
 static const char nested1[] = "nested {\n  foo (\n        bar,\n        buzz\n      )\n}\n";
 static const char nested2[] = "nested {\n  foo\n  (\n    bar,\n    buzz\n  )\n}\n";
+static const char bare1[] = "foo\nbar\n";
+static const char bare2[] = "foo bar\n";
+static const char bare3[] = "foo\n1234\n";
+static const char bare4[] = "foo\n\"bar\"\n";
+static const char bare5[] = "foo \"bar\"\n";
 static char outbuf[1024];
 
 static const char *curtest;
@@ -74,5 +79,10 @@ int main(void)
     runparsedumptest(test7);
     runparsedumptest(nested1);
     runparsedumptest(nested2);
+    runparsedumptest(bare1);
+    runparsedumptest(bare2);
+    runparsedumptest(bare3);
+    runparsedumptest(bare4);
+    runparsedumptest(bare5);
     return TEST_PASS;
 }
--- a/usr/lib/config/cfgparse.y
+++ b/usr/lib/config/cfgparse.y
@@ -219,13 +219,18 @@ configelem:
     }
     |
     NUMBER eocstar {
-        
         struct ConfigBareNumConstNode *n = confignode_allocbarenumconst($1, @1.first_line);
         if (!n) { YYERROR; }
         $$ = confignode_append(&(n->base), $2);
         $2 = NULL;
     }
-    
+    |
+    STRING_TOK eocstar {
+        struct ConfigBareStringConstNode *n = confignode_allocbarestringconst($1, @1.first_line);
+        if (!n) { YYERROR; }
+        $$ = confignode_append(&(n->base), $2);
+        $2 = NULL;
+    }
 
 /*
 A possibly empty list of barewords or comments.  Two bare words have to be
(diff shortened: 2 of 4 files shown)


juergenchrist (approved):

ifranzki:
Unfortunately the previously added bare number constant does collide with another config element, when it is preceded with a bare constant:

`"foo 1234\n"` is also matched to `BARE NUMBER eocstar BEGIN_DEF commentedconfigelemstar END_DEF eocstar `(i.e. `"conf 42 { subconf = 73 }"`) and results in `line 1, col 10: syntax error, unexpected end of file, expecting EOL or BEGIN_DEF or COMMENT_TOK`

So bare numbers only work if starting on a new line, `"bare\n1234\n"` works....

@juergenchrist Do you see a way how this could be changed?

juergenchrist:
Here is the shift/reduce conflict I did not see in the review:
```
State 13

    6 configelem: BARE . EQUAL NUMBER eocstar
    7           | BARE . EQUAL VERSION_TOK eocstar
    8           | BARE . EQUAL STRING_TOK eocstar
    9           | BARE . EQUAL BARE eocstar
   10           | BARE . NUMBER eocstar BEGIN_DEF commentedconfigelemstar END_DEF eocstar
   11           | BARE . eocstar BEGIN_DEF commentedconfigelemstar END_DEF eocstar
   12           | BARE . eocstar BEGIN_LIST barelist END_LIST eocstar
   13           | BARE . eocstar
   20 eoc: . COMMENT_TOK EOL
   21    | . EOL
   22 eocstar: . eocplus
   23        | . %empty  [$end, BEGIN_DEF, END_DEF, BEGIN_LIST, FILEVERSION, NUMBER, BARE, STRING_TOK]
   24 eocplus: . eoc
   25        | . eoc eocplus

    EOL          shift, and go to state 1
    EQUAL        shift, and go to state 19
    NUMBER       shift, and go to state 20
    COMMENT_TOK  shift, and go to state 2

    NUMBER    [reduce using rule 23 (eocstar)]
    $default  reduce using rule 23 (eocstar)

    eoc      go to state 5
    eocstar  go to state 21
    eocplus  go to state 7
 ```
The problem is the combination of BARE, EOC* and NUMBER.  One simple solution would be to require a newline after a BARE (replace EOC* with EOC+) for real BARE configurations.

**Comment by ifranzki** [comment]
>  One simple solution would be to require a newline after a BARE

That would mean that after a BARE no other tokens can follow on the same line. Unfortunately the EP11 token config file does have at least 2 BARE's in one line, e.g. `PKEY_MODE DISABLED`,  `DIGEST_LIBICA <libica-path>`,  `CPFILTER /etc/opencryptoki/ep11cpfilter.conf`.

We also still have a case with a BARE following a NUMBER on the same line: `LOGLEVEL <value>`.
However `LOGLEVEL` is not supported anymore, but still accepted and ignored for compatibility reasons. I would be OK to drop support for that keyword though. 

For the `APQN_ALLOWLIST` keyword, we usually have the BARE `APQN_ALLOWLIST`  in one line, and the card and domain BARENUMs in the next line(s). But that's just how our example shows it, one could also put everything in one single line....

**Comment by juergenchrist** [comment]
Well, adding something for `LOGLEVEL <value>` is not a problem (rule would be `BARE NUMBER eolstar`), but just a single bare will not work with LALR(1).
Maybe we should go for the more specialized solution and add something like a BARE followed by a list of number pairs.  Something like
```
configelem: ...
| BARE numpairlist

numpairlist: NUMBER NUMBER | NUMBER NUMBER numpairlist
```
Note, however, that an empty `numpairlist` will not work since that again introduces ambiguities.  Also, using a simple number list completely breaks the grammar.

ifranzki:
hmmm, well in the EP11 token config file we do have single BAREs.... currently each on a separate line, so that's why it seems to work so far. We also have a singe BARE in opencryptoki.conf, i.e.  `disable-event-support`

Actually we do have the following in the EP!1 token config file:

- single BARE , e.g. `VHSM_MODE`
- BARE followed by BARE, e.g. `PKEY_MODE DISABLED`.
- BARE followed by quoted string, e.g. `CPFILTER "path"`.
- BARE followed by NUMBER, e.g. `LOGLEVEL <value>`.
- BARE followed by an arbitrary number of pairs of NUMBERs terminated with another BARE. (APQN_ALLOWLIST specification).

We might come to the conclusion that we better keep the 'hand-parsing' of the EP11 token config file, if it causes too much trouble for the generic config parser. Or we use an specific flex/bison based parser for EP11, that supports just what we need for EP11.

ifranzki:
Dropping this PR.
For a reworked version see https://github.com/opencryptoki/opencryptoki/pull/548

PR #545, opened 2022-05-03
Use generic config parser for parsing EP11 token config file

ifranzki:
The syntax of the EP11 token file stays the same, but for the following keywords the 'key = value' and 'key = "value"' form is also accepted:
CPFILTER, PKEY_MODE, DIGEST_LIBICA.

Furthermore, end of line comments (using '#') are now accepted at any places. Previously such comments were only accepted at the beginning of a line.

With this, only the EP11 CP-filter config file is still 'hand-parsed', all other config files use the generic config parser.

PR #548, opened 2022-05-09
config: support number pair lists and bare quoted strings

ifranzki:
This is a try to support parsing the EP11 token config file by adding special node types for a number pair list, and a number pair within a list. This can parse something like this:
```
      APQN_ALLOWLIST # comment 1
       8 13 # comment 2
      10 13 # comment 3
      END # comment 4
```
Makes handling in the consumer a little bit easier than just a series bare number constants.

While at it, remove the bare number const config element, as it causes problems and is no longer required, if we remove the `LOGLEVEL <value>` support from the EP11 config file (was removed in OCK v3.3 in 2015, and is since then allowed, but ignored and just produces a SYSLOG warning but has no effect). 

It still shows the one shift/reduce conflict which is caused by the single BARE element. Unfortunately the opencryptoki.conf already requires single BARE elements (so will the EP11 config file), so we cant simply remove this.  It looks like this conflict does not really hurt so far, at least I have not noticed anything except the following:

Once consequence of this is that after the APQN_ALLOWLIST start bare of a number pair list a EOC must follow, otherwise it expects a indexed structure (`conf 42 { subconf = 73 }`.  I guess I could live with that..... Maybe we should then change it to `BARE eocplus numpairlist BARE eocstar` right away?

But at least this change does not make things worse than it was before.....

We would still need the bare quoted string constants from PR https://github.com/opencryptoki/opencryptoki/pull/547 and PR https://github.com/opencryptoki/opencryptoki/pull/545 would need changes to support the number pair lists.

### DIFF
--- a/usr/lib/config/configuration.c
+++ b/usr/lib/config/configuration.c
@@ -521,22 +521,3 @@ confignode_allocbareconstdumpable(char *key, int line, char *comment)
     }
     return res;
 }
-
-struct ConfigBareNumConstNode *
-confignode_allocbarenumconstdumpable(unsigned long num, int line, char *comment)
-{
-    struct ConfigBareNumConstNode *res;
-    struct ConfigEOCNode *eoc;
-
-    res = confignode_allocbarenumconst(num, line);
-    if (res) {
-        eoc = confignode_alloceoc(comment ? strdup(comment) : NULL, line);
-        if (eoc) {
-            confignode_append(&(res->base), &(eoc->base));
-        } else {
-            confignode_freebarenumconst(res);
-            res = NULL;
-        }
-    }
-    return res;
-}
--- a/usr/lib/config/configuration.c
+++ b/usr/lib/config/configuration.c
@@ -53,6 +53,12 @@ void confignode_free(struct ConfigBaseNode *n)
         case CT_BARECONST:
             confignode_freebareconst(confignode_to_bareconst(n));
             break;
+        case CT_NUMPAIR:
+            confignode_freenumpair(confignode_to_numpair(n));
+            break;
+        case CT_NUMPAIRLIST:
+            confignode_freenumpairlist(confignode_to_numpairlist(n));
+            break;
         default:
             break;
         }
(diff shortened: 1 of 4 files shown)


ifranzki:
>  Also, technically, we should either enforce a "numpairlist" or simply call it "numlist" since it also allows an odd number of list entries.

Why do you think we would not enforce a numpairlist? The numpairlist element requires 2 NUMBERs, so a single number will cause an error:

```
      APQN_ALLOWLIST # comment 1
       8 # comment 2
      10 13 # comment 3
      END # comment 4
```

---> syntax error, unexpected COMMENT_TOK, expecting NUMBER

juergenchrist:
> The numpairlist element requires 2 NUMBERs, so a single number will cause an error

Try three numbers or any other odd number bigger than one.

ifranzki:
> Try three numbers or any other odd number bigger than one.

Didn't I do this already above? I had 3 numbers. 
Tried again, with 3 numbers in one line followed by 2 numbers in another line. 
Still get the expected error:
```
 APQN_ALLOWLIST
       8 13 14
      10 13
 END
```
syntax error, unexpected EOL, expecting NUMBER

```
numpairlist:
    NUMBER NUMBER eocstar
      |
    NUMBER NUMBER eocstar numpairlist
```

So why would you think it would accept an odd number of numbers?

ifranzki:
Maybe its important to realize that I reverted the support for bare numbers in this PR.  So it can't confuse one of the NUMBERs with a bare number.....

usr/lib/config/configuration.c
```diff
@@ -655,3 +661,22 @@ confignode_allocnumpairlistdumpable(char *key, char* end,
     }
     return res;
 }
+
+struct ConfigBareStringConstNode *
+confignode_allocbarestringconstdumpable(char *key, int line, char *comment)
+{
```
jschmidb, line 667: When looking at the existing config-related code in OCK, parameters like "key" are usually consts. Could "char *key" be "const char *key" here as well? Maybe even "comment". Also other places below.

ifranzki: Well, these things could probably be const, but they are not on all the other config structures and functions, so I guess I will leave it as is for now. 
Maybe we can do this for all places at a later time.

usr/lib/config/configuration.c
```diff
@@ -151,6 +160,64 @@ static void confignode_dumpintval(FILE *fp, struct ConfigIntValNode *n,
...
+                }
+                fputc('\n', fp);
+                atbol = 1;
+                break;
+
+            default:
+                break;
+            }
+        }
+
```
jschmidb, line 213: remove empty line?

ifranzki: fixed

jschmidb (approved):

ifranzki:
@juergenchrist ping - do you still have concerns about odd number of NUMBERs ?

usr/lib/config/configuration.h
```diff
@@ -65,18 +65,26 @@
...
+#define CT_BARESTRINGCONST (1u << 13u)
 
 /*
  * Mask for all types that have a key.  This excludes FILEVERSION,
  * EOC, and BARE which reuse the key field for the value.
  */
 #define CT_HAS_KEY_MASK (CT_INTVAL | CT_STRINGVAL | CT_VERSIONVAL |  \
 			 CT_BAREVAL | CT_STRINGVAL | CT_IDX_STRUCT | \
-			 CT_BARELIST | CT_BARECONST)
+			 CT_BARELIST | CT_BARECONST | CT_NUMPAIRLIST | CT_BARESTRINGCONST)
```
juergenchrist, line 87: This looks wrong.  Does a bare string really have an own key?  I thought the key field is reused to store the bare string.  On the other hand, bare consts are in the list.  Should they be in there?

juergenchrist: Well, having it in there makes it possible to `confignode_find` the node by its key string.  So maybe that was intended.

juergenchrist (approved):

usr/lib/config/configuration.h
ifranzki: >having it in there makes it possible to confignode_find the node by its key string 

Yes I guess that was the intention.

PR #549, opened 2022-05-11
Add field ulIvBits to CK_GCM_PARAMS, but also accept old version parameter

ifranzki:
There is some confusion in the PKCS#11 standard, where the doc says nothing about this field, while the official C header has it:
https://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/errata01/os/include/pkcs11-v2.40/pkcs11t.h
Nevertheless, the header files contain the normative definition.

Add code to accept both versions of the CK_GCM_PARAMS structure, with or without the ulIvBits field. Detection is based on the size of the mechanism parameter. When a GCM mechanism parameter without the ulIvBits field is encountered, it is translated into a GCM parameter with that field, setting ulIvBits to ulIvLen * 8.

### DIFF
--- a/usr/lib/common/encr_mgr.c
+++ b/usr/lib/common/encr_mgr.c
@@ -45,6 +45,8 @@ CK_RV encr_mgr_init(STDLL_TokData_t *tokdata,
     CK_RV rc;
     int check;
     CK_ULONG strength = POLICY_STRENGTH_IDX_0;
+    CK_GCM_PARAMS gcm_params;
+    CK_MECHANISM temp_mech;
 
     if (!sess || !ctx || !mech) {
         TRACE_ERROR("Invalid function arguments.\n");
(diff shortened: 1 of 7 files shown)


usr/lib/common/encr_mgr.c
```diff
@@ -523,13 +525,23 @@ CK_RV encr_mgr_init(STDLL_TokData_t *tokdata,
...
+             mech->ulParameterLen != sizeof(CK_GCM_PARAMS_COMPAT)) ||
             mech->pParameter == NULL) {
             TRACE_ERROR("%s\n", ock_err(ERR_MECHANISM_PARAM_INVALID));
             rc = CKR_MECHANISM_PARAM_INVALID;
             goto done;
         }
 
+        if (mech->ulParameterLen == sizeof(CK_GCM_PARAMS_COMPAT)) {
+            aes_gcm_param_from_compat((CK_GCM_PARAMS_COMPAT *)mech->pParameter,
+                                      &gcm_params);
```
jschmidb, line 538: Routine aes_gcm_param_from_compat has a return code CK_RV, which is CKR_ARGUMENTS_BAD if one or both of the parms are NULL. Here, the routine is used without rc checking. However, it's obvious that the parms are not NULL. Just to mention ...

usr/lib/common/mech_aes.c
```diff
@@ -3472,6 +3474,21 @@ CK_RV aes_gcm_free_param(CK_GCM_PARAMS *params)
     return CKR_OK;
 }
 
+CK_RV aes_gcm_param_from_compat(CK_GCM_PARAMS_COMPAT *from, CK_GCM_PARAMS *to)
+{
```
jschmidb, line 3478: parm from could be const

usr/lib/common/encr_mgr.c
ifranzki: will change that function to void. Its an internal function, and the parameters have been NULL checked already.

usr/lib/common/mech_aes.c
ifranzki: will fix

PR #550, opened 2022-05-11
EP11: CKA_DERIVE is not applicable for RSA, DSA, and DILITHIUM keys

ifranzki:
Does not happen in our test cases, but you never know what applications do.....

### DIFF
--- a/usr/lib/ep11_stdll/ep11_specific.c
+++ b/usr/lib/ep11_stdll/ep11_specific.c
@@ -1825,6 +1825,8 @@ static CK_BBOOL attr_applicable_for_ep11(STDLL_TokData_t * tokdata,
     case CKK_RSA:
         if (class == CKO_PRIVATE_KEY && attr->type == CKA_PUBLIC_EXPONENT)
             return CK_FALSE;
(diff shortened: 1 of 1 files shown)


jschmidb:
Yes, that's correct. Also note that Edwards curves (ed25519 and ed448) do not support derive, but x25519 and x448 do.

jschmidb (approved):

ifranzki:
> Yes, that's correct. Also note that Edwards curves (ed25519 and ed448) do not support derive, but x25519 and x448 do.

Right, this was already covered by the code:

```
        /* Montgomery curves cannot be used for sign/verify */
        if (class == CKO_PRIVATE_KEY && curve_type == MONTGOMERY_CURVE && attr->type == CKA_SIGN)
            return CK_FALSE;
        if (class == CKO_PUBLIC_KEY && curve_type == MONTGOMERY_CURVE && attr->type == CKA_VERIFY)
            return CK_FALSE;
        /* Edwards curves cannot be used for derive */
        if (curve_type == EDWARDS_CURVE && attr->type == CKA_DERIVE)
            return CK_FALSE;
        break;
```

PR #552, opened 2022-05-16
CCA: EC signature verify may also return 12 / 769 for bad signatures

ifranzki:
According to the description of CCA verb "Digital Signature Verify" (CSNDDSV) in "Secure Key Solution with the Common Cryptographic Architecture Application Programmer's Guide Version 7.1" the signature verify request may also return 12 / 769 for bad signatures, in addition to 4 / 429. Treat both return codes as bad signature indication.


jschmidb (approved):
lgtm

PR #553, opened 2022-05-17
Write out short message when ab_tests are not supported on a specific…

chrisRund:
Hello Ingo,

as discussed. Please review. 
Thanks.

Signed-off-by: Christian Rund <christian.rund@de.ibm.com>


testcases/crypto/abfunc.c
```diff
@@ -1191,6 +1191,7 @@ int main(int argc, char **argv)
         }
     } else {
         rc = 0;
+        printf("%s only supported on the EP11 token.\n", argv[0]);
```
ifranzki, line 1194: Maybe better use testsuite_skip() so that it produces a correct SKIP message?

ifranzki: well, I mean testcase_skip() ....

chrisRund: Will do, good idea, the skip message is even more informative.
Thanks.

ifranzki:
now we have 4 commits (one merge commit).... Thats not going to work.
Please squash everything into one single commit.

juergenchrist (approved):

ifranzki (approved):

PR #554, opened 2022-05-18
COMMON: Retry RSA key gen if it fails with OpenSSL 3.0

ifranzki:
In OpenSSL 3.0 the RSA key gen algorithm has been changed and can now fail to generate a key. Retry up to 10 times in such a case.


ifranzki:
@jschmidb ping ?

jschmidb (approved):
lgtm. Same as for libica.

ifranzki:
Just added a retry for the ICA token as well. Although the libica also has this retry already, we might run with an distro built libica that does not have the retry logic yet. So do the retry in the ICA token as well.

jschmidb (approved):
perfect

PR #555, opened 2022-05-23
EP11/CCA: Check that all configure APQNs have the same and expected master key and check keys after creation to have the expected MKVP

ifranzki:


usr/lib/cca_stdll/cca_specific.c
```diff
@@ -567,41 +589,250 @@ static CK_RV cca_resolve_lib_sym(void *hdl)
...
+
+        OCK_SYSLOG(LOG_ERR, "Error parsing config file '%s': unexpected token "
+                   "'%s' at line %d\n", fname, c->key, c->line);
+        TRACE_ERROR("Error parsing config file '%s': unexpected token '%s' "
+                    "at line %d\n", fname, c->key, c->line);
+        return CKR_FUNCTION_FAILED;
+    }
+
+    TRACE_DEBUG("Expected master key verification patters\n");
+    if (cca_private->expected_sym_mkvp_set == TRUE)
```
jschmidb: typo: patterns

usr/lib/cca_stdll/ccatok.conf
```diff
@@ -0,0 +1,19 @@
+version cca-0
+
+# Optionally specify the expected master key verification patters for the 
+# SYM, AES, and APKA master key. The CCA token does not use the ASYM master key.
```
jschmidb, line 4: patterns

usr/lib/cca_stdll/ccatok.conf
```diff
@@ -0,0 +1,19 @@
...
+#
+# You can use the TKE or panel.exe to query the curent master key verification
+# pattern:
```
jschmidb, line 7: current

jschmidb:
One typo in commit msg:  CCA: Get and check the CCA host library version
... used in a following commit requires    <- require

jschmidb:
Typo in commit msg:  CCA: Get master key verification patterns in token initialization
Obtain the master key verification patters        <- patterns

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -648,6 +655,426 @@ static CK_RV cca_get_version(STDLL_TokData_t *tokdata)
...
+
+    id = (unsigned short *)(verb_data + CCA_STATICSB_APKA_CMK_ID_OFFSET);
+    if (*id != CCA_STATICSB_APKA_CMK_ID) {
+        TRACE_ERROR("CSUACFQ (STATICSB) APKA MKVP not available for %s/%s\n",
+                    adapter, domain);
+        return CKR_FUNCTION_FAILED;
+    }
+
+    TRACE_DEBUG("Master key verification patters for %s/%s\n",
+                adapter, domain);
```
jschmidb, line 891: patterns

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -447,6 +447,57 @@ static CK_BBOOL analyse_cca_key_token(const CK_BYTE *t, CK_ULONG tlen,
...
+    case sec_hmac_key:
+        expected_mkvp = cca_private->expected_aes_mkvp;
+        mktype = "AES";
+        break;
+
+    case sec_rsa_priv_key:
+    case sec_ecc_priv_key:
+        expected_mkvp = cca_private->expected_apka_mkvp;
+        mktype = "APKA";
+        break;
```
jschmidb, line 480: Shouldn't be RSA = ASYM, but ECC = APKA?

ifranzki: No APKA is used for 'newer' asymmetric key tokens, and this includes besides ECC also the newer RSA key token types that opencryptoki is using.

jschmidb:
Little typo in commit msg  CCA: Add APQN event handling support
When an event for an CCA APQN ...    <- a CCA APQN

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -648,6 +655,426 @@ static CK_RV cca_get_version(STDLL_TokData_t *tokdata)
...
+    CK_RV rc;
+
+    /* Get current adapter serial number */
+    memcpy(rule_array, "STATCRD2", 8);
+    rule_array_count = 1;
+    verb_data_length = 0;
+    dll_CSUACFQ(&return_code, &reason_code,
+                NULL, NULL,
+                &rule_array_count, rule_array,
+                &verb_data_length, NULL);
```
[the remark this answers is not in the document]
ifranzki: did you actually see this return code ever? Or did you just look for all possible return codes and this one sounded like it needs to be handled? I would tend to not take care about this return code at the moment, unless we see it happening in real life.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -6839,3 +6970,109 @@ CK_RV token_specific_reencrypt_single(STDLL_TokData_t *tokdata,
...
+    end = memchr(buf, '\n', buflen);
+    if (end)
+        *end = 0;
+    else
+        buf[buflen - 1] = 0;
+
+    if (strlen(buf) == 0) {
+        rc = CKR_FUNCTION_FAILED;
+        goto out_fclose;
+    }
```
jschmidb, line 7002: goto not necessary here

jschmidb:
Typo in commit msg  EP11: Add config option to specify the expected WKVP
...  
... can only be up to 64 bits (8 bytes), but the EKVP is 16 bytes.    <- WKVP

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -250,6 +254,24 @@ enum cca_token_type {
...
+    unsigned char expected_apka_mkvp[CCA_MKVP_LENGTH];
+    CK_BBOOL expected_sym_mkvp_set;
+    CK_BBOOL expected_aes_mkvp_set;
+    CK_BBOOL expected_apka_mkvp_set;
+};
+
+#define CCA_CFG_EXPECTED_MKVPS  "EXPECTED_MKVPS"
+#define CCA_CFG_SYM_MKVP        "SYM"
+#define CCA_CFG_AES_MKVP        "AES"
+#define CCA_CFG_APKA_MKVP       "APKA"
```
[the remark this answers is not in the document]
ifranzki: because openCryptoki does not use key tokens that use ASYM master keys.

usr/lib/cca_stdll/cca_specific.c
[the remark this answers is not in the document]
jschmidb: I saw it in the testcases project, /crypto/pkeytools/ccagenecckey.c

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -629,17 +861,29 @@ CK_RV token_specific_init(STDLL_TokData_t * tokdata, CK_SLOT_ID SlotNumber,
...
 
 CK_RV token_specific_final(STDLL_TokData_t *tokdata,
                            CK_BBOOL in_fork_initializer)
 {
+    struct cca_private_data *cca_private = tokdata->private_data;
+
     TRACE_INFO("cca %s running\n", __func__);
 
-    free(tokdata->mech_list);
+    if (tokdata->mech_list != NULL)
```
hfreude, line 1415: you don't need to check for NULL before calling free()

usr/lib/cca_stdll/cca_specific.c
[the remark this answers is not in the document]
ifranzki: ok, but there is is for the ECC key gen verb....  I guess I would still tend to leave it as is. We do not handle this return code at any place in the CCA token and as far as I can tell we never actually saw it being retuned, not even for the key gen verbs....

usr/lib/cca_stdll/ccatok.conf
hfreude, line 4: Well are you sure the CCA token does not use the ASYM master key ? Each RSA key pair generate uses it...

usr/lib/cca_stdll/ccatok.conf
hfreude, line 7: and of course the sysfs api also provides this info.
I still don't have this als part of lszcrypt but this comes soon.

usr/lib/cca_stdll/ccatok.conf
ifranzki: No APKA is used for 'newer' asymmetric key tokens, and this includes besides ECC also the newer RSA key token types that opencryptoki is using.

usr/lib/cca_stdll/ccatok.conf
ifranzki: right, will add this info.

usr/lib/cca_stdll/cca_specific.c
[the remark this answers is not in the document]
hfreude: Well, there are some CCA coding examples which use this kind of loop. To be honest I never saw it and there is tons of code out there which does not check for this funny 1105.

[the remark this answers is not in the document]
jschmidb: OK, so I will leave it out also for libzpc then

usr/lib/cca_stdll/cca_specific.c
hfreude: Ok, if you are sure ... Well the 7.1 CCA docu says:
 On the cryptographic coprocessor, PKA keys are protected by the Asymmetric-Keys Master Key (ASYM-MK).
The ASYM-MK is a triple-length DES key used to protect PKA private keys. On the cryptographic coprocessor, the ASYM-MK protects RSA private keys. There are two PKA master keys: the ASYM-MK mentioned above, and the 256-bit AES PKA Master Key (APKA-MK), used to protect ECC private keys stored in ECC key tokens.

usr/lib/ep11_stdll/ep11tok.conf
```diff
@@ -77,6 +77,15 @@
 #                         but not CKA_IBM_PROTKEY_EXTRACTABLE, new keys get 
 #                         CKA_IBM_PROTKEY_EXTRACTABLE=true internally.
 # 
+# Optionally specify the expected wrapping key verification pattern. When 
+# specified, all APQNs configured for the token must have been set up with this
+# wrapping key verification pattern.
+# You can use the TKE or ep11info to query the curent wrapping key verification
+# pattern of an APQN:
```
hfreude, line 84: or sysfs

hfreude:
Ingo, resolve this thing with ASYM. And it would be nice to have also hints to the sysfs attributes for the user to fetch mkvps values (well, DES isn't there). But in general I approve this PR.

ifranzki:
Regarding ASYM vs APKA master keys for CCA:

According to the CCA manual, the ASYM master key is a 3DES key (192 bits), similar to the SYM master key. It is used to encrypt older RSA secure keys. ASYM master keys are sometimes also called PKA master keys. 
The APKA master key is an AES 256 bit key, similar to the AES master key.

Section "Symmetric and asymmetric master keys" in chapter 6 "Chapter 6. Understanding and managing master keys" states:

_The PKA master-key register set is used to wrap and unwrap RSA (asymmetric) private working keys, or
the object protection keys (OPKs) of the RSA working keys that have an OPK defined, **excluding private
key sections X'30' and X'31', which have their OPKs wrapped and unwrapped by the APKA master key**._

_The APKA master-key register set is used to wrap and unwrap the Object Protection Key (OPK) that is
itself used to wrap the key material of an **Elliptic Curve Cryptography (ECC) key or the OPK of RSA
private key sections X'30' and X'31'**. ECC keys are asymmetric._

Also, the description of verb PKA Key Generate (CSNDPKG) describes:

_Enciphered using a master key
When rule-array keyword MASTER is specified, a master key is used to protect the private key or its
OPK if the private key has one. The enciphered key is returned in an internal PKA key-token. **ECC keys
(section X'20') and RSA keys in section X'30' and X'31' have an OPK that is enciphered by the APKA
master key**. All other RSA keys are enciphered by the PKA master key._

The CCA token of Opencryptoki uses RSA keys with sections X'30' and X'31' only. 
Section X'30' corresponds to  keyword RSA-AESM (ME format) , and section X'31' corresponds to RSA-AESC (CRT format). The CCA token uses "RSA-AESC" for generating and for importing RSA keys. 

@hfreude @jschmidb does this clarify the use of APKA vs ASYM ?

usr/lib/ep11_stdll/ep11tok.conf
ifranzki: will add.

usr/lib/cca_stdll/cca_specific.c
ifranzki: The CCA doc is not very consistent when it talks about APKA master keys and where they are used. They always mention ECC keys, but only sometimes mention newer RSA key tokens. I guess the confusion is because the APKA master keys were introduced with CCA's ECC support, and thus ECC were the first that uses APKA keys. When newer RSA key tokens were added later on, they forgot to adjust many places in the manual that talk about APKA and ECC only....

See my comment below that hopefully clarifies it.

jschmidb:
OK, I agree. I wasn't aware of this exception with RSA sections x'30' and x'31.

jschmidb (approved):

hfreude (approved):
I approve.

PR #557, opened 2022-05-24
TESTCASE: Accept CKR_FUNCTION_FAILED for invalid EC signatures with ICA token

ifranzki:
Due to improper return code handling in libica, an EC signature verification using an (intentionally) incorrect signature may return CKR_FUNCTION_FAILED when using the ICA token.

This has already been fixed in libica, but it may take a while until the fix arrives in the distro used in the CI. Until then the testcases would continue to fail. Temporarily accept CKR_FUNCTION_FAILED in the test case for this situation.


jschmidb (approved):
lgtm

PR #559, opened 2022-06-07
pkcsstats: Fix JSON output in case of errors

ifranzki:


jschmidb:
The code looks good. 
Routine print_json_start() obtains some system information via uname(). When looking at the manpage of uname (https://man7.org/linux/man-pages/man2/uname.2.html), I found two points interesting:

- The field nodename seems to be meaningless, so maybe we could just drop the related output here
- Part of the utsname information is also accessible via /proc/sys/kernel/{ostype, hostname, osrelease, version,domainname}, so we could read this info directly from disk. But I see the point that this would mean more code and it's far easier to just call uname().

jschmidb (approved):

Issue #562, opened 2022-06-09
multiple defects found by covscan in opencryptoki-3.18.0


ngothan:
there are multiple defects in opencryptoki-3.18.0 found by covscan. The list of defects is large. I attach the covscan-result.log file.

[opencryptoki-3.18.0-covscan.log](https://github.com/opencryptoki/opencryptoki/files/8872836/opencryptoki-3.18.0-covscan.log)

It looks like there are a lot of detects which are false positives.

Thanks

ifranzki:
Thanks for reporting this. I will go over the list as time permits, but this will take a while, I guess.

PR #564, opened 2022-07-19
Fixes for coverty scan findings

ifranzki:
Closes: https://github.com/opencryptoki/opencryptoki/issues/562

### DIFF
--- a/usr/lib/api/policy.c
+++ b/usr/lib/api/policy.c
@@ -315,7 +315,7 @@ static CK_RV policy_get_sig_size(CK_MECHANISM_PTR mech, struct objstrength *s,
     const struct mechrow *col = mechrow_from_numeric(mech->mechanism);
     CK_ULONG size;
 
-    if (!col)
+    if (!col || !s)
         return CKR_FUNCTION_FAILED;
     if (col->flags & MCF_MAC_GENERAL) {
         if (mech->ulParameterLen != sizeof(CK_MAC_GENERAL_PARAMS)) {
--- a/usr/lib/api/policy.c
+++ b/usr/lib/api/policy.c
@@ -1717,7 +1717,8 @@ CK_RV policy_load(struct policy *p)
                     strerror(err));
         OCK_SYSLOG(LOG_ERR, "POLICY: Failed to open " OCK_POLICY_CFG ": %s\n",
                    strerror(err));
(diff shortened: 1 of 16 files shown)


juergenchrist (approved):

usr/lib/api/policy.c
```diff
@@ -315,7 +315,7 @@ static CK_RV policy_get_sig_size(CK_MECHANISM_PTR mech, struct objstrength *s,
     const struct mechrow *col = mechrow_from_numeric(mech->mechanism);
     CK_ULONG size;
 
-    if (!col)
+    if (!col || !s)
```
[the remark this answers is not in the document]
ifranzki: I would say no, but its hard to tell.
policy_get_sig_size() is for example called from policy_is_mech_allowed() dependent on the check type, passing the caller provded s.
policy_is_mech_allowed() is called with s = NULL, but as far as I can tell only for such checks that do not require s (i.e. for digests or keygen).

usr/lib/common/loadsave.c
```diff
@@ -811,7 +811,12 @@ CK_RV load_private_token_objects_old(STDLL_TokData_t *tokdata)
...
             continue;
         }
-        //size--;
+        if (size <= sizeof(CK_ULONG_32) + sizeof(CK_BBOOL)) {
+            fclose(fp2);
+            OCK_SYSLOG(LOG_ERR, "Improper size of object %s (ignoring it)\n",
+                       fname);
+            continue;
+        }
         size = size - sizeof(CK_ULONG_32) - sizeof(CK_BBOOL);
```
holger-dengler, line 820: It is not part of your change, but maybe you can also cover the next line as well

size -= sizeof(CK_ULONG_32) - sizeof(CK_BBOOL);

holger-dengler (approved):
Looks good to me. I approve this PR with or without my suggestions.

usr/lib/common/loadsave.c
juergenchrist: I think I know why this was not done.  Your version is wrong!!!  The second minus then has to be a plus.

holger-dengler: You're right. It should be 
`size -= sizeof(CK_ULONG_32) + sizeof(CK_BBOOL);`

PR #565, opened 2022-07-20
v4: pkcs11 URI generation support

holger-dengler:
This PR contains a first implementation of the PKCS#11 URI generation support. It applies on top of my latest maint/autotools-cleanup branch. So please ignore patches 1-9/15.

First, a few generic helper functions are added, like a strlen-helper for non-null terminated strings in CK-structs and a string buffer with a dynamic memory management. Patch 12/15 contains the pkcs11 URI implementation. Patch 13-15/15 make use of it in the related tools.

As soon as maint/autotools-cleanup has been accepted and mergen into master, I'll provice a rebased version of this branch.

### DIFF
--- a/testcases/misc_tests/p11sak_test.sh
+++ b/testcases/misc_tests/p11sak_test.sh
@@ -148,19 +148,19 @@ fi
 
 
 # CK_BBOOL
-if [[ $(grep -A 19 'p11sak-des' $P11SAK_DES_LONG | grep -c 'CKA_IBM_PROTKEY_EXTRACTABLE: CK_FALSE') == "1" ]]; then
+if [[ $(grep -A 20 'p11sak-des' $P11SAK_DES_LONG | grep -c 'CKA_IBM_PROTKEY_EXTRACTABLE: CK_FALSE') == "1" ]]; then
 echo "* TESTCASE list-key des PASS Listed random des public keys CK_BBOOL attribute"
 else
 echo "* TESTCASE list-key des FAIL Failed to list des public keys CK_BBOOL attribute"
 fi
 # CK_ULONG
-if [[ $(grep -A 19 'p11sak-des' $P11SAK_DES_LONG | grep -c 'CKA_MODULUS_BITS:') == "0" ]]; then
+if [[ $(grep -A 20 'p11sak-des' $P11SAK_DES_LONG | grep -c 'CKA_MODULUS_BITS:') == "0" ]]; then
 echo "* TESTCASE list-key des PASS Listed random des public keys CK_ULONG attribute"
 else
 echo "* TESTCASE list-key des FAIL Failed to list des public keys CK_ULONG attribute"
 fi
 # CK_BYTE
-if [[ $(grep -A 19 'p11sak-des' $P11SAK_DES_LONG | grep -c 'CKA_MODULUS:') == "0" ]]; then
+if [[ $(grep -A 20 'p11sak-des' $P11SAK_DES_LONG | grep -c 'CKA_MODULUS:') == "0" ]]; then
 echo "* TESTCASE list-key des PASS Listed random des public keys CK_BYTE attribute"
 else
 echo "* TESTCASE list-key des FAIL Failed to list des public keys CK_BYTE attribute"
@@ -185,19 +185,19 @@ fi
 
 
 # CK_BBOOL
-if [[ $(grep -A 19 'p11sak-3des' $P11SAK_3DES_LONG | grep -c 'CKA_IBM_PROTKEY_EXTRACTABLE: CK_FALSE') == "1" ]]; then
+if [[ $(grep -A 20 'p11sak-3des' $P11SAK_3DES_LONG | grep -c 'CKA_IBM_PROTKEY_EXTRACTABLE: CK_FALSE') == "1" ]]; then
 echo "* TESTCASE list-key 3des PASS Listed random 3des public keys CK_BBOOL attribute"
 else
 echo "* TESTCASE list-key 3des FAIL Failed to list 3des public keys CK_BBOOL attribute"
 fi
 # CK_ULONG
-if [[ $(grep -A 19 'p11sak-3des' $P11SAK_3DES_LONG | grep -c 'CKA_MODULUS_BITS:') == "0" ]]; then
+if [[ $(grep -A 20 'p11sak-3des' $P11SAK_3DES_LONG | grep -c 'CKA_MODULUS_BITS:') == "0" ]]; then
 echo "* TESTCASE list-key 3des PASS Listed random 3des public keys CK_ULONG attribute"
 else
 echo "* TESTCASE list-key 3des FAIL Failed to list 3des public keys CK_ULONG attribute"
 fi
 # CK_BYTE
-if [[ $(grep -A 19 'p11sak-3des' $P11SAK_3DES_LONG | grep -c 'CKA_MODULUS:') == "0" ]]; then
+if [[ $(grep -A 20 'p11sak-3des' $P11SAK_3DES_LONG | grep -c 'CKA_MODULUS:') == "0" ]]; then
 echo "* TESTCASE list-key 3des PASS Listed random 3des public keys CK_BYTE attribute"
 else
 echo "* TESTCASE list-key 3des FAIL Failed to list 3des public keys CK_BYTE attribute"
@@ -256,19 +256,19 @@ fi
 
 
 # CK_BBOOL
-if [[ $(grep -A 57 'p11sak-aes-128' $P11SAK_AES_LONG | grep -c 'CKA_IBM_PROTKEY_EXTRACTABLE: CK_FALSE') == "3" ]]; then
+if [[ $(grep -A 60 'p11sak-aes-128' $P11SAK_AES_LONG | grep -c 'CKA_IBM_PROTKEY_EXTRACTABLE: CK_FALSE') == "3" ]]; then
 echo "* TESTCASE list-key aes PASS Listed random aes public keys CK_BBOOL attribute"
 else
 echo "* TESTCASE list-key aes FAIL Failed to list aes public keys CK_BBOOL attribute"
 fi
 # CK_ULONG
-if [[ $(grep -A 57 'p11sak-aes-128' $P11SAK_AES_LONG | grep -c 'CKA_MODULUS_BITS:') == "0" ]]; then
+if [[ $(grep -A 60 'p11sak-aes-128' $P11SAK_AES_LONG | grep -c 'CKA_MODULUS_BITS:') == "0" ]]; then
 echo "* TESTCASE list-key aes PASS Listed random aes public keys CK_ULONG attribute"
 else
 echo "* TESTCASE list-key aes FAIL Failed to list aes public keys CK_ULONG attribute"
 fi
 # CK_BYTE
-if [[ $(grep -A 57 'p11sak-aes-128' $P11SAK_AES_LONG | grep -c 'CKA_MODULUS:') == "0" ]]; then
+if [[ $(grep -A 60 'p11sak-aes-128' $P11SAK_AES_LONG | grep -c 'CKA_MODULUS:') == "0" ]]; then
 echo "* TESTCASE list-key aes PASS Listed random aes public keys CK_BYTE attribute"
 else
 echo "* TESTCASE list-key aes FAIL Failed to list aes public keys CK_BYTE attribute"
@@ -378,19 +378,19 @@ fi
 
 
 # CK_BBOOL
-if [[ $(grep -A 205 'p11sak-rsa-1024:pub' $P11SAK_RSA_LONG | grep -c 'CKA_IBM_PROTKEY_EXTRACTABLE: CK_FALSE') == "6" ]]; then
+if [[ $(grep -A 211 'p11sak-rsa-1024:pub' $P11SAK_RSA_LONG | grep -c 'CKA_IBM_PROTKEY_EXTRACTABLE: CK_FALSE') == "6" ]]; then
 echo "* TESTCASE list-key rsa PASS Listed random rsa public keys CK_BBOOL attribute"
 else
 echo "* TESTCASE list-key rsa FAIL Failed to list rsa public keys CK_BBOOL attribute"
 fi
 # CK_ULONG
-if [[ $(grep -A 205 'p11sak-rsa-1024:pub' $P11SAK_RSA_LONG | grep -c 'CKA_MODULUS_BITS:') == "3" ]]; then
+if [[ $(grep -A 211 'p11sak-rsa-1024:pub' $P11SAK_RSA_LONG | grep -c 'CKA_MODULUS_BITS:') == "3" ]]; then
 echo "* TESTCASE list-key rsa PASS Listed random rsa public keys CK_ULONG attribute"
 else
 echo "* TESTCASE list-key rsa FAIL Failed to list rsa public keys CK_ULONG attribute"
 fi
 # CK_BYTE
-if [[ $(grep -A 205 'p11sak-rsa-1024:pub' $P11SAK_RSA_LONG | grep -c 'CKA_MODULUS:') == "6" ]]; then
+if [[ $(grep -A 211 'p11sak-rsa-1024:pub' $P11SAK_RSA_LONG | grep -c 'CKA_MODULUS:') == "6" ]]; then
 echo "* TESTCASE list-key rsa PASS Listed random rsa public keys CK_BYTE attribute"
 else
 echo "* TESTCASE list-key rsa FAIL Failed to list rsa public keys CK_BYTE attribute"
@@ -500,19 +500,19 @@ fi
 
 
 # CK_BBOOL
-if [[ $(grep -A 84 'p11sak-ec-prime256v1:pub' $P11SAK_EC_LONG | grep -c 'CKA_IBM_PROTKEY_EXTRACTABLE: CK_FALSE') == "6" ]]; then
+if [[ $(grep -A 90 'p11sak-ec-prime256v1:pub' $P11SAK_EC_LONG | grep -c 'CKA_IBM_PROTKEY_EXTRACTABLE: CK_FALSE') == "6" ]]; then
 echo "* TESTCASE list-key ec PASS Listed random ec public keys CK_BBOOL attribute"
 else
 echo "* TESTCASE list-key ec FAIL Failed to list ec public keys CK_BBOOL attribute"
 fi
 # CK_ULONG
-if [[ $(grep -A 84 'p11sak-ec-prime256v1:pub' $P11SAK_EC_LONG | grep -c 'CKA_MODULUS_BITS:') == "0" ]]; then
+if [[ $(grep -A 90 'p11sak-ec-prime256v1:pub' $P11SAK_EC_LONG | grep -c 'CKA_MODULUS_BITS:') == "0" ]]; then
 echo "* TESTCASE list-key ec PASS Listed random ec public keys CK_ULONG attribute"
 else
 echo "* TESTCASE list-key ec FAIL Failed to list ec public keys CK_ULONG attribute"
 fi
 # CK_BYTE
-if [[ $(grep -A 84 'p11sak-ec-prime256v1:pub' $P11SAK_EC_LONG | grep -c 'CKA_MODULUS:') == "0" ]]; then
+if [[ $(grep -A 90 'p11sak-ec-prime256v1:pub' $P11SAK_EC_LONG | grep -c 'CKA_MODULUS:') == "0" ]]; then
 echo "* TESTCASE list-key ec PASS Listed random ec public keys CK_BYTE attribute"
 else
 echo "* TESTCASE list-key ec FAIL Failed to list ec public keys CK_BYTE attribute"
--- a/testcases/misc_tests/p11sak_test.sh
+++ b/testcases/misc_tests/p11sak_test.sh
@@ -165,6 +165,12 @@ echo "* TESTCASE list-key des PASS Listed random des public keys CK_BYTE attribu
 else
 echo "* TESTCASE list-key des FAIL Failed to list des public keys CK_BYTE attribute"
 fi
+# URI
+if [[ $(grep -A 20 'p11sak-des' $P11SAK_DES_LONG | grep -c 'URI: pkcs11:.*type=secret-key') == "1" ]]; then
+echo "* TESTCASE list-key des PASS list des key pkcs#11 URI"
+else
+echo "* TESTCASE list-key des FAIL list des key pkcs#11 URI"
+fi
 
 
 # check 3DES
@@ -201,6 +207,12 @@ if [[ $(grep -A 20 'p11sak-3des' $P11SAK_3DES_LONG | grep -c 'CKA_MODULUS:') == 
 echo "* TESTCASE list-key 3des PASS Listed random 3des public keys CK_BYTE attribute"
 else
 echo "* TESTCASE list-key 3des FAIL Failed to list 3des public keys CK_BYTE attribute"
+fi
+# URI
+if [[ $(grep -A 20 'p11sak-3des' $P11SAK_3DES_LONG | grep -c 'URI: pkcs11:.*type=secret-key') == "1" ]]; then
+echo "* TESTCASE list-key 3des PASS list 3des key pkcs#11 URI"
+else
+echo "* TESTCASE list-key 3des FAIL list 3des key pkcs#11 URI"
 fi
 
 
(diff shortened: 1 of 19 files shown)


ifranzki:
There are compile errors in Travis, please take a look.

ifranzki:
it sees that this PR also include the commits from your other PR ?
I guess you need to rebase on top of current master to get a clean PR.

holger-dengler:
I'll do a rebase to master, as soon as the previous PR has been accepted and integrated. 

I'll also check the build errors. I'm a bit confused, because I checked my build outputs very carefully. Is there a way to trigger travis runs before sending a PR? Or just to the build checks locally?

ifranzki:
> Is there a way to trigger travis runs before sending a PR?

No, not that I know of.

usr/lib/common/buffer.c
```diff
@@ -0,0 +1,108 @@
...
+{
+    size_t new_len = strlen(buf->b) + len;
+
+    if (!s)
+        return strlen(buf->b);
+
+    /* extend buffer if required */
+    if ((new_len) > (buf->b_size - 1)) {
+        size_t new_b_size =
+            (((new_len) / BUFFER_INIT_SIZE) + 1) *BUFFER_INIT_SIZE;
```
ifranzki, line 76: space between '*' and 'BUFFER_INIT_SIZE'

usr/lib/common/buffer.c
```diff
@@ -0,0 +1,108 @@
...
+int p11_buffer_append(p11_buffer_t *buf, const char *s)
+{
+    return p11_buffer_append_len(buf, s, strlen(s));
+}
+
+int p11_buffer_printf(p11_buffer_t *buf, const char *fmt, ...)
+{
+    va_list ap;
+    int len;
+
```
ifranzki, line 100: remove empty line

usr/lib/common/buffer.c
```diff
@@ -0,0 +1,108 @@
...
+    strncat(buf->b, s, len);
+    return new_len;
+}
+
+int p11_buffer_append(p11_buffer_t *buf, const char *s)
+{
+    return p11_buffer_append_len(buf, s, strlen(s));
+}
+
+int p11_buffer_printf(p11_buffer_t *buf, const char *fmt, ...)
```
ifranzki, line 96: if I understand the function correctly, this appends the result of the printf result to the buffer. Should the function then better be named something including 'append', i.e. p11_buffer_append_printf ? 
From the name p11_buffer_printf I originally thought, it would print the buffer contents.....

usr/lib/common/buffer.c
```diff
@@ -0,0 +1,108 @@
...
+{
+    va_list ap;
+    int len;
+
+    char tmp[BUFFER_INIT_SIZE];
+
+    va_start(ap, fmt);
+    len = vsnprintf(tmp, BUFFER_INIT_SIZE - 1, fmt, ap);
```
ifranzki, line 104: BUFFER_INIT_SIZE - 1 --> sizeof(tmp) - 1

usr/lib/common/buffer.c
```diff
@@ -0,0 +1,108 @@
...
+{
+    return p11_buffer_append_len(buf, s, strlen(s));
+}
+
+int p11_buffer_printf(p11_buffer_t *buf, const char *fmt, ...)
+{
+    va_list ap;
+    int len;
+
+    char tmp[BUFFER_INIT_SIZE];
```
ifranzki, line 101: so you limit the amount of chars to be appended to BUFFER_INIT_SIZE? And if it overruns, you skip silently?
How about checking the length after the vsnprintf, and if it indicates that chars were skipped, then allocate a temp buffer dynamically?

usr/lib/common/uri.c
```diff
@@ -0,0 +1,268 @@
...
+        if ((ul) != (CK_ULONG)-1) {                      \
+            p11_buffer_append_len(buf, sep, 1);          \
+            p11_buffer_printf(buf, txt "%lu", ul);       \
+            *sep = P11_URI_SEP_ATTR;                     \
+        }                                                \
+    } while(0)
+
+static const char HEX_CHARS_LOWER[] = "0123456789abcdef";
+
+static char *c2t(CK_OBJECT_CLASS_PTR class)
```
ifranzki, line 100: How about moving this function to p11utils.c, i.e. with a different name, e.g. p11_get_cko() ?
This might be useful in other cases, too.

usr/lib/common/uri.c
```diff
@@ -0,0 +1,268 @@
...
+        return "cert";
+    case CKO_DATA:
+        return "data";
+    case CKO_PRIVATE_KEY:
+        return "private";
+    case CKO_PUBLIC_KEY:
+        return "public";
+    case CKO_SECRET_KEY:
+        return "secret-key";
+    default:
```
ifranzki, line 113: How about the other CKO's:
#define CKO_HW_FEATURE        0x00000005
#define CKO_DOMAIN_PARAMETERS 0x00000006
#define CKO_PROFILE           0x00000009UL

usr/sbin/pkcsconf/pkcsconf.c
```diff
@@ -560,6 +561,22 @@ CK_RV check_user_and_group(void)
...
+    uri = p11_uri_new();
+    if (!uri)
+        return;
+
+    uri->info = info;
+
+    printf("\tURI: %s\n", p11_uri_format(uri));
+
+    p11_uri_free(uri);
+    uri = NULL;
```
ifranzki, line 577: not really required here, but OK (I sometimes also use this to make sure it can't be used afterwards).

usr/lib/common/buffer.c
holger-dengler: ok

usr/lib/common/buffer.c
holger-dengler: ok

usr/sbin/p11sak/p11sak.c
```diff
@@ -2537,7 +2616,44 @@ static CK_RV list_ckey(CK_SESSION_HANDLE session, p11sak_kt kt, int long_print)
...
+        uri = p11_uri_new();
+        if (!uri)
+            return CKR_HOST_MEMORY;
+
+        uri->info = &info;
+        uri->slot_id = slot;
+        uri->slot_info = &slot_info;
+        uri->token_info = &token_info;
+
+        if (CKR_OK == tok_attribute_alloc(session, hkey, uri->obj_class)) {
```
ifranzki, line 2634: we usually have it the other way around, i.e. `tok_attribute_alloc(session, hkey, uri->obj_class) == CKR_OK`.

usr/lib/common/buffer.c
holger-dengler: It is a printf() to a buffer. As you proposed, I will rename it to p11_buffer_append_printf().

usr/lib/common/buffer.c
holger-dengler: ok

usr/lib/common/buffer.c
holger-dengler: You're right. I will do a new implementation, which covers also overruns.

usr/lib/common/uri.c
holger-dengler: Ok, I will move c2t() to p11utils.c. It will also go into a separate commit.

usr/lib/common/uri.c
holger-dengler: You're right, if we move it to p11utils.c, we should also implement the other CKOs as well. I not covered them in my first implementation, because the RFC 7512 only mentions the implemented ones.

usr/sbin/p11sak/p11sak.c
holger-dengler: ok

usr/sbin/pkcsconf/pkcsconf.c
holger-dengler: I will keep it as is.

usr/lib/common/uri.c
holger-dengler: I would like to discuss another aspect: the c2t() function is very pkcs11-URI-specific, because the RFC 7512 defines the strings, which are returned by the function. If I would implement it in p11utils.c, I would have selected other strings. e.g.
CKO_CERTIFICATE --> "certificate" (instead of "cert")
CKO_PRIVATE_KEY --> "private-key" (instead of "private")
CKO_PUBLIC_KEY --> "public-key" (instead of "public")

Here is my proposal: I will leave the c2t() function in uri.c but rename it to rfc7512_get_cko(). I will also provide a p11_get_cko() function in p11utils.c.

ifranzki: yes that makes a lot of sense. 
If nobody is currently using p11_get_cko(), then you may even decide to not provide it in p11utils.c at the moment, and just have rfc7512_get_cko() in uri.c.

holger-dengler:
The new version of this PR contains:
- fixes for the travis failures
- coding style changes (missing spaces, remove of empty lines, ordering, etc)
- rewrite of the p11_buffer_append_printf() function
- function rename c2t() --> rfc7512_get_cko()

The branch has also been rebased to the current origin/master.

ifranzki:
You still have compile errors in Travis on the GCC builds:
```
In file included from /usr/include/string.h:495,
                 from usr/lib/common/buffer.c:14:
In function ‘strncat’,
    inlined from ‘p11_buffer_append_len’ at usr/lib/common/buffer.c:87:5,
    inlined from ‘p11_buffer_append’ at usr/lib/common/buffer.c:93:12:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:136:10: error: ‘__builtin_strncat’ output truncated before terminating nul copying as many bytes from a string as its length [-Werror=stringop-truncation]
  136 |   return __builtin___strncat_chk (__dest, __src, __len, __bos (__dest));
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
usr/lib/common/buffer.c: In function ‘p11_buffer_append’:
usr/lib/common/buffer.c:93:12: note: length computed here
   93 |     return p11_buffer_append_len(buf, s, strlen(s));
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

ifranzki:
Did you think about how to test the URI support in the CI? We should at a minimum call `pkcsconf -t` and `p11sak list` once during a CI run to at least run through that code.

For p11sak, you could enhance `testcases/misc_tests/p11sak_test.sh` to check the list output for the URIs to be as expected.

For pkcsconf, I guess an additional (shell script based) testcase similar to `p11sak_test.sh` would beneficial. You can call it in Makefile.am in target `ci-installcheck` similar to `p11sak_test.sh` so that it gets run during the CI run.

Please note that the slot configuration is different from the default when running in the CI.

holger-dengler:
I'm currently working on tests for the URI. The intention for the PR was, to start with the discussion about the changes as soon as possible. Maybe we can later also add an URI parser to the testcases. For the moment I'll just grep for the expected output.

holger-dengler:
Unfortunately I can not reproduce the build error on my machine, but I think the correct solution is:
```
-    size_t new_len = strlen(buf->b) + len;
+    size_t new_len = strlen(buf->b) + len + 1;
```

ifranzki:
Just give it a try and force push your fix to see if the warning is one.

holger-dengler:
The tests still fail with the same error. But I do not get why...
I concatenate two strings in p11_buffer_append_len(), buf->b and s. buf->b is \0 terminated, s may or may not, only ensured in cases, where it is called from p11_buffer_append().

The man page of strncat (3) says:
```
If src contains n or more bytes, strncat() writes n+1 bytes to dest (n from src plus the terminating null byte).  Therefore, the size of 
dest must be at least strlen(dest)+n+1.
```

So my previous proposed change is absolutely correct. new_len must be sizeof(b->buf) + len +1 (the latter for the \0 char).

But the error message says:
```
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:136:10: error: ‘__builtin_strncat’ output truncated before terminating nul copying as many bytes from a string as its length [-Werror=stringop-truncation]
  136 |   return __builtin___strncat_chk (__dest, __src, __len, __bos (__dest));
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
usr/lib/common/buffer.c: In function ‘p11_buffer_append’:
usr/lib/common/buffer.c:93:12: note: length computed here
   93 |     return p11_buffer_append_len(buf, s, strlen(s));
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
I don't get it... 

I checked string_fortified.h on my machine (gcc12) and it looks a bit different. __builtin___strncat_chk() is called with `__glibc_objsize (__dest)` instead of `__bos (__dest)`. Maybe we're hitting a gcc bug, I'll check this now.

If someone has an idea whats wrong, I would appreciate any help!

ifranzki:
In https://developers.redhat.com/blog/2018/05/24/detecting-string-truncation-with-gcc-8#forming_truncated_strings_with_strncat it states:

---------------------------------------------------

**Forming Truncated Strings with strncat**
String truncation can also occur as a result of calling strncat. The origins of strncat (and strncpy) can be traced to [Version 7 UNIX](https://en.wikipedia.org/wiki/Version_7_Unix), which was released in 1979 and in which functions were introduced to manipulate arrays of binary data not necessarily terminated by the NUL character, such as directory entries or encryption keys. Unlike the other functions discussed in this article, strncat is impossible to use safely even in the originally intended cases. To be used safely, the function would need to take as arguments not just the size of the remaining space in the destination but also the maximum number of characters to copy from a non-string. However, by providing only one size argument, it is impossible to avoid both buffer overflow and string truncation. Since preventing buffer overflow tends to be viewed as more important than preventing string truncation, GCC assumes the size argument refers to the remaining space in the destination buffer and expects safe calls to match the following pattern (see also the US-CERT article on [strncpy() and strncat()](https://www.us-cert.gov/bsi/articles/knowledge/coding-practices/strncpy-and-strncat)):

    `strncat (dest, src, dest_size - strlen (dest) - 1);`
Calls that have this form are not diagnosed. Other calls, such as those where the size is derived in some way from the size or length of the source string, are diagnosed by -Wstringop-overflow. That includes unsafe calls like

    `strncat (dest, src, strlen (src));   // incorrect - warning`
and

    strncat (dest, src, sizeof src);   // incorrect - warning

---------------------------------------------------------

Maybe the solution is simply to avoid strncat and use something like `memcpy(dest, src, MIN(len, strlen(src))` and add the terminating NUL byte manually afterwards.

holger-dengler:
I'll replace the `strncat()` with `memcpy()` with manual termination.

holger-dengler:
This next version contains the following changes:
- workaround for strncat() for append strings to a buffer
- new implementation of p11_buffer_append_printf() with dynamic memory allocation

This series contains all comments from the previous reviews.

New in this series:
- unit testcases for buffer and uri
- adjustments for p11sak test script
- p11sak URI tests in test script (WIP)
- pkcsconf URI tests in test script (WIP)

The testscripts are still WIP and haven't been executed in a real setup yet.

holger-dengler:
The series has also been rebased to current master.

testcases/misc_tests/p11sak_test.sh
```diff
@@ -517,6 +541,12 @@ echo "* TESTCASE list-key ec PASS Listed random ec public keys CK_BYTE attribute
 else
 echo "* TESTCASE list-key ec FAIL Failed to list ec public keys CK_BYTE attribute"
 fi
+# URI
+if [[ $(grep -A 85 'p11sak-ec-prime256v1:pub' $P11SAK_EC_LONG | grep -c 'URI: pkcs11:.*type=public') == "1" ]]; then
+echo "* TESTCASE list-key ec PASS list ec public key pkcs#11 URI"
+else
+echo "* TESTCASE list-key ec FAIL list ec public key pkcs#11 URI"
+fi
```
ifranzki, line 551: Is there a reason why no private keys are listed? 
No need to add them now for this PR if they did not exist jet, just wondering....

holger-dengler: One reason might be, that the p11sak listings lists multiple keys, but only the first one in the listing is selected for the test. It is hard in shell scripts to separate multiple multi-line objects from a logfile. The easy way is to pick only the first one. But this is only my guess.

Choosing perl or python for these testscripts would make that job much easier.

holger-dengler:
Series has been updated. Changes since v2:
- use `vasprintf()` instead of manual size allocation in `p11_buffer_append_printf()`.
- run testscripts on build machine and adopt the commits accordingly.

holger-dengler:
If there are no travis problems or other objections, this version could be the final one.

usr/lib/common/buffer.c
```diff
@@ -0,0 +1,128 @@
...
+    va_start(ap, fmt);
+    len = vasprintf(&tmp, fmt, ap);
+    va_end(ap);
+
+    if (len < 0)
+        goto err;
+
+    rc = p11_buffer_append_len(buf, tmp, len);
+err:
+    free(tmp);
```
ifranzki, line 126: if you go here because len from vasprintf is < 0, then tmp is uninitialized. This may cause free to operate on an invalid pointer. I would simply initialize tmp to NULL at the declaration.

testcases/misc_tests/pkcsconf_test.sh
```diff
@@ -0,0 +1,128 @@
+#!/bin/sh
```
ifranzki, line 1: Is this script called during CI test anywhere? 
I would suggest to call it in make taget ci-installcheck (Makefile.am) right after calling p11sak_tests.sh.

usr/lib/common/buffer.c
holger-dengler: I will fix this.

testcases/misc_tests/pkcsconf_test.sh
holger-dengler: I will add it.

holger-dengler: Would it make sense to do the 2 tests in a for loop?

ifranzki: for loop? I don't understand what you mean?

holger-dengler: maybe something like that...
```
ci-installcheck: ci-prepare installcheck
        killall -HUP pkcsslotd || true
        @sbindir@/pkcsslotd
        pushd ${srcdir}/testcases
        export PKCSLIB=@libdir@/opencryptoki/libopencryptoki.so
        export PKCS11_USER_PIN=$(PKCS11_USER_PIN)
        for misctest in p11sak pkcsslot ; do                            \
                ./misc_tests/${misctest}_test.sh                        \
                | tee log-${misctest}.txt                               \
        done
        popd
        @sbindir@/pkcsstats --all
        killall -HUP pkcsslotd
        @echo "done"
```

holger-dengler: ... versus 
```
ci-installcheck: ci-prepare installcheck
        killall -HUP pkcsslotd || true
        @sbindir@/pkcsslotd
        pushd ${srcdir}/testcases
        export PKCSLIB=@libdir@/opencryptoki/libopencryptoki.so
        export PKCS11_USER_PIN=$(PKCS11_USER_PIN)
        ./misc_tests/p11sak_test.sh | tee log-p11sak.txt
        ./misc_tests/pkcsconf_test.sh | tee log-pkcsconf.txt
        done
        popd
        @sbindir@/pkcsstats --all
        killall -HUP pkcsslotd
        @echo "done"
```

ifranzki: Well, with 2 tests I still think the second one is easier to read. Once we get more, then of course a for loop would make sense. Then maybe with placing the test to run in variable, like it is done in ock_tests.sh(.in).

BTW: In the second version the 'done' is superfluous.

holger-dengler: Ok, I'll continue with the second solution.

ifranzki (approved):
looks good now.

usr/lib/common/buffer.c
```diff
@@ -0,0 +1,128 @@
...
+
+#define BUFFER_INIT_SIZE	(128ul)
+
+struct _buffer {
+    size_t b_size;
+    char *b;
+};
+
+p11_buffer_t *p11_buffer_new(void)
+{
```
jschmidb, line 26: I'm sure this suits your needs, however for a general buffer implementation we could allow to pass an initial buffer size here. Might save a realloc in a 2nd step when it's already clear that the initial size is not enough.

usr/lib/common/uri.c
```diff
@@ -0,0 +1,270 @@
...
+            p11_buffer_append_len(buf, sep, 1);                         \
+            p11_buffer_append_printf(buf, txt "%lu", ul);               \
+            *sep = P11_URI_SEP_ATTR;                                    \
+        }                                                               \
+    } while(0)
+
+static const char HEX_CHARS_LOWER[] = "0123456789abcdef";
+
+static char *rfc7512_get_cko(CK_OBJECT_CLASS_PTR class)
+{
```
jschmidb, line 101: Hmm, yes, is correct, but I would code it as (CK_OBJECT_CLASS *class) so that we see the *class as it is used below in the switch.

usr/sbin/p11sak/p11sak.c
```diff
@@ -1622,6 +1623,58 @@ static CK_RV pub_key_print_attributes(CK_SESSION_HANDLE session,
...
+                                 CK_ATTRIBUTE_PTR attribute)
+{
+    CK_RV rv;
+    void *tmp;
+
+    if (!attribute)
+        return CKR_ARGUMENTS_BAD;
+
+    if ((attribute->pValue))
+        return CKR_CANCEL;
```
jschmidb, line 1642: Why double parentheses?

usr/lib/common/buffer.c
holger-dengler: I would like to implement it only when it is needed.

jschmidb (approved):
lgtm

usr/sbin/p11sak/p11sak.c
holger-dengler: Because ehhh... I have no idea. I'll change it.

usr/lib/common/buffer.c
```diff
@@ -0,0 +1,128 @@
...
+
+    if (ret->b == NULL)
+        goto out_free;
+
+    return ret;
+
+out_free:
+    free(ret);
+out:
+    return NULL;
```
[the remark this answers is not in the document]
jschmidb: Ops, just saw that we would save the out label too ...

usr/lib/common/uri.c
holger-dengler: What about using CK_PTR in the switch statement? It looks to me, that this might be the solution, how a PKCS#11-conform application/library should look like.

```
static char *rfc7512_get_cko(CK_OBJECT_CLASS_PTR class)
{
    switch (CK_PTR class) {
```

usr/lib/common/buffer.c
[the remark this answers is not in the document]
holger-dengler: I personally like the separation of the error- and good-path. But if your solution fits better to the rest of the code, I will change it. @ifranzki any preferences?

usr/lib/common/uri.c
holger-dengler: ok, so we will go with `static char *rfc7512_get_cko(CK_OBJECT_CLASS *class)`?

usr/lib/common/buffer.c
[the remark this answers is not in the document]
ifranzki: I don't really mind. OCK has all kind of different return styles.... :-)  
So you can keep it as is or change it, up to you.
Maybe naming the label `err` and `err_free` would make it clearer that this is the error path. `out` somehow sounds like the good path (or combined good and error path) to me.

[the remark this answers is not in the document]
holger-dengler: BTW: your proposed solution has still two returns.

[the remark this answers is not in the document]
holger-dengler: If there is no maintainer-overwrite, I will stay with my solution.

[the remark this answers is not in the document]
holger-dengler: Sorry, I missed your comment before I wrote my last one. I'll keep my solution with changed label names, as you proposed.

holger-dengler:
Another version with the requested changes.

PR #566, opened 2022-07-25
Misc fixes

ifranzki:
A few things that were identified during development work.

### DIFF
--- a/usr/lib/common/loadsave.c
+++ b/usr/lib/common/loadsave.c
@@ -2291,6 +2291,21 @@ static inline int inc32(unsigned char ctr[4])
 #define HEADER_LEN  64
 #define FOOTER_LEN  16
 
+/**
+ * public tok obj layout
+ *
+ * ----------------           <--+
+ * u32 tokversion                | 16-byte header
+ * u8  private_flag              |
+ * u8  reserved[7]               |
+ * u32 object_len                |
+ * ----------------           <--+
+ * u8  object[object_len]        | body
+ * ----------------           <--+
+ */
+#define PUB_HEADER_LEN     16
+#define HEADER_COMMON_LEN  5
+
 //
 // Note: The token lock (XProcLock) must be held when calling this function.
 //
@@ -2634,7 +2649,7 @@ CK_RV reload_token_object(STDLL_TokData_t *tokdata, OBJECT *obj)
 
     set_perm(fileno(fp));
 
-    if (fread(header, HEADER_LEN, 1, fp) != 1) {
+    if (fread(header, HEADER_COMMON_LEN, 1, fp) != 1) {
         OCK_SYSLOG(LOG_ERR, "Cannot read header\n");
         rc = CKR_FUNCTION_FAILED;
         goto done;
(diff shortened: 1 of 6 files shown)


vkanjur (approved):
Looks good to me.

usr/lib/common/loadsave.c
```diff
@@ -2634,15 +2648,31 @@ CK_RV reload_token_object(STDLL_TokData_t *tokdata, OBJECT *obj)
 
     set_perm(fileno(fp));
 
-    if (fread(header, HEADER_LEN, 1, fp) != 1) {
+    if (fread(header, 5, 1, fp) != 1) {
```
holger-dengler, line 2651: Please use a constant instead of 5, e.g. HEADER_COMMON_LEN. And maybe you can do the calculation in another set of constants, e.g.

```
#define HEADER_COMMON_LEN 5
[...]
#define HEADER_PART_LEN (HEADER_LEN + HEADER_PREFIX)
[...]
#define PUB_HEADER_PART_LEN (PUB_HEADER_LEN + HEADER_PREFIX)
```

usr/lib/common/loadsave.c
```diff
@@ -2634,15 +2648,31 @@ CK_RV reload_token_object(STDLL_TokData_t *tokdata, OBJECT *obj)
...
     memcpy(&priv, header + 4, 1);
-    memcpy(&len, header + 60, 4);
+    if (priv) {
+        if (fread(header + 5, HEADER_LEN - 5, 1, fp) != 1) {
+            OCK_SYSLOG(LOG_ERR, "Cannot read header\n");
+            rc = CKR_FUNCTION_FAILED;
+            goto done;
+        }
+
+        memcpy(&len, header + 60, 4);
```
holger-dengler, line 2666: Can we make this calculation a bit more self-explaining? Ok, 60 has been used before, but maybe you can also define a constant for this value as well.

usr/lib/common/loadsave.c
```diff
@@ -2634,15 +2648,31 @@ CK_RV reload_token_object(STDLL_TokData_t *tokdata, OBJECT *obj)
...
+    } else {
+        if (fread(header + 5, PUB_HEADER_LEN - 5, 1, fp) != 1) {
+            OCK_SYSLOG(LOG_ERR, "Cannot read header\n");
+            rc = CKR_FUNCTION_FAILED;
+            goto done;
+        }
+
+        memcpy(&len, header + 12, 4);
```
holger-dengler, line 2674: ...maybe here as well.

usr/lib/common/obj_mgr.c
```diff
@@ -1480,70 +1481,90 @@ CK_RV object_mgr_restore_obj_withSize(STDLL_TokData_t *tokdata, CK_BYTE *data,
     // The calling stack MUST have the mutex
     // to many grab it now.
 
+    obj = oldObj;
```
holger-dengler, line 1484: You can move it up to the declaration.

`OBJECT *obj = oldObj;`

usr/lib/common/obj_mgr.c
```diff
@@ -1877,11 +1885,34 @@ CK_RV object_mgr_check_shm(STDLL_TokData_t *tokdata, OBJECT *obj)
...
+        *entry = &tokdata->global_shm->publ_tok_objs[index];
+    }
+
+    return CKR_OK;
+}
+
+
+// The object must hold the READ lock when this function is called!
+//
+CK_RV object_mgr_check_shm(STDLL_TokData_t *tokdata, OBJECT *obj)
```
holger-dengler, line 1899: This locking is a bit confusing. I'll skip it for the review and try to understand it later...

holger-dengler (commented):
Looks good so far.

usr/lib/common/loadsave.c
ifranzki: introduced HEADER_COMMON_LEN define.

usr/lib/common/loadsave.c
ifranzki: it would certainly make sense to use defines here  (or even better a struct and offsetof), but these offsets of the fields are used at many places within this source file, and we would need to change all of these places then. I would not want to do that now, but will put it on my list if things to do at a later time.

jschmidb (approved):
Looks good as far as I can see.

usr/lib/common/loadsave.c
```diff
@@ -2634,15 +2649,33 @@ CK_RV reload_token_object(STDLL_TokData_t *tokdata, OBJECT *obj)
 
     set_perm(fileno(fp));
 
-    if (fread(header, HEADER_LEN, 1, fp) != 1) {
+    if (fread(header, HEADER_COMMON_LEN, 1, fp) != 1) {
```
ifranzki, line 2652: Just realized that the same fix also needs to go into reload_token_object_old().... Will update the commit asap.

ifranzki: Ah, no, reload_token_object_old() already has the logic to load public or private objects. The old data format uses the same header for both, so no further fix is needed.

PR #567, opened 2022-07-26
EP11: Fix C_GetMechanismList returning CKR_BUFFER_TOO_SMALL

ifranzki:
For mixed card levels, the size query call and the call to obtain the list may run on different cards. When the size query call runs on a card with less mechanisms than the second call, will fail, but it returns the larger larger number of mechanisms.

The code already re-allocates the buffer for retrieving the mechanism list, but does not return the larger number in pulCount. This will lead to a CKR_BUFFER_TOO_SMALL when the application calls C_GetMechanismList
again to obtain the list of mechanisms, because the applications buffer is too small.


chrisRund (approved):
Reproduced the problem without this change. Verified this commit is actually fixing the issue - problem not longer seen.

PR #569, opened 2022-08-02
travis: run 'make check' during travis execution.

ifranzki:
'make check' is currently not run in the CI, so lets run it at least in travis.


holger-dengler (approved):
Looks good to me.

PR #572, opened 2022-08-16
p11sak fixes

ifranzki:
- remove leftover debug printf.
- skip non-key objects during list-keys and remove-key, don't print ugly errors.

### DIFF
--- a/usr/sbin/p11sak/p11sak.c
+++ b/usr/sbin/p11sak/p11sak.c
@@ -1768,10 +1768,8 @@ static CK_RV tok_key_get_key_type(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE hk
         strcpy(buffer, "private ");
         break;
     default:
-        // FIXIT - return code to represent object class invalid
-        rc = CKR_KEY_HANDLE_INVALID;
-        fprintf(stderr, "Object handle invalid (error code 0x%lX: %s)\n",
-               rc, p11_get_ckr(rc));
+        /* its not a key */
+        rc = CKR_KEY_TYPE_INCONSISTENT;
         free(buffer);
         return rc;
     }
@@ -2604,8 +2602,10 @@ static CK_RV list_ckey(CK_SESSION_HANDLE session, CK_SLOT_ID slot,
         rc = tok_key_get_key_type(session, hkey, &keyclass, &keytype,
                 &keylength);
         if (rc != CKR_OK) {
-            fprintf(stderr, "Invalid key type (error code 0x%lX: %s)\n", rc,
-                    p11_get_ckr(rc));
+            if (rc != CKR_KEY_TYPE_INCONSISTENT)
+                fprintf(stderr,
+                        "Retrieval of key type failed (error code 0x%lX: %s)\n",
+                        rc, p11_get_ckr(rc));
             continue;
         }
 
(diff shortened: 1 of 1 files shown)


usr/sbin/p11sak/p11sak.c
```diff
@@ -2604,8 +2602,10 @@ static CK_RV list_ckey(CK_SESSION_HANDLE session, CK_SLOT_ID slot,
         rc = tok_key_get_key_type(session, hkey, &keyclass, &keytype,
                 &keylength);
         if (rc != CKR_OK) {
-            fprintf(stderr, "Invalid key type (error code 0x%lX: %s)\n", rc,
-                    p11_get_ckr(rc));
+            if (rc != CKR_KEY_TYPE_INCONSISTENT)
+                fprintf(stderr,
+                        "Retrieval of key type failed (error code 0x%lX: %s)\n",
+                        rc, p11_get_ckr(rc));
```
jschmidb, line 2608: We could print out the related key type in the error msg to give a hint into what the user's input translated and caused the error

ifranzki: I don't know the key type here. The error CKR_KEY_TYPE_INCONSISTENT is just returned as a hint by tok_key_get_key_type that its not a key object at all.
We could choose another error code if that fits better....

jschmidb: Ah, I see .... I thought this is the p11sak_kt. The error code is fine with me.

jschmidb (approved):

FinnCallies (approved):

usr/sbin/p11sak/p11sak.c
```diff
@@ -931,12 +931,10 @@ static CK_RV set_battr(const char *attr_string, CK_ATTRIBUTE *attr,
         for (i = 0; i < (int) strlen(attr_string); i++) {
             // attr_string length is checked in parse_gen_key_args to avoid memory problems
             if (prv == 1 && attr_na(char2attrtype(toupper(attr_string[i])), kt_PRIVATE) == 0) {
-                printf("private: %d\n", attr_string[i]);
```
finn-callies, line 934: There was a comment in the commit I did (if i remember correctly), suggesting to leave these logs in there, to tell the user, which attributes (of the users input attributes) are actually used for the key creation and which might have been skipped, because of whatever reason.

usr/sbin/p11sak/p11sak.c
```diff
@@ -931,12 +931,10 @@ static CK_RV set_battr(const char *attr_string, CK_ATTRIBUTE *attr,
...
                 set_bool_attr_from_string(&attr[*count], attr_string[i]);
                 (*count)++;
             } 
             if (prv == 0 && attr_na(char2attrtype(toupper(attr_string[i])), kt_PUBLIC) == 0) {
-                printf("public: %d\n", attr_string[i]);
```
finn-callies, line 939: Same as above

finn-callies (approved):

usr/sbin/p11sak/p11sak.c
ifranzki: I don't think that the user understands the output as printed. Its internal information. If really required, then we need to print this in a more user friendly way. I would say lets remove it for now, and if it turns out that this information is really needed, then implement it in a user friendly way.

PR #575, opened 2022-08-17
gitignore: update generated files to ignore

ifranzki:
Tools pkcsep11_migrate and pkcsep11_session must als be ignored.


jschmidb:
Commit msg: ... must als be ignored.  <- also

ifranzki:
> Commit msg: ... must als be ignored. <- also

fixed.

PR #576, opened 2022-08-17
Add pkg-config file for opencryptoki

ifranzki:


configure.ac
```diff
@@ -682,6 +682,7 @@ AC_LANG_POP([C++])
 AC_CONFIG_MACRO_DIRS([m4])
 
 AC_CONFIG_FILES([Makefile				\
+		 opencryptoki.pc			\
```
juergenchrist, line 685: Doesn't work with `make prefix=/foo/bar` which is allowed by autotools.

juergenchrist (approved):

configure.ac
ifranzki: So what would be the solution (if any)?

jschmidb:
- Yes, the usual location of the pkg-config file is ${libdir}/pkgconfig. What I learned with libzpc is: It could also be /usr/share/pkgconfig or /usr/{lib,lib64}/pkgconfig.
- Question: why are you using different notations:  @libdir@ vs. ${libdir} ?
- In libzpc the pkg-config file is also referenced in the spec file: refer to https://github.com/opencryptoki/libzpc/blob/main/libzpc.spec, it's under %files devel

configure.ac
juergenchrist: Either document that this is not supported or generate the .pc-file during make.  Not sure if anybody uses this feature.  Never found a use case myself and always wondered why I can configure a prefix directory and the decide during make to use a different one.  But yes, all variables (`prefix`, `libdir`, ...) can be overwritten when invoking `make`.  But to make things even worse, I think you can even overwrite them when doing a `make install`...

Summing up, I think documenting that his variable overwriting during make is not supported might be the way to go.

ifranzki:
> Question: why are you using different notations: @libdir@ vs. ${libdir} ?

@variable@ is replaced by autoconf when processing the *.in files during configure step.
${variable} is a variable within the pkgconfig file and is resolved by pkg-config at runtime.

ifranzki:
> In libzpc the pkg-config file is also referenced in the spec file

Good point, I will add it to the spec file as well (although the spec file is already very much out of date and does possibly not work anyway....)

jschmidb (approved):
ok, looks good

configure.ac
ifranzki: Created issue https://github.com/opencryptoki/opencryptoki/issues/577 so that we don't forget about this.

Issue #538, opened 2022-04-20
p11sak: support Edwards curves


ifranzki:
ed/x25519 and ed/x448 are currently not supported.
The EP11 token does support them.

PR #580, opened 2022-08-17
p11sak: Add support for Edwards and Montgomery curves

ifranzki:
Add support for generating EC keys with curves Curves curve25519, curve448, ed25519, and ed448.

Currently only the EP11 token supports those curves. For other tokens error CKR_CURVE_NOT_SUPPORTED is returned.

Closes: https://github.com/opencryptoki/opencryptoki/issues/538


jschmidb:
Yes, makes sense. 

- One typo in commit msg: Add ... with curves Curves curve25519 ...  <- there's one 'curves' too much 
- Regarding the naming: sometimes the Montgomery curves are named X25519/X448, sometimes curve25519/curve448 as you are choosing. I think we should try to keep the naming consistent within our libs and e.g. libica already uses the 'X' notation. Is there a special reason that you choose 'curve'?

ifranzki:
> Is there a special reason that you choose 'curve'?

Well, I guess because the EP11 structure document talks about curve25519/curve448.
They also refer to RFC 7748, where you see both curve25519/curve448 and X25519/X448 mentioned:
https://www.rfc-editor.org/rfc/rfc7748.html#section-4
https://www.rfc-editor.org/rfc/rfc7748.html#section-5

The RFC actually seem to differentiate between the curve as such and the X25519 and X448 as underlying DH functions.
So from that I would say the name of the curve is really curve25519/curve448 and not X25519/X448....

The DH operation using those curves is then called X25519/X448. This matches with the corresponding mechanism to perform a DH operation with EP11: CKM_IBM_EC_X25519/CKM_IBM_EC_X448

The OID registry also knows these OIDs as curve25519/curve448, but also mentions the X names:
https://oid-rep.orange-labs.fr/get/1.3.101.110
https://oid-rep.orange-labs.fr/get/1.3.101.111

So I would tend to keep it as is. 

But we might want to accept the names X25519/X448 as synonymes at the keygen command? Would we also document this then in the man page and help text? Could be rather confusing though....
@jschmidb what do you think?

ifranzki:
> One typo in commit msg: Add ... with curves Curves curve25519 ... <- there's one 'curves' too much

fixed.

jschmidb:
> @jschmidb what do you think?

Good question ;-) yes, this is really confusing. As you say, it looks like the curves themselves are called curve25519/curve448, but specific usages, like EdDH, use X25519/X448. Openssl uses nid constants NID_X25519 and NID_X448 to identify curves. OK, so let's think about what p11sak is actually doing: it creates keys for a specific usage in opencryptoki. For Edwards, the usage is EdDSA, for Montgomery it's EdDH. So we could say that an X25519 key is used for EdDH or, as named in libica, X25519_derive / X448_derive. Perhaps it was this reason why libica is using the 'X' notation.

ifranzki:
I guess I leave it as is then. p11sak is creating keys, and you specify the curve (which is Curve...) to used. p11sak does not care about the functions (which would be X....) used with these keys later on. 

Well yes, for ed25519/ed48 it is confusing, since it denotes both the curve name and the function name....

jschmidb:
OK, both is fine with me.

jschmidb (approved):

PR #581, opened 2022-08-19
p11sak fixes

ifranzki:
Todays CI run showed some errors caused by p11sak.


jschmidb:
Looks good to me. 
Just one question on listing keys by label: you mentioned that p11sak has a new function to list keys by label. I don't see any test for this. Did I oversee it?

ifranzki:
>  I don't see any test for this. Did I oversee it?

You looked correctly. There is currently no test for this. Maybe we should add one....

jschmidb (approved):

PR #582, opened 2022-08-19
Add support for dual-function cryptographic functions

ifranzki:
- Add support for functions C_DigestEncryptUpdate, C_DecryptDigestUpdate,
C_SignEncryptUpdate, and C_DecryptVerifyUpdate
- Fix C_Get/SetOperationState with multiple operations active
- Add a testcase for dual-function cryptographic functions

FYI: 
- For the first patch you only need to review one of the new_host.c file updates. The changes in the other 2 new_host.c files are exactly the same.
- For the second patch you only need to review the changes in lock_sess_mgr.c, the updates in sess_mgr.c are exactly the same.
- These semi-duplicated files are required because the files are not 100% the same, but the places where the updates of these commits are done are 100% the same.

### DIFF
--- a/usr/lib/common/lock_sess_mgr.c
+++ b/usr/lib/common/lock_sess_mgr.c
@@ -586,12 +586,14 @@ CK_RV session_mgr_logout_all(STDLL_TokData_t *tokdata)
 
 //
 //
-CK_RV session_mgr_get_op_state(SESSION *sess,
+CK_RV session_mgr_get_op_state(STDLL_TokData_t *tokdata, SESSION *sess,
                                CK_BBOOL length_only,
                                CK_BYTE *data, CK_ULONG *data_len)
 {
     OP_STATE_DATA *op_data = NULL;
-    CK_ULONG op_data_len = 0;
+    CK_ULONG max_data_len = *data_len;
+    CK_ULONG op_data_len;
+    CK_ULONG all_data_len = 0;
     CK_ULONG offset, active_ops;
 
     if (!sess) {
@@ -614,22 +616,29 @@ CK_RV session_mgr_get_op_state(SESSION *sess,
             return CKR_STATE_UNSAVEABLE;
         }
         active_ops++;
-        if (op_data != NULL) {
-            TRACE_ERROR("%s\n", ock_err(ERR_STATE_UNSAVEABLE));
-            return CKR_STATE_UNSAVEABLE;
-        }
         op_data_len = sizeof(OP_STATE_DATA) +
             sizeof(ENCR_DECR_CONTEXT) +
             sess->encr_ctx.context_len + sess->encr_ctx.mech.ulParameterLen;
+        all_data_len += op_data_len;
 
         if (length_only == FALSE) {
             op_data = (OP_STATE_DATA *) data;
 
-            if (*data_len < op_data_len) {
+            if (max_data_len < op_data_len) {
                 TRACE_ERROR("%s\n", ock_err(ERR_BUFFER_TOO_SMALL));
                  return CKR_BUFFER_TOO_SMALL;
             }
 
+            memset(op_data, 0, sizeof(*op_data));
+#ifdef PACKAGE_VERSION
+            strncpy((char *)op_data->library_version, PACKAGE_VERSION,
+                    sizeof(op_data->library_version));
+#endif
+            memcpy(op_data->manufacturerID,
+                   tokdata->nv_token_data->token_info.manufacturerID,
+                   sizeof(op_data->manufacturerID));
+            memcpy(op_data->model, tokdata->nv_token_data->token_info.model,
+                   sizeof(op_data->model));
             op_data->data_len = op_data_len - sizeof(OP_STATE_DATA);
             op_data->session_state = sess->session_info.state;
             op_data->active_operation = STATE_ENCR;
@@ -653,6 +662,9 @@ CK_RV session_mgr_get_op_state(SESSION *sess,
                        sess->encr_ctx.mech.pParameter,
                        sess->encr_ctx.mech.ulParameterLen);
             }
+
+            max_data_len -= op_data_len;
+            data += op_data_len;
         }
     }
 
@@ -662,22 +674,29 @@ CK_RV session_mgr_get_op_state(SESSION *sess,
             return CKR_STATE_UNSAVEABLE;
         }
         active_ops++;
-        if (op_data != NULL) {
-            TRACE_ERROR("%s\n", ock_err(ERR_STATE_UNSAVEABLE));
-            return CKR_STATE_UNSAVEABLE;
-        }
         op_data_len = sizeof(OP_STATE_DATA) +
             sizeof(ENCR_DECR_CONTEXT) +
             sess->decr_ctx.context_len + sess->decr_ctx.mech.ulParameterLen;
+        all_data_len += op_data_len;
 
         if (length_only == FALSE) {
             op_data = (OP_STATE_DATA *) data;
 
-            if (*data_len < op_data_len) {
+            if (max_data_len < op_data_len) {
                 TRACE_ERROR("%s\n", ock_err(ERR_BUFFER_TOO_SMALL));
                  return CKR_BUFFER_TOO_SMALL;
             }
 
+            memset(op_data, 0, sizeof(*op_data));
+#ifdef PACKAGE_VERSION
+            strncpy((char *)op_data->library_version, PACKAGE_VERSION,
+                    sizeof(op_data->library_version));
+#endif
+            memcpy(op_data->manufacturerID,
+                   tokdata->nv_token_data->token_info.manufacturerID,
+                   sizeof(op_data->manufacturerID));
+            memcpy(op_data->model, tokdata->nv_token_data->token_info.model,
+                   sizeof(op_data->model));
             op_data->data_len = op_data_len - sizeof(OP_STATE_DATA);
             op_data->session_state = sess->session_info.state;
             op_data->active_operation = STATE_DECR;
@@ -701,6 +720,9 @@ CK_RV session_mgr_get_op_state(SESSION *sess,
                        sess->decr_ctx.mech.pParameter,
                        sess->decr_ctx.mech.ulParameterLen);
             }
+
+            max_data_len -= op_data_len;
+            data += op_data_len;
         }
     }
 
@@ -710,22 +732,29 @@ CK_RV session_mgr_get_op_state(SESSION *sess,
             return CKR_STATE_UNSAVEABLE;
         }
         active_ops++;
-        if (op_data != NULL) {
-            TRACE_ERROR("%s\n", ock_err(ERR_STATE_UNSAVEABLE));
-            return CKR_STATE_UNSAVEABLE;
-        }
         op_data_len = sizeof(OP_STATE_DATA) +
             sizeof(DIGEST_CONTEXT) +
             sess->digest_ctx.context_len + sess->digest_ctx.mech.ulParameterLen;
+        all_data_len += op_data_len;
 
         if (length_only == FALSE) {
             op_data = (OP_STATE_DATA *) data;
 
-            if (*data_len < op_data_len) {
+            if (max_data_len < op_data_len) {
                 TRACE_ERROR("%s\n", ock_err(ERR_BUFFER_TOO_SMALL));
                  return CKR_BUFFER_TOO_SMALL;
             }
 
+            memset(op_data, 0, sizeof(*op_data));
+#ifdef PACKAGE_VERSION
+            strncpy((char *)op_data->library_version, PACKAGE_VERSION,
+                    sizeof(op_data->library_version));
+#endif
+            memcpy(op_data->manufacturerID,
+                   tokdata->nv_token_data->token_info.manufacturerID,
+                   sizeof(op_data->manufacturerID));
+            memcpy(op_data->model, tokdata->nv_token_data->token_info.model,
+                   sizeof(op_data->model));
             op_data->data_len = op_data_len - sizeof(OP_STATE_DATA);
             op_data->session_state = sess->session_info.state;
             op_data->active_operation = STATE_DIGEST;
@@ -749,6 +778,9 @@ CK_RV session_mgr_get_op_state(SESSION *sess,
                        sess->digest_ctx.mech.pParameter,
                        sess->digest_ctx.mech.ulParameterLen);
             }
+
+            max_data_len -= op_data_len;
+            data += op_data_len;
         }
     }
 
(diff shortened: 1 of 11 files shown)


usr/lib/common/new_host.c
```diff
@@ -3380,20 +3380,17 @@ CK_RV SC_DigestEncryptUpdate(STDLL_TokData_t *tokdata,
...
+    CK_RV rc;
 
-    if (tokdata->initialized == FALSE) {
-        TRACE_ERROR("%s\n", ock_err(ERR_CRYPTOKI_NOT_INITIALIZED));
-        return CKR_CRYPTOKI_NOT_INITIALIZED;
-    }
+    rc = SC_EncryptUpdate(tokdata, sSession, pPart, ulPartLen,
+                          pEncryptedPart, pulEncryptedPartLen);
+    if (rc != CKR_OK || pEncryptedPart == NULL)
+        goto out;
```
jschmidb, line 3388: What if pEncryptedPart != NULL but ulPartLen == 0?

usr/lib/common/new_host.c
```diff
@@ -3380,20 +3380,17 @@ CK_RV SC_DigestEncryptUpdate(STDLL_TokData_t *tokdata,
                              CK_ULONG ulPartLen, CK_BYTE_PTR pEncryptedPart,
                              CK_ULONG_PTR pulEncryptedPartLen)
 {
-    UNUSED(sSession);
-    UNUSED(pPart);
-    UNUSED(ulPartLen);
-    UNUSED(pEncryptedPart);
-    UNUSED(pulEncryptedPartLen);
+    CK_RV rc;
```
jschmidb, line 3384: Hmm, should we check if pulEncryptedPartLen == NULL? Or would SC_EncryptUpdate handle this?

usr/lib/common/new_host.c
ifranzki: Then C_DigestEncryptUpdate was called with a zero length update. That is a valid case. Adding zero bytes to a digest is allowed (but will of course not update the digest).

usr/lib/common/new_host.c
ifranzki: I would leave this to SC_EncryptUpdate, which does check for NULL.

usr/lib/common/lock_sess_mgr.c
```diff
@@ -586,12 +586,14 @@ CK_RV session_mgr_logout_all(STDLL_TokData_t *tokdata)
...
 //
-CK_RV session_mgr_get_op_state(SESSION *sess,
+CK_RV session_mgr_get_op_state(STDLL_TokData_t *tokdata, SESSION *sess,
                                CK_BBOOL length_only,
                                CK_BYTE *data, CK_ULONG *data_len)
 {
     OP_STATE_DATA *op_data = NULL;
-    CK_ULONG op_data_len = 0;
+    CK_ULONG max_data_len = *data_len;
+    CK_ULONG op_data_len;
```
jschmidb, line 595: Can it happen that other input parms than sess, e.g. data_len, are NULL?

usr/lib/common/lock_sess_mgr.c
```diff
@@ -865,233 +917,338 @@ CK_RV session_mgr_set_op_state(STDLL_TokData_t *tokdata, SESSION *sess,
...
     CK_BYTE *context = NULL;
     CK_BYTE *ptr1 = NULL;
     CK_BYTE *ptr2 = NULL;
     CK_BYTE *ptr3 = NULL;
     CK_ULONG len;
+    CK_ULONG encr_key_needed = 0;
+    CK_ULONG auth_key_needed = 0;
 
     if (!sess || !data) {
         TRACE_ERROR("%s received bad argument(s)\n", __func__);
```
jschmidb, line 933: Can it happen that data != NULL but data_len == 0?

usr/lib/common/lock_sess_mgr.c
ifranzki: This is already checked in SC_GetOperationState.

usr/lib/common/lock_sess_mgr.c
ifranzki: This is already checked in SC_SetOperationState.
And even if data_len would be zero, then the while loop below would just be skipped and the function would essentially do nothing.

usr/lib/common/lock_sess_mgr.c
```diff
@@ -865,233 +917,338 @@ CK_RV session_mgr_set_op_state(STDLL_TokData_t *tokdata, SESSION *sess,
...
+                    TRACE_ERROR("%s\n", ock_err(ERR_SAVED_STATE_INVALID));
+                    return CKR_SAVED_STATE_INVALID;
+                }
+                ptr1 = (CK_BYTE *) ctx;
+                ptr2 = ptr1 + sizeof(DIGEST_CONTEXT);
+                ptr3 = ptr2 + ctx->context_len;
+
+                if (ctx->context_len) {
+                    context = (CK_BYTE *) malloc(ctx->context_len);
+                    if (!context) {
```
jschmidb, line 1180: Where is 'context' freed in the non-error case? After the 'switch' context is set to NULL, but not freed?

ifranzki: The newly allocated context is set into sess->xxx_ctx->context in the switch and thus belongs now to the session. Between allocation and the place when it is sett into the session it can't fail. Once it is set into the session, it will be freed when the operation is terminated (e.g. by a Final call, or when the session is closed).
 
So if setting another operation fails after one has been set successfully, the successful one is still set in the session, and will thus be freed when the (first) operation is terminated.

We could argue that if setting one operations fails (e.g. the second one), then all previously set operations should also be terminated. This could be done. However, the second while loop can only fail due to allocation failures (out of memory). All other possible errors have already been checked in the first while loop. The second while loop would not run at all if the first one detected an error.

usr/lib/common/lock_sess_mgr.c
```diff
@@ -865,233 +917,338 @@ CK_RV session_mgr_set_op_state(STDLL_TokData_t *tokdata, SESSION *sess,
...
+                    memcpy(mech_param, ptr3, ctx->mech.ulParameterLen);
+                }
+            }
+            break;
+        default:
+            TRACE_ERROR("%s\n", ock_err(ERR_SAVED_STATE_INVALID));
+            return CKR_SAVED_STATE_INVALID;
+        }
+
+        /* clear current state (if active) and copy the new state information */
```
ifranzki, line 1204: This comment is slightly wrong. The current state has already been cleared above. So I'll remove the first first part of the comment.

usr/lib/common/new_host.c
```diff
@@ -3380,20 +3380,17 @@ CK_RV SC_DigestEncryptUpdate(STDLL_TokData_t *tokdata,
...
 
-    if (tokdata->initialized == FALSE) {
-        TRACE_ERROR("%s\n", ock_err(ERR_CRYPTOKI_NOT_INITIALIZED));
-        return CKR_CRYPTOKI_NOT_INITIALIZED;
-    }
+    rc = SC_EncryptUpdate(tokdata, sSession, pPart, ulPartLen,
```
holger-dengler, line 3385: I don't see the code path, where `SC_EncryptUpdate()` returns with `rc == CKR_OK` so that the second check (`pEncryptedPart == NULL`) will be true. Can you please give an example, in which situations this is the case?

usr/lib/common/new_host.c
```diff
@@ -3380,20 +3380,17 @@ CK_RV SC_DigestEncryptUpdate(STDLL_TokData_t *tokdata,
...
+                          pEncryptedPart, pulEncryptedPartLen);
+    if (rc != CKR_OK || pEncryptedPart == NULL)
```
holger-dengler, line 3387: In my opinion, the first check is sufficient. But maybe I'm wrong.

usr/lib/common/new_host.c
ifranzki: This is the case for a size-quer call. You call SC_EncryptUpdate with pEncryptedPart = NULL but pulEncryptedPartLen != NULL. This indicates that you want to know the required size of the output buffer.
The same can be used for SC_DigestEncryptUpdate. So the code must cover these size-query calls.
Note that the digest update must only happen when the encrypt step actually has produced output. 
PKCS#11:
_C_DigestEncryptUpdate uses the convention described in Section 5.2 on producing output. If a
C_DigestEncryptUpdate call does not produce encrypted output (because an error occurs, or because
pEncryptedPart has the value NULL_PTR, or because pulEncryptedPartLen is too small to hold the entire
encrypted part output), then no plaintext is passed to the active digest operation._

usr/lib/common/new_host.c
ifranzki: A size-query call returns CKR_OK, but did not produce any encryption output. So it must not call DigestUpdate in that case.

holger-dengler: OK, thanks.

usr/lib/common/sess_mgr.c
```diff
@@ -566,22 +568,29 @@ CK_RV session_mgr_get_op_state(SESSION *sess,
...
-            if (*data_len < op_data_len) {
+            if (max_data_len < op_data_len) {
                 TRACE_ERROR("%s\n", ock_err(ERR_BUFFER_TOO_SMALL));
                  return CKR_BUFFER_TOO_SMALL;
             }
 
+            memset(op_data, 0, sizeof(*op_data));
+#ifdef PACKAGE_VERSION
+            strncpy((char *)op_data->library_version, PACKAGE_VERSION,
+                    sizeof(op_data->library_version));
```
holger-dengler, line 587: If the size of `PACKAGE_VERSION` is greater than 16 (`sizeof(op_data->library_version)`), the string in `op_data->library_version` will not be \0 terminated.

Change the string copy here and in all other places to:
```
strncpy((char *)op_data->library_version, PACKAGE_VERSION, sizeof(op_data->library_version) - 1);
```
It is not required to explicitly set the terminating char, because the struct has been initialized in the line before.

ifranzki: Well yes, I also thought about this, but it actually does not matter if it is zero terminated or not. 
In `session_mgr_set_op_state()` it is compared using `strncmp`, so zero termination does not matter:
```
if (strncmp((char *)op_data->library_version, PACKAGE_VERSION,
                sizeof(op_data->library_version)) != 0) {
```
With your suggestion, I would also have to change the strncmp() call to use `sizeof(...) - 1` I guess, otherwise it would not match if it has been truncated.

holger-dengler: `strncmp()` skip the termination character implicitly. There is no need to change these calls.

holger-dengler: Here is an example:
```
$ cat xxx.c
#include <string.h>
#include <stdlib.h>
#include <stdio.h>

int main(void)
{
        char dest[16];
        char *src_short = "abc";
        char *src_long = "abcdefghijklmnopqrstuvwxyz";

        memset(dest, 0, sizeof(dest));
        printf("%s %ld %ld\n", dest, sizeof(dest), strlen(dest));

        strncpy(dest, src_short, sizeof(dest));
        printf("%s %ld %ld %d\n", dest, sizeof(dest), strlen(dest),
               strncmp(dest, src_short, sizeof(dest)));

        strncpy(dest, src_long, sizeof(dest));
        printf("%s %ld %ld %d\n", dest, sizeof(dest), strlen(dest),
               strncmp(dest, src_long, sizeof(dest)));

        return 0;
}
$ gcc -O0 -g -Wextra -Wall -o xxx xxx.c
$ ./xxx
 16 0
abc 16 3 0
abcdefghijklmno�д�U 16 22 0
```

ifranzki: right, but lets assume `PACKAGE_VERSION` is a string of 20 characters. With `op_data->library_version` being truncated to 15 characters (plus 1 zero byte), then the 2 strings will simply not match if `strncmp` is used with a length of 16.

holger-dengler: They do match, as you can see in my example.

ifranzki: But you are using `sizeof(dest)` (which is what I do in my code), not `sizeof(dest) - 1` as you suggested.

holger-dengler: Ah, sorry. They don't. My fault.

ifranzki: so bottom line, I would tend to leave it as is, but maybe add a comment to it that zero termination doesn't matter here.

holger-dengler: But anyhow, please descide, if `op_data->library_version` is holding a string or not. If it is defined as a string, if must have a \0 termination. Otherwise I would prefer to use `memcpy()` and `memcmp()` instead of the string functions.

ifranzki: well, technically ist a CK_CHAR array.... not sure if that makes it a C-string.
The other 2 fields manufacturerID and model are also CK_CHAR arrays and are not zero terminated (but blank padded)....

With memcpy, I would need to copy the C-string PACKAGE_VERSION to it but have to care about its length myself (strlen(PACKAGE_VERSION) - but what of its length is larger then 16....).
Well this is what strncpy is for. Why reinvent the wheel?

I will add a comment to the declaration and leave the code as is.:
`CK_CHAR library_version[16]; /* zero termination does not matter here */`

holger-dengler: ok, fine with me. Let's keep fingers crossed, that it won't hurt us in the future.

ifranzki: Well, PACKAGE_VERSION contains "3.18.0" as of today. Imagine a real live version number that's larger than 15 characters.... ;-)

usr/lib/common/lock_sess_mgr.c
jschmidb: ok, sounds reasonable.

jschmidb (approved):

usr/lib/common/lock_sess_mgr.c
```diff
@@ -614,22 +616,29 @@ CK_RV session_mgr_get_op_state(SESSION *sess,
...
+            memset(op_data, 0, sizeof(*op_data));
+#ifdef PACKAGE_VERSION
+            strncpy((char *)op_data->library_version, PACKAGE_VERSION,
+                    sizeof(op_data->library_version));
+#endif
+            memcpy(op_data->manufacturerID,
+                   tokdata->nv_token_data->token_info.manufacturerID,
+                   sizeof(op_data->manufacturerID));
+            memcpy(op_data->model, tokdata->nv_token_data->token_info.model,
+                   sizeof(op_data->model));
```
holger-dengler, line 641: These lines are copied 3 times. Wouldn't it make sense to make a function for it?

holger-dengler (approved):
The second commit is hard to understand and review. I personally prefer more but smaller commits. Anyhow, the PR looks good so far.

usr/lib/common/lock_sess_mgr.c
ifranzki: Yes actually the whole code around these lines is pretty much the same all 5 times... It just differs slightly by the context type it operates on.... So maybe the code should be refactored to have that lines only once, either in a macro or in a function.

I put this on my TODO list for a later PR.....

PR #583, opened 2022-08-22
doc: Document that 'make prefix=/foo/bar' or similar is not supported

ifranzki:
Variables like 'prefix', 'libdir', or similar should not be specified with 'make', but should be specified with 'configure' instead. Specifying them with 'make' is not supported by the openCryptoki package and may produce unexpected results!

Suggested-by: @juergenchrist

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -96,6 +96,11 @@ find the X include and library files automatically, but if it doesn't,
 you can use the `configure' options `--x-includes=DIR' and
 `--x-libraries=DIR' to specify their locations.
 
+   Note: Do not specify 'prefix=/foo/bar', 'libdir=/foo/bar' with
+the 'make' invocation. Specify them with 'configure' instead. Specifying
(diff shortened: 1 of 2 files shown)


juergenchrist (approved):

INSTALL
```diff
@@ -96,6 +96,11 @@ find the X include and library files automatically, but if it doesn't,
 you can use the `configure' options `--x-includes=DIR' and
 `--x-libraries=DIR' to specify their locations.
 
+   Note: Do not specify 'prefix=/foo/bar', 'libidir=/foo/bar' with
+the 'make' invocation. Specify them with 'configure' instead. Specifying
```
jschmidb, line 100: Do you mean 'libdir=/foo/bar' ?

ifranzki: yes, fixed.

jschmidb (approved):

PR #584, opened 2022-08-23
Replace own HMAC implementation with OpenSSL & disable MD2

ifranzki:


jschmidb:
lgtm. 
Just some typos in commit msgs:
- common: Build without support for MD2 by default
MD2 is considered very weak, and has been disabled in may other   <- many other 
- common: The OpenSSL HMAC context is not savable  <- saveable
Mark it as not savable in the ...   <- saveable

jschmidb (approved):

ifranzki:
Fixed typos in commit messages, no code changes.

PR #585, opened 2022-08-24
Supply useful information in CK_TOKEN_INFO.ulRwSessionCount

ifranzki:
Currently only CK_TOKEN_INFO.ulSessionCount is maintained with the total number of sessions open by a token, but ulRwSessionCount is set to CK_UNAVAILABLE_INFORMATION.

Besides the total number of sessions also maintain the number of R/W sessions, so that ulRwSessionCount can be returned with accurate information.

Similar as for the total number of sessions, the counter is kept in the shared memory segment of the pkcsslotd. It exists once as global counter, and once per process. When a process dies without unregistering from pkcsslotd, the garbage collector of pkcsslotd decrements the global counter by the amount of the per-process counter to keep the global counter consistent.


jschmidb (approved):
lgtm

PR #588, opened 2022-08-30
Misc EP11 fixes

ifranzki:


jschmidb:
Question on wording in commit msg:

> In case of a deferred init operation, perform mechanism and policy checks during the C_xxxInit call already. 

Do you mean:

> In case of a deferred init operation, mechanism and policy checks are already performed during the C_xxxInit call, but we want the checks as early as possible.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -7239,6 +7238,21 @@ CK_RV ep11tok_sign_init(STDLL_TokData_t * tokdata, SESSION * session,
         ctx->context = ep11_sign_state;
         ctx->context_len = ep11_sign_state_l;
         ctx->pkey_active = FALSE;
+        if (mech != &ctx->mech) { /* deferred init dup'ed mech already */
+            ctx->mech.mechanism = mech->mechanism;
+            if (mech->ulParameterLen > 0 && mech->pParameter != NULL) {
+                ctx->mech.pParameter = (CK_BYTE *) malloc(mech->ulParameterLen);
+                if (ctx->mech.pParameter == NULL) {
```
jschmidb, line 7245: What if ctx->mech.pParameter was not NULL before? Would we have to free the old buffer before allocating and assigning the new one?

ifranzki:
No, the point is that they are currently NOT performed during the C_xxxInit call, so this is what we add in this commit. We now DO perform these checks at the C_xxxInit.

Maybe the word 'already' at the end of the sentence is misleading. It was meant to point out to perform it already now at C_xxInit, and not later...

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: If we are in an init call at that point, it has already been checked that the context is not already active. In that case, ctx->mech.pParameter is supposed to be NULL.

jschmidb (approved):
Looks good.

ifranzki:
Removed the world 'already' from the commit message. No code change.

PR #589, opened 2022-08-30
CCA: Fix typo in define name

ifranzki:
Looks like @jschmidb has not spotted all my typos I usually do back then :-)


jschmidb:
Ha ha, yes, I didn't see this one. But looks good then.

jschmidb (approved):

Issue #579, opened 2022-08-17
Get rid of --disable-locks (aka transactional memory) configure option


ifranzki:
The use of transactional memory is broken in openCryptoki anyway. Also, many code paces use locks, even with `--disable-locks`. `--enable-locks` is the default anyway.

This would also allow us to remove some source files that are duplicates from the with-locks version:
- lock_btree.c vs. btree.c
- lock_sess_mgr.c vs. sess_mgr.c

Some source files have `#ifdef ENABLE_LOCKS` to distinguish the 2 cases:
- local_types.h
- hostdefs.h
- apiutil.c

PR #592, opened 2022-09-08
Add support for C_SessionCancel (PKCS#11 v 3.0)

ifranzki:
The changes in the new_host.c files are all the same in all 3 files.
Similar, the changes in sess_mgr.c and lock_sess_mgr.c are also the same.
So you only need to review one of each.

### DIFF
--- a/testcases/pkcs11/sess_mgmt.c
+++ b/testcases/pkcs11/sess_mgmt.c
@@ -1197,6 +1197,170 @@ CK_RV do_OperationState3(void)
     return rc;
 }
 
+CK_RV do_SessionCancel(void)
+{
+    CK_SLOT_ID slot_id;
+    CK_SESSION_HANDLE session1;
+    CK_FLAGS flags;
+    CK_BYTE user_pin[PKCS11_MAX_PIN_LEN];
+    CK_ULONG user_pin_len;
+    CK_RV rc;
+
+    CK_BYTE original[1024];
+    CK_BYTE crypt1[1024];
+    CK_BYTE crypt2[1024];
+    CK_BYTE trash1[8];
+    CK_BYTE trash2[8];
+
+    CK_ULONG orig_len, crypt1_len, crypt2_len, trash1_len, trash2_len;
+    CK_ULONG i;
+
+    CK_ULONG key_len = 16;
+    CK_ATTRIBUTE key_gen_tmpl[] = {
+        {CKA_VALUE_LEN, &key_len, sizeof(CK_ULONG)}
+    };
+
+    CK_MECHANISM mech;
+    CK_OBJECT_HANDLE h_key;
+
+    printf("do_SessionCancel...\n");
+    slot_id = SLOT_ID;
+
+    // create two USER RW sessions
+    flags = CKF_SERIAL_SESSION | CKF_RW_SESSION;
+    rc = funcs->C_OpenSession(slot_id, flags, NULL, NULL, &session1);
+    if (rc != CKR_OK) {
+        show_error("   C_OpenSession #1", rc);
+        return rc;
+    }
+
+    if (get_user_pin(user_pin))
+        return CKR_FUNCTION_FAILED;
+
+    user_pin_len = (CK_ULONG) strlen((char *) user_pin);
+
+    rc = funcs->C_Login(session1, CKU_USER, user_pin, user_pin_len);
+    if (rc != CKR_OK) {
+        show_error("   C_Login #1", rc);
+        return rc;
+    }
+
+    orig_len = sizeof(original);
+    for (i = 0; i < orig_len; i++)
+        original[i] = i % 255;
+
+    trash1_len = sizeof(trash1);
+    memcpy(trash1, "asdflkjasdlkjadslkj", trash1_len);
+
+    // first generate a AES key
+    mech.mechanism = CKM_AES_KEY_GEN;
+    mech.ulParameterLen = 0;
+    mech.pParameter = NULL;
+
+    if (!mech_supported(slot_id, mech.mechanism)) {
+        printf("Mechanism %s not supported. (skipped)\n",
+               mech_to_str(mech.mechanism));
+        funcs->C_CloseSession(session1);
+        return 0;
+    }
+
+    rc = funcs->C_GenerateKey(session1, &mech, key_gen_tmpl, 1, &h_key);
+    if (rc != CKR_OK) {
+        show_error("   C_GenerateKey #1", rc);
+        return rc;
+    }
+    // now encrypt the original data all at once using CBC
+    mech.mechanism = CKM_AES_CBC;
+    mech.ulParameterLen = 16;
+    mech.pParameter = "1234qwerasdfyxcv";
+
+    rc = funcs->C_EncryptInit(session1, &mech, h_key);
+    if (rc != CKR_OK) {
+        show_error("   C_EncryptInit #1", rc);
+        return rc;
+    }
+
+    // cancel the encrypt session
+    rc = funcs3->C_SessionCancel(session1, CKF_ENCRYPT);
+    if (rc != CKR_OK) {
+        show_error("   C_SessionCancel #1", rc);
+        return rc;
+    }
+
+    crypt1_len = sizeof(crypt1);
+    rc = funcs->C_Encrypt(session1, original, orig_len, crypt1, &crypt1_len);
+    if (rc != CKR_OPERATION_NOT_INITIALIZED) {
+        show_error("   C_Encrypt #1", rc);
+        return rc;
+    }
+
+    // now, begin encrypting multipart
+    rc = funcs->C_EncryptInit(session1, &mech, h_key);
+    if (rc != CKR_OK) {
+        show_error("   C_EncryptInit #2", rc);
+        return rc;
+    }
+
+    // cancel non existing sign session (no effect expected)
+    rc = funcs3->C_SessionCancel(session1, CKF_SIGN);
+    if (rc != CKR_OK) {
+        show_error("   C_SessionCancel #2", rc);
+        return rc;
+    }
+
+    crypt2_len = sizeof(crypt2);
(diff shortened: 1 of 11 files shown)


testcases/pkcs11/sess_mgmt.c
```diff
@@ -1197,6 +1197,171 @@ CK_RV do_OperationState3(void)
...
+    CK_ATTRIBUTE key_gen_tmpl[] = {
+        {CKA_VALUE_LEN, &key_len, sizeof(CK_ULONG)}
+    };
+
+    CK_MECHANISM mech;
+    CK_OBJECT_HANDLE h_key;
+
+
+    printf("do_SessionCancel...\n");
+    slot_id = SLOT_ID;
```
jschmidb, line 1228: TRACE_INFO instead of printf ? And maybe remove one of the two blank lines ...

testcases/pkcs11/sess_mgmt.c
```diff
@@ -1197,6 +1197,171 @@ CK_RV do_OperationState3(void)
...
+        return rc;
+    }
+
+    rc = funcs->C_CloseSession(session1);
+    if (rc != CKR_OK) {
+        show_error("   C_CloseSession #1", rc);
+        return rc;
+    }
+
+    printf("Looks okay...\n");
```
jschmidb, line 1360: printf ?

usr/lib/api/api_interface.c
```diff
@@ -4864,21 +4864,44 @@ CK_RV C_LoginUser(CK_SESSION_HANDLE hSession, CK_USER_TYPE userType,
 CK_RV C_SessionCancel(CK_SESSION_HANDLE hSession, CK_FLAGS flags)
 {
     CK_RV rv;
-
-    UNUSED(hSession);
-    UNUSED(flags);
+    API_Slot_t *sltp;
+    STDLL_FcnList_t *fcn;
```
jschmidb, line 4868: Are there spaces for indentation in line 4867?

usr/lib/api/api_interface.c
```diff
@@ -4864,21 +4864,44 @@ CK_RV C_LoginUser(CK_SESSION_HANDLE hSession, CK_USER_TYPE userType,
...
+    if ((fcn = sltp->FcnList) == NULL) {
+        TRACE_ERROR("%s\n", ock_err(ERR_TOKEN_NOT_PRESENT));
+        return CKR_TOKEN_NOT_PRESENT;
+    }
+    if (fcn->ST_SessionCancel) {
+        BEGIN_OPENSSL_LIBCTX(Anchor->openssl_libctx, rv)
+        // Map the Session to the slot session
+        rv = fcn->ST_SessionCancel(sltp->TokData, &rSession, flags);
+        END_OPENSSL_LIBCTX(rv)
+
```
jschmidb, line 4899: remove blank line?

usr/include/pkcs11types.h
```diff
@@ -1151,6 +1159,10 @@ typedef CK_ULONG CK_RV;
 /* CKR_FUNCTION_REJECTED is new for v2.20 */
 #define CKR_FUNCTION_REJECTED                 0x00000200
 
+/* New to v3.0 */
+#define CKR_TOKEN_RESOURCE_EXCEEDED           0x00000201
+#define CKR_OPERATION_CANCEL_FAILED           0x00000202
```
holger-dengler, line 1164: CKR_OPERATION_CANCEL_FAILED is not used in your implementation below. Does that mean, that you support the session cancel also for dual function operations?

The standard also describes, that this value should be returned, if an operation can not be canceled. How do you handle these cases?

testcases/pkcs11/sess_mgmt.c
ifranzki: The whole sess_mgmt testcase does not use the regular testcase_begin, etc macros from regress.h....
I would like to change it to use that at some point in time, but not in this PR. 
The other functions also use printf at that places, so lets leave it as is for now, and adapt in globally at a later time.

I will remove the blank line.

testcases/pkcs11/sess_mgmt.c
ifranzki: see above

usr/lib/api/api_interface.c
ifranzki: Yes there are spaces. Github does not color the spaces because the were already there with the removed line content.

usr/lib/api/api_interface.c
ifranzki: done

usr/lib/icsf_stdll/new_host.c
```diff
@@ -766,6 +766,35 @@ CK_RV SC_SetOperationState(STDLL_TokData_t *tokdata,
...
+    if (!sess) {
+        TRACE_ERROR("%s\n", ock_err(ERR_SESSION_HANDLE_INVALID));
+        rc = CKR_SESSION_HANDLE_INVALID;
+        goto done;
+    }
+
+    rc = session_mgr_cancel(tokdata, sess, flags);
+
+done:
+    TRACE_INFO("SC_SessionCancel: sess = %lu\n", sSession->sessionh);
```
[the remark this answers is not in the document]
ifranzki: Yes thats intended. `sSession->sessionh` is the session handle , while `sess` is the pointer to the the session structure. The session handle is of more interest in the trace, than the pointer.

usr/lib/common/sess_mgr.c
```diff
@@ -1206,6 +1206,46 @@ CK_RV session_mgr_set_op_state(STDLL_TokData_t *tokdata, SESSION *sess,
     return CKR_OK;
 }
 
+CK_RV session_mgr_cancel(STDLL_TokData_t *tokdata, SESSION *sess,
```
holger-dengler, line 1209: Less than 10% of the sess_mgr.c and lock_sess_mgr.c differs. I would highly recommend to fix this as soon as possible, instead of duplicating more and more code with each new feature.
Sure, this will not be fixed in the current PR, but please add it to your ToDo list.

usr/lib/icsf_stdll/new_host.c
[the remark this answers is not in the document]
holger-dengler: `sSession->sessionh` is the value of the caller. But the caller does not know about the local variable `sess`. I would prefer to use the variable name in the message, which is known to the caller. Otherwise it might be a bit confusing.

testcases/pkcs11/sess_mgmt.c
```diff
@@ -1197,6 +1197,170 @@ CK_RV do_OperationState3(void)
     return rc;
 }
 
+CK_RV do_SessionCancel(void)
```
holger-dengler, line 1200: The C_SessionCancel() function can still return with "not implemented", depending of the capabilities of the token, right? In these cases, the test should be skipped.

usr/lib/icsf_stdll/new_host.c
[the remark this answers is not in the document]
ifranzki: > sSession->sessionh is the value of the caller

Well, yes, and no. The sessions are managed in a 2 level btree setup. There is a session btree managed by the API layer, and there is a per token session btree managed by the token. `sSession->sessionh` is actually the session handle as in the token session btree. The external session handle is not known by the token (i.e. in new_host.c).
So tracing `sSession->sessionh` is only only somewhat useful, better would be to trace the external session handle, but in any case its better than tracing the session pointer.

usr/include/pkcs11types.h
ifranzki: Right, the code also supports canceling dual function operations. For dual function operations simply both session states are active, and thus can be canceled without any special handling.

In opencryptoki, canceling a session can not fail. If a operation is not active that is to be canceled, this should not be treated as an error, according to the standard.

So yes, adding these return codes is somewhat superfluous, but I wanted to have them in pkcs11types.h anyway, just for completeness.

usr/lib/common/sess_mgr.c
ifranzki: 100% agreed! See also issue https://github.com/opencryptoki/opencryptoki/issues/579
With that we could get rid of these duplicated files.

usr/lib/icsf_stdll/new_host.c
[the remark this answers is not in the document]
holger-dengler: Ok, but log the "real" variable name could help a lot in debugging sessions.

testcases/pkcs11/sess_mgmt.c
ifranzki: In theory yes, but all our tokens support it since its implemented in common code.
Your statement would be true for almost all C_xxx functions of PKCS#11, so we should add it all over the place. However, this are opencryptoki testcases, so we can assume to know what opencryptoki implements.

usr/lib/icsf_stdll/new_host.c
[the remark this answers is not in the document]
ifranzki: Its done that way in many other functions.... I would like to stick with how it is elsewhere (although I somewhat agree with you).

testcases/pkcs11/sess_mgmt.c
holger-dengler: ok

usr/lib/common/sess_mgr.c
holger-dengler: ok

usr/include/pkcs11types.h
holger-dengler: I agree, the error code has to be added in anyways. If it is not used, because it could not happen in your implementation, it is also ok with me.

usr/lib/icsf_stdll/new_host.c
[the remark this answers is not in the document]
holger-dengler: ok. But please add to your ToDo list, that we have to re-design the tracing for all these cases.

jschmidb (approved):

holger-dengler (approved):
ACK

PR #593, opened 2022-09-08
pkcs11: rework pin handling

holger-dengler:
This PR contains some patches to make the PIN handling work as expected. Also add a some documentation about the usage of the environment variable to the man page.

### DIFF
--- a/usr/sbin/p11sak/p11sak.c
+++ b/usr/sbin/p11sak/p11sak.c
@@ -112,13 +112,15 @@ static void load_pkcs11lib(void)
     atexit(unload_pkcs11lib);
 }
 
-static CK_RV get_pin(char **pin, size_t *pinlen)
+static CK_RV get_pin_prompt(char **pin, size_t *pinlen)
 {
     struct termios old, new;
     int nread;
     char *user_input = NULL;
     size_t buflen = 0;
     CK_RV rc = 0;
+
+    printf("Please enter user PIN: ");
 
     /* turn echoing off */
     if (tcgetattr(fileno(stdin), &old) != 0)
@@ -3186,28 +3188,17 @@ int main(int argc, char *argv[])
     /* now try to load the pkcs11 lib (will exit(99) on failure) */
     load_pkcs11lib();
 
-    /* Prompt for PIN if not already set via option */
-    if (!pin) {
-        printf("Please enter user PIN: ");
-        rc = get_pin(&pin, &pinlen);
-        if (rc != 0)
-            rc = CKR_FUNCTION_FAILED;
-        else
-            pin_allocated = ckb_true;
+    /* no pin in argv, fallback 1: try env */
+    if (!pin)
+        pin = getenv("PKCS11_USER_PIN");
 
-        if (pin == NULL || strlen(pin) == 0) {
-            char *s = getenv("PKCS11_USER_PIN");
-            if (s) {
-                free(pin);
-                pin = strdup(s);
-                if (!pin) {
-                    rc = CKR_HOST_MEMORY;
-                    goto done;
-                }
-            } else {
-                goto done;
-            }
+    /* no pin in argv or env, fallback 2: prompt */
+    if (!pin) {
+        if (get_pin_prompt(&pin, &pinlen) != 0) {
+            rc = CKR_FUNCTION_FAILED;
+            goto done;
         }
+        pin_allocated = ckb_true;
     }
 
     /* Open PKCS#11 session */
--- a/usr/sbin/p11sak/p11sak.c
+++ b/usr/sbin/p11sak/p11sak.c
@@ -112,53 +112,47 @@ static void load_pkcs11lib(void)
     atexit(unload_pkcs11lib);
 }
 
-static CK_RV get_pin_prompt(char **pin, size_t *pinlen)
+static char *get_pin_prompt(void)
 {
     struct termios old, new;
     int nread;
-    char *user_input = NULL;
-    size_t buflen = 0;
-    CK_RV rc = 0;
+    size_t pinlen = 0;
+    char *pin = NULL;
 
     printf("Please enter user PIN: ");
+    fflush(stdout);
 
-    /* turn echoing off */
+    /* turn echo off */
     if (tcgetattr(fileno(stdin), &old) != 0)
-        return -1;
+        goto out;
 
     new = old;
     new.c_lflag &= ~ECHO;
+
     if (tcsetattr(fileno(stdin), TCSAFLUSH, &new) != 0)
-        return -1;
+        goto out;
 
-    /* read the pin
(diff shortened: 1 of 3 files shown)


holger-dengler:
Force push:
- remove more code
- add testcases

usr/sbin/p11sak/p11sak.c
```diff
@@ -112,14 +112,16 @@ static void load_pkcs11lib(void)
...
-static CK_RV get_pin(char **pin, size_t *pinlen)
+static CK_RV get_pin_prompt(char **pin, size_t *pinlen)
 {
     struct termios old, new;
     int nread;
     char *user_input = NULL;
     size_t buflen = 0;
     CK_RV rc = 0;
 
+    printf("Please enter user PIN: ");
```
ifranzki, line 122: maybe an fflush(stdout) to ensure it is really printed? As far as I remember printf only flushes when the string contains '\n' ....

holger-dengler: As far as I can see, the fflush() is not required here, because it happend anyways as soon as the getline() starts (so, right in time...).

ifranzki: I don't quite understand.... Does getline flush stdout ? 
I do see a fflush after the getline, but thats too late.

usr/sbin/p11sak/p11sak.c
```diff
@@ -112,53 +112,44 @@ static void load_pkcs11lib(void)
...
-     */
-    nread = getline(&user_input, &buflen, stdin);
-    if (nread == -1) {
-        rc = -1;
-        goto done;
-    }
+    /* note: getline will allocate memory for pin. */
+    nread = getline(&pin, &pinlen, stdin);
+    if (nread == -1)
+        goto out;
```
holger-dengler, line 137: I will move the getline() error handling below the re-enable of the console echo. Otherwise, the console echo stays disabled on getline() failures and the caller is required to reset the console.

usr/sbin/p11sak/p11sak.c
holger-dengler: It happens implicitely. In a debug session, I get the output in the line with the getline(). But you're right. It is cleaner to do it explicitely. I will att the fflush for the printf here.

holger-dengler:
force push:
- do `getline()` error handling after re-enabling the echo for console.
- explicit fflush(stdout)

ifranzki:
While you are at it: If you like you could also rework the get_pin() function in common/pkcs_utils.c in a similar way. This one is used by the other command line tools and maybe could/should also be used by p11sak?
Would be great if we would have only one implementation of a pin prompt .....

holger-dengler:
ok.

ifranzki (approved):

holger-dengler:
While thinking about a consolidation of the various get_pin() implementations in opencryptoki, I would like to rename the env variables for user- and so-pin to OCK_USER_PIN and OCK_SO_PIN. @ifranzki , @jschmidb, would that be ok for you?

Just to make it clear: this is no rename of the env variables, which are currently used for all the tests. It is just a (re)name of the env variables, which will be checked in all tools which today asks for a user- or so-pin.

usr/sbin/p11sak/p11sak.c
```diff
@@ -3186,28 +3188,17 @@ int main(int argc, char *argv[])
     /* now try to load the pkcs11 lib (will exit(99) on failure) */
     load_pkcs11lib();
 
-    /* Prompt for PIN if not already set via option */
+    /* no pin in argv, fallback 1: try env */
+    if (!pin)
+        pin = getenv("PKCS11_USER_PIN");
```
holger-dengler, line 3186: RFC: rename PKCS11_USER_PIN to OCK_USER_PIN in preparation for the consolidation.

jschmidb:
I think it's a good idea consolidating the env variables. 
But before just changing the variables in the tools, I would consolidate/rename the variables everywhere in opencryptoki to avoid confusion.
I saw that some variables are documented in our opencryptoki book, so this would have to be adapted also. I also saw that there are other variables like PKCSLIB, PKCS_APP_STORE, and PKCS11_SHMEM_FILE. They could also get the standard OCK_ prefix. 
Lastly, I saw that the PKCS11_USER_PIN is currently not documented in the p11sak manpage.

holger-dengler:
Patch 3 in this series contains some documentation for the env variable in p11sak man page. Therefore I would like to do the renaming of the pin env variables now, instead of documenting it as is and later rename it.

jschmidb:
OK ... 
Just one typo in commit msg here:
Add testcases for ping handling .... <- pin handling

holger-dengler:
ok, I'll fix it.

holger-dengler:
The next PR version will have a new common implementation for the pin handling and it will replace the one in p11sak and pkcs11_utils. Stay tuned...

usr/sbin/p11sak/p11sak.c
ifranzki: hm, well, the external manual documents PKCS11_USER_PIN for p11sak, so changing this is breaking a documented API.....
https://www.ibm.com/docs/en/linux-on-systems?topic=features-environment-variables

holger-dengler:
What? The external documentation differs from the man pages of the tools? hm...

usr/sbin/p11sak/p11sak.c
holger-dengler: I was not aware, that the external documentation differs from the man pages of the tools. In this case, we should not do a renaming.

holger-dengler: Wat about the env variable name for the so pin? PKCS11_SO_PIN or OCK_SO_PIN?

holger-dengler: My proposal: OCK_USER_PIN and OCK_SO_PIN is supported, as well as PKCS11_USER_PIN. The OCK_USER_PIN variable is preferred over PKCS11_USER_PIN. The man pages document all varaibles, but PKCS11_USER_PIN is marked as deprecated.

And we ask ID for a change in the external documentation.

ifranzki: Well, the p11sak man page did not mention the env var at all, so in that terms it was not wrong :-) 

In general I am OK with your approach, although it makes it more complicated at the end....

The other thing is that p11sak (in contrast to the other OCK tools) is supposed to be usable with other PKCS11 implementations as well (you can set env var PKCSLIB to point to a foreign PKCS11 implementation - this env var is also not mentioned in the man page....grrr) . It should in general not be OCK specific. So the name PKCS11_USER_PIN is also more general than OCK_USER_PIN.

So I am not sure of I would not simply leave it as is and keep it as PKCS11_USER_PIN.

PR #594, opened 2022-09-09
EP11: Add support for Bitcoin key derivation and Schnorr signatures

ifranzki:
Add support for IBM specific mechanism:

- CKM_IBM_ECDSA_OTHER - Schnorr signatures
- CKM_IBM_BTC_DERIVE - Bitcoin key derivation

Both require an EP11 host library version 3.1.0, otherwise they are not supported/available.


ifranzki:
@jschmidb I think you have already reviewed this series internally some time ago....

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -7140,6 +7234,57 @@ CK_RV ep11tok_check_single_mech_key(STDLL_TokData_t *tokdata, SESSION * session,
...
+    CK_MECHANISM mech;
+    ECSG_Var_t param;
+};
+
+static CK_RV ep11tok_ecdsa_other_mech_adjust(CK_MECHANISM *mech,
+                                      struct ECDSA_OTHER_MECH_PARAM *mech_ep11)
+{
+    CK_IBM_ECDSA_OTHER_PARAMS *param;
+
+    if (mech->mechanism != CKM_IBM_ECDSA_OTHER)
```
jschmidb, line 7247: Here it must be guaranteed that mech is never null

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -7140,6 +7234,57 @@ CK_RV ep11tok_check_single_mech_key(STDLL_TokData_t *tokdata, SESSION * session,
...
+                   __func__, param->submechanism);
+       return CKR_MECHANISM_PARAM_INVALID;
+    }
+
+    return CKR_OK;
+}
+
+CK_BOOL ep11tok_mech_single_only(CK_MECHANISM *mech)
+{
+    switch (mech->mechanism) {
```
jschmidb, line 7280: also here, mech must not be null

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -5161,12 +5197,46 @@ CK_RV ep11tok_derive_key(STDLL_TokData_t * tokdata, SESSION * session,
...
     }
 
+    if (mech->mechanism == CKM_IBM_BTC_DERIVE) {
+        /*
+         * CKM_IBM_BTC_DERIVE requires CKA_VALUE_LEN to specify the byte length
+         * of the to be derived EC key. CKA_VALUE_LEN is dependent on the
+         * curve used.
+         * CKA_VALUE_LEN can not be already in the user supplied template,
+         * since this is not allwed by the key template check routines.
+         */
```
jschmidb, line 5213: allowed

usr/lib/ep11_stdll/ep11adm.h
```diff
@@ -1,49 +1,18 @@
 /*
- * COPYRIGHT (c) International Business Machines Corp. 2020
+ * (C) Copyright IBM Corp. 2012, 2020
  *
```
jschmidb, line 3: 2012, 2022 ?

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: Yes this is guaranteed, this is the caller code:
```
    if (mech->mechanism == CKM_IBM_ECDSA_OTHER) {
        rc = ep11tok_ecdsa_other_mech_adjust(mech, &mech_ep11);
```
So if mech is NULL it would crash there already, but its caller already checks that mech is not null.

usr/lib/ep11_stdll/ep11adm.h
ifranzki: right, will fix

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: will fix

jschmidb (approved):

Issue #598, opened 2022-09-16
Build with NOAES, NODH, NO_EC, etc: make sure they work or remove them


ifranzki:
For older (unsecure) algorithms it is certainly useful to be able to disable them via compile switch, but for current algorithms I don't see a real need to be able to disable them.

ifranzki:
We also have support for CDMF mechanisms in common code, but no token is actually supporting CDMF.

Issue #174, opened 2018-09-11
Get rid of 32-bit support


p-steuer:
Erase 32-bit support from the codebase.

Issue #21, opened 2017-05-30
Get rid of "-DPKCS64" flag


pvital:
We now support more 64bit capable platforms than non-64bit capable, so the -DPKCS64 flag is not saving us anything.

Migrated from https://sourceforge.net/p/opencryptoki/bugs/7/

ifranzki:
Possibly related to https://github.com/opencryptoki/opencryptoki/issues/174

PR #600, opened 2022-09-16
Remove unused code for old mechanisms and key types

ifranzki:
No token supports those old mechanisms and/or key types, so this does not remove any functionality.
Its essentially removing dead/unused code.

Also fixes https://github.com/opencryptoki/opencryptoki/issues/598


jschmidb:
lgtm and makes absolutely sense!
I had a quick look at compile switches in the code and found:
ENABLE_LOCKS -> do we need this anymore, as transactional memory is removed?
PKCS64 -> do we still support 32 bit slot daemons?
PLUGGABLE_TOKENS_SUPPORTED -> never heard of this ... 
NO_EC in ica_specific.c -> similar to AES, does it make sense to allow excluding EC?

jschmidb (approved):

ifranzki:
> ENABLE_LOCKS -> do we need this anymore, as transactional memory is removed?

See https://github.com/opencryptoki/opencryptoki/issues/579

> PKCS64 -> do we still support 32 bit slot daemons?

See https://github.com/opencryptoki/opencryptoki/issues/21 and https://github.com/opencryptoki/opencryptoki/issues/174

> PLUGGABLE_TOKENS_SUPPORTED -> never heard of this ...

funny.... will take a look.

> NO_EC in ica_specific.c -> similar to AES, does it make sense to allow excluding EC?

Some of those NOxxx defines are to exclude common code parts for these algorithms for tokens that do not need the common code parts because they implement it themselves. Remember, most tokens compile and include all common code parts during linking, so this is a way to reduce the unused common code for those tokens. 
So these defines are at least somewhat useful.... but still questionable.

PR #601, opened 2022-09-20
Bumping up to version-3.19.0

ifranzki:


ChangeLog
```diff
@@ -1,3 +1,17 @@
++ openCryptoki 3.19
+- CCA: check for expected master key verification patterns at token init
+- CCA: check master key verification pattern of created keys to be as expected
+- EP11: check for expected wrapping key verification pattern at token init
+- CCA: check wrapping key verification pattern of created keys to be as expected
+- p11sak/pkcsconf: display PKCS#11 URIs
```
jschmidb, line 6: Here you probably want to say 'EP11'

ifranzki: right, thanks. 
Fixed.

PR #602, opened 2022-09-20
EP11: Fix typo in example EP11 token config file

ifranzki:


jschmidb:
Looks good! 
One question on specifying the expected wkvp: is it just a hex string or (maybe also) prefixed by 0x?

ifranzki:
> is it just a hex string or (maybe also) prefixed by 0x?

It can be prefixed with 0x, but it can not contains blanks or so.

jschmidb (approved):

PR #604, opened 2022-09-22
EP11: Handle m_Get/m_SetAttributeValue and protected key derive with session bound blobs

ifranzki:
Like other EP11 host library functions that operate on key blobs, m_GetAttributeValue and m_SetAttributeValue as well as m_UnwrpKey with CKM_IBM_CPACF_WRAP may also return CKR_SESSION_CLOSED when the EP11 session that the key blob is bound to is not logged in into the adapter that is currently used for the request.

For that the session must be passed through to the functions.

### DIFF
--- a/usr/lib/ep11_stdll/ep11_specific.c
+++ b/usr/lib/ep11_stdll/ep11_specific.c
@@ -582,7 +582,7 @@ static CK_RV get_ep11_target_for_apqn(uint_32 adapter, uint_32 domain,
                                       target_t *target, uint64_t flags);
 static void free_ep11_target_for_apqn(target_t target);
 static CK_RV update_ep11_attrs_from_blob(STDLL_TokData_t *tokdata,
-                                         TEMPLATE *tmpl);
+                                         SESSION * session, TEMPLATE *tmpl);
 
 
 /* defined in the makefile, ep11 library can run standalone (without HW card),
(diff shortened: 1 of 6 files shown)


usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -1518,15 +1518,11 @@ static CK_RV ab_unwrap_update_template(STDLL_TokData_t * tokdata,
...
-
-    rc = dll_m_GetAttributeValue(blob, blob_len, attrs,
-                                 sizeof(attrs) / sizeof(CK_ATTRIBUTE),
-                                 target_info->target);
-
-    put_target_info(tokdata, target_info);
+    RETRY_START(rc, tokdata)
+        rc = dll_m_GetAttributeValue(blob, blob_len, attrs,
+                                     sizeof(attrs) / sizeof(CK_ATTRIBUTE),
+                                     target_info->target);
```
jschmidb, line 1543: This is now a bit tricky to read: I wondered where you get target_info from after deleting the local var. 
But yes, it's defined in macro RETRY_START

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -716,10 +720,21 @@ static CK_RV ep11tok_pkey_wrap_handler(uint_32 adapter, uint_32 domain,
         goto done;
 
     /* Create the protected key via CKM_IBM_CPACF_WRAP */
+repeat:
     ret = dll_m_WrapKey(data->secure_key, data->secure_key_len,
                         NULL, 0, NULL, 0, &mech,
                         data->pkey_buf, data->pkey_buflen_p,
                         target | XCP_TGTFL_SET_SCMD);
+    if (ret == CKR_SESSION_CLOSED && retry == FALSE &&
+        data->ep11_session != NULL) {
```
jschmidb, line 729: Is this a realistic scenario, retrying after a CKR_SESSION_CLOSED? Why should the session then be open? Would it make sense to add a small delay before the retry?

jschmidb (approved):

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: Usually this happens when an APQN just got online, and thus have not yet logged in the session (because the login was performed some time before). 
So it can happen that the APQN just got online before ep11tok_pkey_skey2pkey() got called. So handle_all_ep11_cards() will call ep11tok_pkey_wrap_handler() for that APQN, but it might not yet have the session logged in. 
So yes, it is realistic that this can happen, although it is not very likely.

PR #605, opened 2022-09-22
Fixes for ICSF token

ifranzki:


usr/lib/config/configuration.c
```diff
@@ -233,7 +233,8 @@ static int confignode_dump_i(FILE *fp, struct ConfigBaseNode *n,
         }
         if (atbol) {
             confignode_dump_indent(fp, curindent);
-            atbol = 0;
+            if (curindent != 0)
+                atbol = 0;
```
jschmidb, line 237: I guess that curindent is never < 0, so you could check for > 0 as written in the commit msg: "if indention is > 0 ... "

jschmidb (commented):
lgtm

PR #606, opened 2022-09-23
Opencryptoki common code XTS support

vkanjur:
This patch adds XTS support to the common
code layer of opencryptoki.
Support for the tokens and improvements on the
testcases will come later.

Signed-off-by: kanjur <kanjur@ibm.com>

### DIFF
--- a/usr/lib/api/policy.c
+++ b/usr/lib/api/policy.c
@@ -255,6 +255,7 @@ static CK_RV policy_extract_key_data(get_attr_val_f getattr, void *d,
         *comptarget = COMPARE_SYMMETRIC;
         break;
     case CKK_AES:
+    case CKK_AES_XTS:
         *siglen = 128;
         /* Fallthrough */
     case CKK_GENERIC_SECRET:
@@ -889,7 +890,9 @@ static CK_RV policy_update_mech_info(policy_t p, CK_MECHANISM_TYPE mech,
         case CKM_AES_CTR:
         case CKM_AES_ECB:
         case CKM_AES_GCM:
+        case CKM_AES_XTS:
         case CKM_AES_KEY_GEN:
+        case CKM_AES_XTS_KEY_GEN:
         case CKM_AES_MAC:
         case CKM_AES_MAC_GENERAL:
         case CKM_AES_OFB:
--- a/usr/lib/cca_stdll/tok_struct.h
+++ b/usr/lib/cca_stdll/tok_struct.h
@@ -106,6 +106,7 @@ token_spec_t token_specific = {
 
     // AES
     &token_specific_aes_key_gen,
+    NULL,                       // aes_xts_key_gen
     &token_specific_aes_ecb,
     &token_specific_aes_cbc,
     NULL,                       // aes_ctr
@@ -117,6 +118,7 @@ token_spec_t token_specific = {
     NULL,                       // aes_cfb
     NULL,                       // aes_mac
     NULL,                       // aes_cmac
+	NULL,                       // aes_xts
     // DSA
     NULL,                       // dsa_generate_keypair
     NULL,                       // dsa_sign
--- a/usr/lib/common/decr_mgr.c
+++ b/usr/lib/common/decr_mgr.c
@@ -546,6 +546,38 @@ CK_RV decr_mgr_init(STDLL_TokData_t *tokdata,
         }
         memset(ctx->context, 0x0, sizeof(AES_CONTEXT));
         break;
+    case CKM_AES_XTS:
+        // XXX Copied from DES3, should be verified - KEY
+        if (mech->ulParameterLen != AES_INIT_VECTOR_SIZE ||
+                    mech->pParameter == NULL) {
+                    TRACE_ERROR("%s\n", ock_err(ERR_MECHANISM_PARAM_INVALID));
+                    rc = CKR_MECHANISM_PARAM_INVALID;
+                    goto done;
+                }
+            // is the key type correct?
+            //
+            rc = template_attribute_get_ulong(key_obj->template, CKA_KEY_TYPE,
+                                              &keytype);
+            if (rc != CKR_OK) {
+                TRACE_ERROR("Could not find CKA_KEY_TYPE for the key.\n");
+                goto done;
+            }
+
+            if (keytype != CKK_AES_XTS) {
+                TRACE_ERROR("%s\n", ock_err(ERR_KEY_TYPE_INCONSISTENT));
+                rc = CKR_KEY_TYPE_INCONSISTENT;
+                goto done;
+            }
+
+            ctx->context_len = sizeof(AES_CONTEXT);
+            ctx->context = (CK_BYTE *) malloc(sizeof(AES_CONTEXT));
+            if (!ctx->context) {
+                TRACE_ERROR("%s\n", ock_err(ERR_HOST_MEMORY));
+                rc = CKR_HOST_MEMORY;
+                goto done;
+            }
+            memset(ctx->context, 0x0, sizeof(AES_CONTEXT));
+            break;
     default:
         TRACE_ERROR("%s\n", ock_err(ERR_MECHANISM_PARAM_INVALID));
         rc = CKR_MECHANISM_INVALID;
@@ -755,6 +787,10 @@ CK_RV decr_mgr_decrypt(STDLL_TokData_t *tokdata,
         return rsa_x509_decrypt(tokdata, sess, length_only,
                                 ctx,
                                 in_data, in_data_len, out_data, out_data_len);
+    case CKM_AES_XTS:
+        return aes_xts_decrypt(tokdata, sess, length_only,
+                               ctx,
+                               in_data, in_data_len, out_data, out_data_len);
     case CKM_AES_CBC:
         return aes_cbc_decrypt(tokdata, sess, length_only,
                                ctx,
@@ -947,6 +983,11 @@ CK_RV decr_mgr_decrypt_update(STDLL_TokData_t *tokdata,
                                       ctx,
                                       in_data, in_data_len,
                                       out_data, out_data_len, 0x10);
+    case CKM_AES_XTS:
+        return aes_xts_decrypt_update(tokdata, sess, length_only,
+                                      ctx,
+                                      in_data, in_data_len,
+                                      out_data, out_data_len);
     default:
         TRACE_ERROR("%s\n", ock_err(ERR_MECHANISM_INVALID));
         return CKR_MECHANISM_INVALID;
@@ -1057,6 +1098,9 @@ CK_RV decr_mgr_decrypt_final(STDLL_TokData_t *tokdata,
     case CKM_AES_GCM:
         return aes_gcm_decrypt_final(tokdata, sess, length_only, ctx,
                                      out_data, out_data_len);
+    case CKM_AES_XTS:
+        return aes_xts_decrypt_final(tokdata, sess, length_only,
+                                     ctx, out_data, out_data_len);
     default:
         TRACE_ERROR("%s\n", ock_err(ERR_MECHANISM_INVALID));
         return CKR_MECHANISM_INVALID;
--- a/usr/lib/common/encr_mgr.c
+++ b/usr/lib/common/encr_mgr.c
@@ -548,6 +548,39 @@ CK_RV encr_mgr_init(STDLL_TokData_t *tokdata,
         }
         memset(ctx->context, 0x0, sizeof(AES_CONTEXT));
         break;
+    case CKM_AES_XTS:
+            // XXX Copied in from DES3, should be verified - KEY
+        if (mech->ulParameterLen != AES_INIT_VECTOR_SIZE ||
+                    mech->pParameter == NULL) {
+                    TRACE_ERROR("%s\n", ock_err(ERR_MECHANISM_PARAM_INVALID));
+                    rc = CKR_MECHANISM_PARAM_INVALID;
+                    goto done;
+                }
+        // is the key type correct?
+        //
+        rc = template_attribute_get_ulong(key_obj->template, CKA_KEY_TYPE,
+                                              &keytype);
+        TRACE_ERROR("Key type %lx\n",keytype);
+        if (rc != CKR_OK) {
+            TRACE_ERROR("Could not find CKA_KEY_TYPE for the key.\n");
+            goto done;
+        }
+
+        if (keytype != CKK_AES_XTS) {
+            TRACE_ERROR("%s\n", ock_err(ERR_KEY_TYPE_INCONSISTENT));
+            rc = CKR_KEY_TYPE_INCONSISTENT;
+            goto done;
+        }
+
+        ctx->context_len = sizeof(AES_CONTEXT);
+        ctx->context = (CK_BYTE *) malloc(sizeof(AES_CONTEXT));
+        if (!ctx->context) {
+            TRACE_ERROR("%s\n", ock_err(ERR_HOST_MEMORY));
+            rc = CKR_HOST_MEMORY;
+            goto done;
+        }
+        memset(ctx->context, 0x0, sizeof(AES_CONTEXT));
+        break;
     default:
         TRACE_ERROR("%s\n", ock_err(ERR_MECHANISM_PARAM_INVALID));
         rc = CKR_MECHANISM_INVALID;
@@ -791,6 +824,10 @@ CK_RV encr_mgr_encrypt(STDLL_TokData_t *tokdata,
         return aes_cfb_encrypt(tokdata, sess, length_only, ctx,
                                in_data, in_data_len,
                                out_data, out_data_len, 0x10);
+    case CKM_AES_XTS:
+            return aes_xts_encrypt(tokdata, sess, length_only, ctx,
+                                   in_data, in_data_len,
+                                   out_data, out_data_len);
     default:
         TRACE_ERROR("%s\n", ock_err(ERR_MECHANISM_PARAM_INVALID));
         return CKR_MECHANISM_INVALID;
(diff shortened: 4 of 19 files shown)


ifranzki:
Your signoff looks strange:
> Signed-off-by: kanjur <kanjur@ibm.com>

Could you please change it to your full name?

usr/lib/common/decr_mgr.c
```diff
@@ -546,6 +546,38 @@ CK_RV decr_mgr_init(STDLL_TokData_t *tokdata,
         }
         memset(ctx->context, 0x0, sizeof(AES_CONTEXT));
         break;
+    case CKM_AES_XTS:
+        // XXX Copied from DES3, should be verified - KEY
```
ifranzki, line 550: please remove (or correct) this comment....

usr/lib/cca_stdll/tok_struct.h
```diff
@@ -117,6 +118,7 @@ token_spec_t token_specific = {
     NULL,                       // aes_cfb
     NULL,                       // aes_mac
     NULL,                       // aes_cmac
+	NULL,                       // aes_xts
```
ifranzki, line 121: there is a tab character in front of NULL.

usr/lib/common/encr_mgr.c
```diff
@@ -539,6 +539,39 @@ CK_RV encr_mgr_init(STDLL_TokData_t *tokdata,
...
+        if (!ctx->context) {
+            TRACE_ERROR("%s\n", ock_err(ERR_HOST_MEMORY));
+            rc = CKR_HOST_MEMORY;
+            goto done;
+        }
+        memset(ctx->context, 0x0, sizeof(AES_CONTEXT));
+        break;
+    case CKM_AES_XTS:
+            // XXX Copied in from DES3, should be verified - KEY
+        if (mech->ulParameterLen != AES_INIT_VECTOR_SIZE ||
```
ifranzki, line 553: remove (or correct) the comment

usr/lib/common/encr_mgr.c
```diff
@@ -539,6 +539,39 @@ CK_RV encr_mgr_init(STDLL_TokData_t *tokdata,
...
+                    mech->pParameter == NULL) {
```
ifranzki, line 554: Indention is wrong (same as above)

usr/lib/common/decr_mgr.c
```diff
@@ -546,6 +546,38 @@ CK_RV decr_mgr_init(STDLL_TokData_t *tokdata,
...
+
+            if (keytype != CKK_AES_XTS) {
+                TRACE_ERROR("%s\n", ock_err(ERR_KEY_TYPE_INCONSISTENT));
+                rc = CKR_KEY_TYPE_INCONSISTENT;
+                goto done;
+            }
+
+            ctx->context_len = sizeof(AES_CONTEXT);
+            ctx->context = (CK_BYTE *) malloc(sizeof(AES_CONTEXT));
+            if (!ctx->context) {
```
ifranzki, line 574: You don't seem to be using the `ctx->context` (aka AES_CONTEXT) anywhere in your added code.
For single part support you don't need that at all. Multi part support is completely missing, and this will probably need a different structure then.
So please remove the allocation of the context if you are not using it.

usr/lib/common/encr_mgr.c
```diff
@@ -539,6 +539,39 @@ CK_RV encr_mgr_init(STDLL_TokData_t *tokdata,
...
+            TRACE_ERROR("Could not find CKA_KEY_TYPE for the key.\n");
+            goto done;
+        }
+
+        if (keytype != CKK_AES_XTS) {
+            TRACE_ERROR("%s\n", ock_err(ERR_KEY_TYPE_INCONSISTENT));
+            rc = CKR_KEY_TYPE_INCONSISTENT;
+            goto done;
+        }
+
```
ifranzki, line 574: Remove allocation of context here as well.

usr/lib/common/key.c
```diff
@@ -6194,6 +6194,80 @@ CK_RV des3_wrap_get_data(TEMPLATE *tmpl,
     return CKR_OK;
 }
 
+//  aes_xts_set_default_attributes()
+//
+CK_RV aes_xts_set_default_attributes(TEMPLATE *tmpl, TEMPLATE *basetmpl, CK_ULONG mode)
```
ifranzki, line 6199: this seems to be a copy of function aes_set_default_attributes more or less, except a very few places.
I don't really like tat much code duplication.
I would suggest to modify aes_set_default_attributes and add a key type parameter to it.
All its callers then simply pass the desired key type to it.

usr/lib/common/key.c
```diff
@@ -6346,6 +6420,54 @@ CK_RV aes_validate_attribute(STDLL_TokData_t *tokdata, TEMPLATE *tmpl,
     }
 }
 
+//
+//
+CK_RV aes_xts_validate_attribute(STDLL_TokData_t *tokdata, TEMPLATE *tmpl,
```
ifranzki, line 6425: Same here, this also seems to be a duplication of aes_validate_attribute with the exception of a few places.
Also modify aes_validate_attribute and pass a keytype parameter to it.

usr/lib/common/decr_mgr.c
ifranzki: ah never mind, you did implement multi part support....

usr/lib/common/encr_mgr.c
ifranzki: ah never mind, you did implement multi part support....

usr/lib/common/mech_aes.c
```diff
@@ -488,6 +488,107 @@ CK_RV aes_ctr_decrypt(STDLL_TokData_t *tokdata,
...
+                      CK_BYTE *out_data, CK_ULONG *out_data_len)
+{
+	OBJECT *key = NULL;
+	CK_RV rc;
+
+	if (!sess || !ctx || !out_data_len) {
+		TRACE_ERROR("%s received bad argument(s)\n", __func__);
+		return CKR_FUNCTION_FAILED;
+	}
+	// CKM_DES3_CBC requires the input data to be an integral
```
ifranzki, line 506: please correct the comment

usr/lib/common/key.c
```diff
@@ -6346,6 +6420,54 @@ CK_RV aes_validate_attribute(STDLL_TokData_t *tokdata, TEMPLATE *tmpl,
...
+                             CK_ATTRIBUTE *attr, CK_ULONG mode)
```
ifranzki, line 6426: indention is wrong.

usr/lib/common/mech_aes.c
```diff
@@ -488,6 +488,107 @@ CK_RV aes_ctr_decrypt(STDLL_TokData_t *tokdata,
...
+	// multiple of the block size
+	//
+	if (in_data_len % AES_BLOCK_SIZE != 0) {
```
ifranzki, line 509: As far as I know for XTS the data must be at least a full block, but can be any number of bytes other than that. So for single part operation (this function), the check should be `if (in_data_len >= AES_BLOCK_SIZE) {`

usr/lib/common/mech_aes.c
```diff
@@ -488,6 +488,107 @@ CK_RV aes_ctr_decrypt(STDLL_TokData_t *tokdata,
...
```
ifranzki, line 560: As far as I know for XTS the data must be at least a full block, but can be any number of bytes other than that. So for single part operation (this function), the check should be if (in_data_len >= AES_BLOCK_SIZE) {

usr/lib/common/mech_aes.c
```diff
@@ -488,6 +488,107 @@ CK_RV aes_ctr_decrypt(STDLL_TokData_t *tokdata,
...
 
+CK_RV aes_xts_encrypt(STDLL_TokData_t *tokdata,
+                      SESSION *sess,
+                      CK_BBOOL length_only,
+                      ENCR_DECR_CONTEXT *ctx,
+                      CK_BYTE *in_data,
+                      CK_ULONG in_data_len,
+                      CK_BYTE *out_data, CK_ULONG *out_data_len)
+{
+	OBJECT *key = NULL;
```
ifranzki, line 499: please do NOT use tabs for indention. Use 4 blanks instead of 1 tab. (throughout the whole function)

usr/lib/common/mech_aes.c
```diff
@@ -488,6 +488,107 @@ CK_RV aes_ctr_decrypt(STDLL_TokData_t *tokdata,
...
+
+CK_RV aes_xts_decrypt(STDLL_TokData_t *tokdata,
+                      SESSION *sess,
+                      CK_BBOOL length_only,
+                      ENCR_DECR_CONTEXT *ctx,
+                      CK_BYTE *in_data,
+                      CK_ULONG in_data_len,
+                      CK_BYTE *out_data, CK_ULONG *out_data_len)
+{
+	OBJECT *key = NULL;
```
ifranzki, line 550: please do NOT use tabs for indention. Use 4 blanks instead of 1 tab. (throughout the whole function)

usr/lib/common/mech_aes.c
```diff
@@ -3636,6 +3992,152 @@ CK_RV ckm_aes_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl)
...
+    CK_ATTRIBUTE *value_attr = NULL;
+    CK_ATTRIBUTE *key_type_attr = NULL;
+    CK_ATTRIBUTE *class_attr = NULL;
+    CK_ATTRIBUTE *local_attr = NULL;
+    CK_BYTE *aes_key = NULL;
+    CK_ULONG rc;
+    CK_ULONG key_size;
+    CK_ULONG token_keysize;
+    CK_BBOOL is_opaque = FALSE;
+    TRACE_ERROR("Reached here....AES_XTS\n");
```
ifranzki, line 4009: remove trace entry. Add one empty line here instead

usr/lib/common/mech_aes.c
```diff
@@ -3636,6 +3992,152 @@ CK_RV ckm_aes_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl)
...
+    rc = template_attribute_get_ulong(tmpl, CKA_VALUE_LEN, &key_size);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("Could not find CKA_VALUE_LEN for the key.\n");
+        return rc;
+    }
+
+    if (key_size != (AES_KEY_SIZE_128 *2) &&
+        key_size != (AES_KEY_SIZE_256 *2) ) {
+        return CKR_ATTRIBUTE_VALUE_INVALID;
```
ifranzki, line 4018: add trace entry here

usr/lib/common/mech_aes.c
```diff
@@ -3636,6 +3992,152 @@ CK_RV ckm_aes_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl)
...
+    }
+
+    if (token_specific.t_aes_xts_key_gen == NULL) {
+        TRACE_ERROR("%s\n", ock_err(ERR_MECHANISM_INVALID));
+        return CKR_MECHANISM_INVALID;
+    }
+
+    rc = token_specific.t_aes_xts_key_gen(tokdata, &aes_key, &token_keysize,
+                                      key_size, &is_opaque);
```
ifranzki, line 4027: indention is wrong here

usr/lib/common/mech_aes.c
```diff
@@ -3636,6 +3992,152 @@ CK_RV ckm_aes_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl)
 }
 
 
+//
+//
+CK_RV ckm_aes_xts_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl)
+{
```
ifranzki, line 3998: This function seem to be a copy of ckm_aes_key_gen with only a few changes. 
I would modify ckm_aes_key_gen and pass a key type parameter to it instead of duplicating all the code

usr/lib/common/mech_aes.c
```diff
@@ -3898,3 +4400,73 @@ CK_RV ckm_aes_wrap_format(STDLL_TokData_t *tokdata,
...
+
+
+//
+//
+CK_RV ckm_aes_xts_decrypt(STDLL_TokData_t *tokdata,
+                          CK_BYTE *in_data,
+                          CK_ULONG in_data_len,
+                          CK_BYTE *out_data,
+                          CK_ULONG *out_data_len,
+                          CK_BYTE *init_v, OBJECT *key)
```
ifranzki, line 4447: Maybe call init_v 'tweak' instead? For XTS it is usually called tweak.

usr/lib/common/mech_aes.c
```diff
@@ -1191,6 +1292,189 @@ CK_RV aes_ctr_decrypt_update(STDLL_TokData_t *tokdata,
     }
 }
 
+//
+//
+CK_RV aes_xts_encrypt_update(STDLL_TokData_t *tokdata,
```
ifranzki, line 1297: Regarding multi-part support:
I don't think that it will work that way.
XTS is somewhat special when it comes to chaining. 
See https://en.wikipedia.org/wiki/Disk_encryption_theory#/media/File:XTS_mode_encryption.svg
So the last 2 blocks need some special handling I guess. All parts before are easy for chaining.
I guess you need to buffer up to 2 blocks of input data in an AES-XTS context structure.

ifranzki (changes requested):
I don't think that the XTS support is ready to be merged tat that point in time, especially the multi part support needs much more carful design. I don't think that multi part  support can work the way it is implemented currently. XTS needs special handling because it does things quite different compared to AES CBC or similar:
- The IV used for chaining is calculated from the tweak with a special code using the first key of the XTS key.
- Thus, the OV (result of encryption of one part) can not be simply feed into the subsequent part, because the OV can not be simply used a tweak for the next block. 
- The very last 2 block of the data needs special handling, if the last block is not a full block. For that the common code needs to buffer up  to 2 blocks of data to be able to perform the special handling on the last 2 block on the Final call.
- The token specific XTS functions need to get an indication of it is the initial part, intermediate part, or final part. Dependent on that, it needs to deal with tweak, IV, OV differently.
- Since all this may influence the interface of the token specific XTS functions, I would like to have worked that out before adding any code to OCK that will be changed later on.

ifranzki:
close/reopen to trigger Travis....

usr/lib/api/policy.c
```diff
@@ -255,6 +255,7 @@ static CK_RV policy_extract_key_data(get_attr_val_f getattr, void *d,
         *comptarget = COMPARE_SYMMETRIC;
         break;
     case CKK_AES:
+    case CKK_AES_XTS:
         *siglen = 128;
```
juergenchrist, line 259: Are there signatures with AES_XTS?  If not, you should set `*siglen = 0` here.  I think AES-XTS needs an own case without fallthrough.

usr/lib/api/policy.c
```diff
@@ -255,6 +255,7 @@ static CK_RV policy_extract_key_data(get_attr_val_f getattr, void *d,
...
         /* Fallthrough */
```
juergenchrist, line 260: As far as I understand the PKCS11 description of AES-XTS, `CKA_VALUE_LEN` is either 32 or 64, i.e., the length of the concatenation of the two keys.  To compute the strength, you have to consider the keys individually.  So you actually need to divide the computed size by 2 for AES-XTS.

juergenchrist (changes requested):
As per request during todays call, I checked the policy implementation.

usr/lib/common/mech_aes.c
```diff
@@ -488,6 +488,107 @@ CK_RV aes_ctr_decrypt(STDLL_TokData_t *tokdata,
...
+                      CK_BYTE *out_data, CK_ULONG *out_data_len)
+{
+	OBJECT *key = NULL;
+	CK_RV rc;
+
+	if (!sess || !ctx || !out_data_len) {
+		TRACE_ERROR("%s received bad argument(s)\n", __func__);
+		return CKR_FUNCTION_FAILED;
+	}
+	// CKM_DES3_CBC requires the input data to be an integral
```
jschmidb, line 557: Looks like the DES3 related comment is copied from somewhere ... should be for AES-XTS

usr/lib/common/mech_aes.c
```diff
@@ -1191,6 +1292,189 @@ CK_RV aes_ctr_decrypt_update(STDLL_TokData_t *tokdata,
     }
 }
 
+//
+//
+CK_RV aes_xts_encrypt_update(STDLL_TokData_t *tokdata,
+                             SESSION *sess,
+                             CK_BBOOL length_only,
+                             ENCR_DECR_CONTEXT *ctx,
+                             CK_BYTE *in_data,
```
jschmidb, line 1301: General question: we currently never use 'const' in any of the function headers in mech_aes.c and others. But in_data should be 'const' here.

usr/lib/common/mech_aes.c
```diff
@@ -1556,6 +1840,78 @@ CK_RV aes_ctr_decrypt_final(STDLL_TokData_t *tokdata,
...
+    UNUSED(out_data);
+
+    if (!sess || !ctx || !out_data_len) {
+        TRACE_ERROR("%s received bad argument(s)\n", __func__);
+        return CKR_FUNCTION_FAILED;
+    }
+    // satisfy the compiler
+    //
+    if (length_only)
+        context = NULL;
```
jschmidb, line 1863: Is this stmt necessary?

usr/lib/common/mech_aes.c
```diff
@@ -1556,6 +1840,78 @@ CK_RV aes_ctr_decrypt_final(STDLL_TokData_t *tokdata,
...
+
+    context = (AES_CONTEXT *) ctx->context;
+
+    // DES3-CBC does no padding so there had better not be
```
jschmidb, line 1867: comment wrongly mentions DES3

usr/lib/common/mech_aes.c
```diff
@@ -3636,6 +3992,152 @@ CK_RV ckm_aes_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl)
...
+        TRACE_ERROR("template_update_attribute failed\n");
+        goto err;
+    }
+    local_attr = NULL;
+
+    return CKR_OK;
+
+err:
+    if (aes_key)
+        free(aes_key);
```
jschmidb, line 4127: The free routine can handle NULL args, so it's not necessary to do the checks. On the other hand they don't hurt ...

ifranzki:
Closing, because PRhttps://github.com/opencryptoki/opencryptoki/pull/614 has been merged instead.

PR #607, opened 2022-09-26
lib: asn1: initialize len variables

holger-dengler:
It could happen in error cases, that the "len" variables in some functions are used uninitialized. Set len to 0 in all cases, where this could happen.

This also fixes some warnings, if newer compilers are used.

Signed-off-by: Holger Dengler <dengler@linux.ibm.com>


holger-dengler:
fix typo in commit message

jschmidb (approved):

holger-dengler:
I've seen the compile warnings on a fedora 36 (gcc 12.1.1 20220507 (Red Hat 12.1.1-1)).

PR #609, opened 2022-09-27
EP11: Fail token init if EP11 token config file has missing APQN settings

ifranzki:
Fail the EP11 token init if neither APQN_ALLOWLIST nor APQN_ANY is specified, or if APQN_ALLOWLIST but no APQN is specified.


jschmidb (approved):
lgtm

PR #610, opened 2022-09-27
API: Trace OpenSSL error stack on return from API function

ifranzki:
In case an OpenSSL function failed inside of the API function, the OpenSSL error stack should show the reasons for the failure.

Please note that an error on the OpenSSL error stack does not necessarily mean that an error has happened that would cause the API function to fail. Some error situations are expected and internally handled, and do not lead to a failure of the API function.

### DIFF
--- a/usr/lib/api/api_interface.c
+++ b/usr/lib/api/api_interface.c
@@ -41,6 +41,15 @@
 
 void api_init();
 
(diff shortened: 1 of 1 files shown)


jschmidb (approved):
lgtm

PR #612, opened 2022-09-28
testcases: pkcsconf_tests.sh: fix interpreter

holger-dengler:
The script uses functions/commands, which are only available in bash. Change the interpreter line accordingly.

Signed-off-by: Holger Dengler <dengler@linux.ibm.com>


holger-dengler:
No. But if the other script don't use bash-exclusive syntax or commands, it should be ok.

jschmidb (approved):

ifranzki (approved):

PR #614, opened 2022-10-10
Common code support for AES-XTS

ifranzki:
This PR took the code from https://github.com/opencryptoki/opencryptoki/pull/606 from @vkanjur and reworked it to support multi-part operations as well.

@vkanjur I left you as the Author of the very first commit, since you originally developed it. I also kept your sign-off but added mine as well. I hope this is OK with you? 

It also adds a generic AES-XTS implementation that can be used by tokens to implement AES-XTS support. The generic implementation only requires the tokens to provide two callback functions. One to calculate the initial IV from the tweak value, and one for performing the AES-XTS full-block encryption/decryption.

Furthermore, this PR adds support for AES-XTS for the Soft and ICA tokens. Since the OpenSSL AES-XTS implementation only supports single-part operations, the generic AES implementation is used to implement multi part AES-XTS with the help of OpenSSL. For ICA, an enhanced libica providing a multi-part capable AES-XTS implementation is required. 

We will need more test vectors for AES-XTS, especially for non-full-block cases. Those should be added at a later time.


usr/lib/ica_s390_stdll/tok_struct.h
```diff
@@ -109,7 +109,7 @@ token_spec_t token_specific = {
     &token_specific_generic_secret_key_gen,
     // AES
     &token_specific_aes_key_gen,
-    NULL,                       // aes_xts_key_gen
+    &token_specific_aes_key_gen,
```
jschmidb, line 112: Is this intended: lines 111 and 112 contain the same function ptr?

usr/lib/common/mech_openssl.c
```diff
@@ -4076,6 +4090,170 @@ CK_RV openssl_specific_aes_cmac(STDLL_TokData_t *tokdata, CK_BYTE *message,
...
+
+    rc = aes_xts_cipher(in_data, in_data_len, out_data, out_data_len,
+                        tweak, encrypt, initial, final, iv,
+                        aes_xts_iv_from_tweak,
+                        aes_xts_cipher_blocks,
+                        &data);
+
+out:
+    if (data.tweak_ctx != NULL)
+            EVP_CIPHER_CTX_free(data.tweak_ctx);
```
jschmidb, line 4250: Indentation?

usr/lib/soft_stdll/soft_specific.c
```diff
@@ -530,6 +532,19 @@ CK_RV token_specific_aes_key_gen(STDLL_TokData_t *tokdata, CK_BYTE **key,
...
 }
 
+CK_RV token_specific_aes_xts_key_gen(STDLL_TokData_t *tokdata, CK_BYTE **key,
+                                    CK_ULONG *len, CK_ULONG keysize,
+                                    CK_BBOOL *is_opaque)
+{
+    *key = malloc(keysize);
+    if (*key == NULL)
+        return CKR_HOST_MEMORY;
+    *len = keysize;
```
jschmidb, line 542: I guess the keysize was already checked before calling this routine? (only 128 and 256)

usr/lib/common/h_extern.h
```diff
@@ -1482,6 +1482,17 @@ CK_RV ckm_aes_xts_crypt(STDLL_TokData_t *tokdata,
                         CK_BBOOL initial, CK_BBOOL final,
                         CK_BYTE *iv, CK_BBOOL encrypt);
 
+CK_RV aes_xts_cipher(CK_BYTE *in_data, CK_ULONG in_data_len,
```
jschmidb, line 1485: Similar to the previous pr: I guess at least in_data is 'const' here.

usr/lib/common/h_extern.h
```diff
@@ -1337,6 +1337,16 @@ CK_RV aes_ctr_decrypt_update(STDLL_TokData_t *tokdata, SESSION *sess,
                              CK_BYTE *in_data, CK_ULONG in_data_len,
                              CK_BYTE *out_data, CK_ULONG *out_data_len);
 
+CK_RV aes_xts_encrypt_update(STDLL_TokData_t *tokdata, SESSION *sess,
+                             CK_BBOOL length_only, ENCR_DECR_CONTEXT *context,
+                             CK_BYTE *in_data, CK_ULONG in_data_len,
```
jschmidb, line 1342: const CK_BYTE *in_data   ?

usr/lib/ica_s390_stdll/tok_struct.h
ifranzki: Oh, no, not intended :-) Although it technically does not make a difference, since it just generates a random key in the length as specified by the caller. Will fix anyway .

usr/lib/common/mech_openssl.c
ifranzki: fixed.

usr/lib/soft_stdll/soft_specific.c
ifranzki: yes, that has already been checked.

usr/lib/common/mech_aes.c
```diff
@@ -1191,6 +1270,135 @@ CK_RV aes_ctr_decrypt_update(STDLL_TokData_t *tokdata,
...
+                                  CK_BYTE *in_data,
+                                  CK_ULONG in_data_len,
+                                  CK_BYTE *out_data, CK_ULONG *out_data_len)
+{
+    AES_XTS_CONTEXT *context = NULL;
+    OBJECT *key = NULL;
+    CK_BYTE *clear = NULL;
+    CK_ULONG total, remain, out_len;
+    CK_RV rc;
+
```
jschmidb, line 1287: For shorter path length we could also check here if (in_data == NULL || in_data_len == 0) and just return

usr/lib/common/h_extern.h
ifranzki: Hm yes, should be const indeed. The problem is that the caller (encr_mgr_encrypt_update) does not define in_data as const (although it should be const there, too). 
I leave it as is for this PR, but I put it on my TODP list to update all places where arguments could be const.

usr/lib/common/mech_aes.c
ifranzki: this has already been checked by the caller (i.e. in SC_EncryptUpdate)

ifranzki: ... and `in_data_len == 0` is actually allowed. Also `in_data` may be `NULL` if `in_data_len` is zero. 
Only if `in_data_len` is not zero, then `in_data` must also be non-NULL.

usr/lib/common/mech_aes.c
```diff
@@ -4214,3 +4214,126 @@ CK_RV ckm_aes_xts_crypt(STDLL_TokData_t *tokdata,
 
     return rc;
 }
+
+CK_RV aes_xts_cipher(CK_BYTE *in_data, CK_ULONG in_data_len,
```
jschmidb, line 4218: What was the reason to add this generic implementation? I mean, yes, it's good to have one, but was it not possible to use openssl with a bit of glue code for this?

ifranzki: OpenSSL does not support multi-part operations with AES-XTS ! 

The OpenSSL man page states:

_The XTS implementation in OpenSSL does not support streaming. That is there must only be one EVP_EncryptUpdate call per EVP_EncryptInit_ex call (and similarly with the "Decrypt" functions)._

So with the generic implementation we can implement multi-part operations for the SoftToken by using OpenSSL's AES-ECB functions, also see commit message of Soft Token commit: https://github.com/opencryptoki/opencryptoki/pull/614/commits/a0a502ca6ec757cea9a7248219d086a956eb1ca9

ifranzki: ... and it also helps to implement AES-XTS with multi part support for the EP11 token (with protected keys) for example. The 2 callback functions corelate to what the PCC and KM instructions do....

vkanjur:
Approved

ifranzki:
@jschmidb Are you OK with this PR? If so, please approve.

@vkanjur Thanks for your approval, can you please also press the Approve button? (Navigate to one commit of this PR, then the green button "Review changes", then select Approve and press submit review - for whatever reason the approve function can only be found when in a commit .....)

Once this one is merged, I will close/cancel https://github.com/opencryptoki/opencryptoki/pull/606 .

vkanjur (approved):

jschmidb (approved):
lgtm

PR #616, opened 2022-10-18
Remove superfluous/unused files from misc directory

ifranzki:
These files are not used, have not been maintained for a long time, and thus are superfluous.


jschmidb:
From review together with Vishnu: you are removing superfluous files, but not all files in the misc directory. Why are you keeping e.g. pkcsslotd.in and pkcsslotd.service.in? Does someone need these files to start pkcsslotd as a service?

Makefile.am
```diff
@@ -41,6 +41,7 @@ include tools/tools.mk
 include doc/doc.mk
 
 install-data-hook:
+	getent group pkcs11 > /dev/null || $(GROUPADD) -r pkcs11
 if ENABLE_LIBRARY
```
jschmidb, line 45: Do we need to mention this changed behavior in the readme or manpage? Or even the opencryptoki book?

usr/sbin/pkcstok_migrate/pkcstok_migrate.mk
```diff
@@ -1,11 +1,4 @@
 sbin_PROGRAMS += usr/sbin/pkcstok_migrate/pkcstok_migrate
-noinst_HEADERS += misc/mech_types.h
-noinst_HEADERS += usr/lib/common/defs.h
-noinst_HEADERS += usr/lib/common/host_defs.h
-noinst_HEADERS += usr/include/local_types.h
-noinst_HEADERS += usr/lib/common/h_extern.h
-noinst_HEADERS += usr/lib/common/pkcs_utils.h
-noinst_HEADERS += usr/lib/common/pin_prompt.h
```
jschmidb, line 8: OK, /misc/mech_types.h is removed, so also remove the dependency here. But why removing the other lines? The h-files in usr/lib/common and /usr/include are still there?

Makefile.am
ifranzki: This is only during 'make install'. Without that it hits you when you build your own opencryptoki from source and want to install it without previously having the distro-opencryptoki installed (which did create the pkcs11 group), or created the group manually before. 

So this change will not make any difference for distro installed opencryptoki.

usr/sbin/pkcstok_migrate/pkcstok_migrate.mk
ifranzki: I am removing all of them because it does not make any sense to add them to noinst_HEADERS here. You would do that only if that directory contains headers that you do not want to install. 
Thefact that the sources include those headers is no reason for adding them to noinst_HEADERS.

ifranzki:
> From review together with Vishnu: you are removing superfluous files, but not all files in the misc directory. Why are you keeping e.g. pkcsslotd.in and pkcsslotd.service.in? Does someone need these files to start pkcsslotd as a service?

Because these other files are still used and needed. Those are installed by 'make install' so that pkcsslotd can be started as service via systemd or the old init way.

jschmidb (approved):
ok

vkanjur (approved):
Approved

PR #618, opened 2022-10-21
EP11 fixes

ifranzki:


usr/lib/api/mechtable.inc
```diff
@@ -80,7 +80,7 @@ const struct mechrow mechtable_rows[] =
      { "CKM_IBM_ATTRIBUTEBOUND_WRAP", CKM_IBM_ATTRIBUTEBOUND_WRAP,   0, MC_INFORMATION_UNAVAILABLE, MCF_WRAPUNWRAP | MCF_NEEDSPARAM },
      { "CKM_IBM_BTC_DERIVE",          CKM_IBM_BTC_DERIVE,            0, MC_INFORMATION_UNAVAILABLE, MCF_DERIVE | MCF_NEEDSPARAM },
      { "CKM_IBM_CMAC",                CKM_IBM_CMAC,                  0,           MC_KEY_DEPENDENT, MCF_SIGNVERIFY },
-     { "CKM_IBM_DILITHIUM",           CKM_IBM_DILITHIUM,             0,                       3366, MCF_SIGNVERIFY },/* Size unknown */
+     { "CKM_IBM_DILITHIUM",           CKM_IBM_DILITHIUM,             0,                       3366, MCF_KEYGEN | MCF_SIGNVERIFY },/* Size unknown */
```
jschmidb, line 83: 3366 is the size of a Dilithium signature. The Dilithium key size would be the sum of the key parts (rho, seed, ...). Shouldn't we use MC_KEY_DEPENDENT for keygen? But yes, CKM_IBM_DILITHIUM is used for both, sign/verify and keygen. Just wondering what's correct here ... 
BTW: may we remove the /* Size unknown */ comment?

jschmidb:
From review together with Vishnu:
Typo in commit msg: 
Currently this is not used anywhere, but lets correct it anyway.  <- let's correct it ...

usr/lib/api/mechtable.inc
ifranzki: > 3366 is the size of a Dilithium signature.

Correct (for that variant that we support currently - this will change once we support other variants). This column value is used by the policy checks to determine the signature size. So that value is correct. 

The  /* Size unknown */ comment is for the blocksize column, which is zero, thus the comment (like with other rows).

ifranzki:
Fixed typo in commit msg.

vkanjur (approved):
Approved

usr/lib/api/mechtable.inc
ifranzki: hm, well, maybe the /* Size unknown */ comment is really superfluous.... I guess I'll remove it, does not make much sense anyway.

ifranzki: removed comment.

jschmidb (approved):
approved

PR #619, opened 2022-10-24
EP11: Remove DSA/DH parameter generation mechanisms from CP filter

ifranzki:
Commit 836f2a25b928127c3bfb8f94a57cc16aa76a84c3 removed the DSA/DH parameter generation mechanisms CKM_DSA_PARAMETER_GEN and CKM_DH_PKCS_PARAMETER_GEN from the EP11 token code. Also remove them from the CP filter config file.


vkanjur:
In ep11_specific.c line number 1722, we should remove CKM_DH_PKCS_PARAMETER_GEN in the comment.
Is these two parameters supported by icsf token? There, it is still supported in the code.

ifranzki:
> In ep11_specific.c line number 1722, we should remove CKM_DH_PKCS_PARAMETER_GEN in the comment.

Right, I now have removed it there as well. Thanks for noticing this.

> Is these two parameters supported by icsf token? There, it is still supported in the code.

Yes, the ICSF token seems to support this still. However, I have never actually tested it, because none of our testcases use that mechs. Another thing to add to my TODO list :-)

vkanjur (approved):
Approved

jschmidb (approved):
approved

PR #620, opened 2022-11-04
EP11: fixes for CKM_IBM_BTC_DERIVE 

ifranzki:
- pass back chain code in mech param
- supply SPKI when deriving public key
- supply SPKI for private key import (all asym key types)

### DIFF
--- a/usr/lib/ep11_stdll/ep11_specific.c
+++ b/usr/lib/ep11_stdll/ep11_specific.c
@@ -2744,9 +2744,10 @@ static int get_curve_type_from_template(TEMPLATE *tmpl)
  * SPKIs for public imported RSA keys.
  * Similar to rawkey_2_blob, but keys must follow a standard BER encoding.
  */
-static CK_RV import_RSA_key(STDLL_TokData_t * tokdata, SESSION * sess,
-                            OBJECT * rsa_key_obj,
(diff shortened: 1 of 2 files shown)


usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -3746,9 +3755,13 @@ CK_RV token_specific_object_add(STDLL_TokData_t * tokdata, SESSION * sess,
     CK_ATTRIBUTE *attr = NULL;
     CK_BYTE blob[MAX_BLOBSIZE];
     size_t blobsize = sizeof(blob);
+    CK_BYTE spki[MAX_BLOBSIZE];
+    CK_ULONG spkisize = sizeof(spki);
     CK_RV rc;
```
jschmidb, line 3760: From review together with Vishnu: here you are using CK_ULONG spkisize, but in the functions import_xxxxx_key() you are using size_t *spki_size. Yes, it's the same, but just to mention. For comparison: blob_size is a size_t at both places.

ifranzki: Ah right. Will change it to size_t here as well (for consistency)

jschmidb (approved):
lgtm

vkanjur (approved):
Approved

PR #622, opened 2022-11-08
p11sak: Fix segfault when dilithium version is not specified with generate-key

ifranzki:
Command 'p11sak generate-key ibm-dilithium' segfaults because the dilithium version is not specified, but the code does not check for a NULL pointer.

### DIFF
--- a/usr/sbin/p11sak/p11sak.c
+++ b/usr/sbin/p11sak/p11sak.c
@@ -1843,6 +1843,12 @@ static CK_RV check_args_gen_key(p11sak_kt *kt, CK_ULONG keylength,
     case kt_3DES:
         break;
     case kt_IBM_DILITHIUM:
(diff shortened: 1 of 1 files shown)


jschmidb:
Just an idea: as far as we know, r2_65 is currently the only supported version. So we could make this the default and not issue an error msg. However, when we will have more than one, then it will make sense to require the version parm. 
Otherwise looks good to us (Vishnu and me).

jschmidb (approved):

vkanjur (approved):

ifranzki:
> Just an idea: as far as we know, r2_65 is currently the only supported version. So we could make this the default and not issue an error msg.

Yes, I thought about this, too, but ew will soon have newer variants, and I don't think that a (then outdated) round 2 variant should then be default....

PR #624, opened 2022-11-09
c99 fixes

ifranzki:


ifranzki:
I have missed to update the unit tests.... will work on this.

ifranzki:
> In .travis.yml, we also have some CFLAGS settings.

Good catch.

I have now removed the duplicate CFLAGS from travis (`-Wextra -std=c99`) since they are already set by configure.

I also changed configure to use `-std=c99` (instead of `-std=gnu99`). We do not use any `asm` statements (the code is using `__asm__` already), and Travis did enforce c99 level already so there is no reason to use the GNU variant of C99 here.

.travis.yml
```diff
@@ -18,41 +18,41 @@ matrix:
         - name: "linux-x86-clang-locks"
           os: linux
           compiler: clang
-          env: CONFIG_OPTS="--enable-swtok --enable-icsftok --enable-ccatok --enable-testcases --enable-locks --with-systemd" CFLAGS="-O3 -Wextra -std=c99 -pedantic -Werror -DDEBUG"
+          env: CONFIG_OPTS="--enable-swtok --enable-icsftok --enable-ccatok --enable-testcases --enable-locks --with-systemd" CFLAGS="-O3 -pedantic -Werror -DDEBUG"
```
holger-dengler, line 21: I would also move the `-pedantic` setting to configure.ac (here and in all other sections). Otherwise you're developing something and got findings in your work very late in the travis run. If you get warnings/errors earlier, you can fix it before creating the pull request.

testcases/common/common.c
```diff
@@ -1237,21 +1240,21 @@ CK_BBOOL do_GetFunctionList(void)
     if (pkcs11lib == NULL)
         goto ret;
 
-    *(void **)(&pfoo) = dlsym(pkcs11lib, "C_GetFunctionList");
-    if (pfoo == NULL)
+    *(void **)(&getfunclist) = dlsym(pkcs11lib, "C_GetFunctionList");
```
holger-dengler, line 1243: Maybe a dump question: why are you casting the variable here and not the return value of the dlsym() call? I think it would not make any difference, but for me it looks a bit uncommon.

Same is true for all other dlsym() calls.

holger-dengler (approved):
Looks good to me. 

I leave it up to you, if you want move the `-pedantic` setting from travis to configure.ac.  Also the my question about the cast may not necessarily need a change.

testcases/common/common.c
holger-dengler: ok, make sense. Thanks a lot! So we should leave it as is.

.travis.yml
ifranzki: Yes maybe that's a good idea. 
But the `-Werror` is left out by intention, as one may want to compile while there are still warnings.

ifranzki:
Added `-pedantic` to configure.ac and removed it from .travis.yml

jschmidb (approved):

vkanjur (approved):

PR #627, opened 2022-11-15
cppcheck fixes

ifranzki:


jschmidb:
While reviewing the "TESTCASES: Fix testcase return code" commit I saw that there's a mess with rc's of do_GetFunctionList().
1. There are two implementations of  do_GetFunctionList(): one in testcase/common/common.c and one in testcases/misc_tests/tok_obj2.c
2. The first one returns a CK_BBOOL, the 2nd one an int.

So the call in getobjectsize.c is probably ok, but while at it, we could consolidate the two routines into one returning an int  (or even a CK_RV as most of the other do_xxxx() routines).

ifranzki:
> we could consolidate the two routines into one returning an int (or even a CK_RV as most of the other do_xxxx() routines).

Oh body, this is a complete mess ;-) 

Not only do_GetFunctionList() but also do_ParseArgs() return a Boolean or similar value, but also the handing in case of a failure of this functions is wrong. Returning rc from main/() if rc = 0/false indicates an error does not make too much sense....

I guess this is something for another patch series. I put it onto my TODO list.

jschmidb (approved):
OK, after the fix is before the fix :-)

vkanjur (approved):

PR #628, opened 2022-11-16
More Covertry fixes

ifranzki:


usr/sbin/pkcsslotd/garbage_linux.c
```diff
@@ -493,7 +493,7 @@ BOOL IsValidProcessEntry(pid_t_64 pid, time_t_64 RegTime)
         Err = errno;
         if (Err == ESRCH) {
             /* The process was not found */
-            DbgLog(DL3, "IsValidProcessEntry: PID %d was not found in the "
+            DbgLog(DL3, "IsValidProcessEntry: PID %llu was not found in the "
                    "process table (kill() returned %s)",
```
jschmidb, line 497: I saw that pid_t_64 is a 'signed long long' (see slotmgr.h). Why not %lld?

usr/lib/common/new_host.c
```diff
@@ -3939,6 +3939,9 @@ CK_RV SC_GenerateRandom(STDLL_TokData_t *tokdata, ST_SESSION_HANDLE *sSession,
         goto done;
     }
 
+    if (ulRandomLen == 0)
+        goto done;
+
```
jschmidb, line 3944: Yes, makes sense. Unfortunately the pkcs#11 spec does not say what to do if the number of bytes is (perhaps accidentally) zero. 
Variants would be: zero out the buffer before returning? Or return with an error, e.g. CKR_ARGUMENTS_BAD, telling the application that the buffer shall not be used after the call to C_SeedRandom?

usr/lib/common/obj_mgr.c
```diff
@@ -949,9 +949,9 @@ CK_RV object_mgr_find_in_map1(STDLL_TokData_t *tokdata,
         }
 
         rc = object_unlock(obj);
+        locked = FALSE;
         if (rc != CKR_OK)
```
jschmidb, line 953: Commit msg above:
When object_unlock() fails, do not attempt  **to** unlock the object again

usr/lib/common/asn1.c
```diff
@@ -2642,8 +2642,15 @@ CK_RV der_encode_ECPrivateKey(CK_BBOOL length_only,
             goto error;
        }
 
-        ber_encode_CHOICE(FALSE, 1, &buf2, &len, (CK_BYTE *)val->bv_val,
-                          val->bv_len);
+        rc = ber_encode_CHOICE(FALSE, 1, &buf2, &len, (CK_BYTE *)val->bv_val,
```
vkanjur, line 2645: Do we need to do the same above in line 2567 ?

usr/sbin/pkcsslotd/garbage_linux.c
ifranzki: Ah yes, you are correct. Will change to `lld` then.

usr/lib/common/new_host.c
ifranzki: > Variants would be: zero out the buffer before returning?

Hmm, zeroing out a buffer of 0 bytes in length :-) 

> Or return with an error, e.g. CKR_ARGUMENTS_BAD

I thought about that, but is it really an error? The application asked for 0 bytes of random data and got 0 bytes of random data. So no error in my opinion.

Actually, this fix is because Coverty complained about a possible NULL pointer access when pRandom = NULL. 
The code already checks for `if (!pRandomData && ulRandomLen != 0)` and returns an error if so.
So pRandom is only allowed to be NULL if ulRandomLen is zero. This should not have caused any real NULL pointer access in the further code anyway, but shortcutting this makes sense, I would say.

usr/lib/common/obj_mgr.c
ifranzki: will fix.

usr/lib/common/asn1.c
ifranzki: Good catch! 

I guess Coverty did not flag that one because it's a `length_only = TRUE` operation, so it won't set `buf2` anyway, but just update `len`. Also `buf2` is not touched/used afterwards. 
Nevertheless, we should also check the return code here, otherwise we might continue with a wrong length if that `ber_encode_CHOICE` did fail.

Will fix.

jschmidb (approved):

vkanjur (approved):

PR #629, opened 2022-11-18
More Coverty fixes (for EP11 and ICA tokens)

ifranzki:
With some temporary code tweaking (comment asm statements out, replace libica calls with dummy functions, etc), I was able to build the EP11 and ICA tokens on x86 as well, and thus let Coverty find problems there as well.

### DIFF
--- a/usr/lib/ep11_stdll/ep11_specific.c
+++ b/usr/lib/ep11_stdll/ep11_specific.c
@@ -12735,6 +12735,8 @@ static ep11_target_info_t *get_target_info(STDLL_TokData_t *tokdata)
     target_info = *((void * volatile *)&ep11_data->target_info);
     if (target_info == NULL) {
         TRACE_ERROR("%s: target_info is NULL\n", __func__);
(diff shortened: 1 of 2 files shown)


usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -12735,6 +12735,7 @@ static ep11_target_info_t *get_target_info(STDLL_TokData_t *tokdata)
     target_info = *((void * volatile *)&ep11_data->target_info);
     if (target_info == NULL) {
         TRACE_ERROR("%s: target_info is NULL\n", __func__);
+        pthread_rwlock_unlock(&ep11_data->target_rwlock);
         return NULL;
```
jschmidb, line 12713: Below there is an error check for unlock:

> if (pthread_rwlock_unlock(&ep11_data->target_rwlock) != 0)

Would it make sense to also check and issue a TRACE_DEVEL in case the unlock fails?

ifranzki: Right. I added the trace stmt there.

jschmidb (approved):

PR #630, opened 2022-11-21
RPM: Update dependencies and config file permissions

ifranzki:


jschmidb (approved):
lgtm

vkanjur (approved):

ifranzki:
removed gcc-c++ and expect from requirements, since those are only needed for testcase build (which are not built during RPM processing).

Issue #635, opened 2023-01-09
Cross compilation is broken


mhartmay:
There are some "build tools" that are built for the target system, but should have been built for the build system, since they are run on the build system during the build process (note: this can work, but usually doesn't).

One example is the `tools/tableidxgen` binary.

PR #637, opened 2023-01-13
configure: Support cross compilation

ifranzki:
Fixes https://github.com/opencryptoki/opencryptoki/issues/635 from @mhartmay

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -85,6 +85,8 @@ AC_SUBST(lockdir)
 logdir=$localstatedir/log/opencryptoki
 AC_SUBST(logdir)
 
+AM_CONDITIONAL([CROSS], [test "x$cross_compiling" = xyes])
+
 dnl ---
 dnl --- Check all --enable/--disable-features
 dnl ---
@@ -595,6 +597,7 @@ CFLAGS+=" -std=c99 -Werror=implicit-int -Werror=implicit-function-declaration -W
 
 # At this point, CFLAGS is set to something sensible
 AC_PROG_CC
+AX_PROG_CC_FOR_BUILD
 AC_PROG_CXX
 # AC_PROG_CXX will return "g++" even if no c++ compiler is installed.
 # Check for that case, and issue an error if so.
--- a/tools/tools.mk
+++ b/tools/tools.mk
@@ -1,7 +1,27 @@
 noinst_PROGRAMS += tools/tableidxgen tools/policyexamplegen
 
+TOOLS_CFLAGS = -I${srcdir}/usr/include -I${srcdir}/usr/lib/api -I${top_builddir}/usr/lib/api
+
 tools_tableidxgen_SOURCES = tools/tableidxgen.c usr/lib/api/mechtable.inc
-tools_tableidxgen_CFLAGS = -I${srcdir}/usr/include -I${srcdir}/usr/lib/api
 
 tools_policyexamplegen_SOURCES = tools/policyexamplegen.c usr/lib/api/mechtable.c
-tools_policyexamplegen_CFLAGS = -I${srcdir}/usr/include -I${srcdir}/usr/lib/api -I${top_builddir}/usr/lib/api
+
+if CROSS
+tools_tableidxgen_LINK = $(CC_FOR_BUILD) $(CFLAGS_FOR_BUILD) $(LDFLAGS_FOR_BUILD) -o $@
+
+$(tools_tableidxgen_OBJECTS): CC=$(CC_FOR_BUILD)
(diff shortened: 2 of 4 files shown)


ifranzki:
Travis failures: seem to be unrelated to the change, just timeouts.
The OSS credits for the OpenSource project OpenCryptoki seem to have just ran out again, so I can trigger a build at the moment. I have sent t not to Travis support to refill OSS credits.

tools/tools.mk
```diff
@@ -1,7 +1,33 @@
...
+tools_tableidxgen_LDADD =
+tools_tableidxgen_LINK = $(LIBTOOL) --tag=CC --mode=link $(CC_FOR_BUILD) \
+		$(CFLAGS_FOR_BUILD) $(LDFLAGS_FOR_BUILD) -o $@
+
+$(tools_tableidxgen_OBJECTS): CC=$(CC_FOR_BUILD)
+$(tools_tableidxgen_OBJECTS): CFLAGS=$(CFLAGS_FOR_BUILD) $(TOOLS_CFLAGS)
+$(tools_tableidxgen_OBJECTS): LDFLAGS=$(LDFLAGS_FOR_BUILD)
+
+tools_policyexamplegen_LDADD =
+tools_policyexamplegen_LINK = $(LIBTOOL) --tag=CC --mode=link $(CC_FOR_BUILD) \
```
juergenchrist, line 19: Any particular reason why we use libtool to build the policy generator?  It should not have any dependencies to a library that requires libtool.

tools/tools.mk
```diff
@@ -1,7 +1,33 @@
...
+		$(CFLAGS_FOR_BUILD) $(LDFLAGS_FOR_BUILD) -o $@
+
+$(tools_policyexamplegen_OBJECTS): CC=$(CC_FOR_BUILD)
+$(tools_policyexamplegen_OBJECTS): CFLAGS=$(CFLAGS_FOR_BUILD)  $(TOOLS_CFLAGS)
+$(tools_policyexamplegen_OBJECTS): LDFLAGS=$(LDFLAGS_FOR_BUILD)
+else
+tools_tableidxgen_LDADD = $(LDADD)
```
juergenchrist, line 26: If we cross-compile, we have not additional link-time libraries, but if we do not, we do?  Why?

tools/tools.mk
```diff
@@ -1,7 +1,33 @@
...
+tools_tableidxgen_LINK = $(LINK)
+tools_tableidxgen_CFLAGS = $(AM_CFLAGS) $(TOOLS_CFLAGS)
+
+tools_policyexamplegen_LDADD = $(LDADD)
```
juergenchrist, line 30: If we cross-compile, we have not additional link-time libraries, but if we do not, we do?  Why?

juergenchrist (changes requested):

tools/tools.mk
ifranzki: no, no reason. I just copied that from somewhere else :-) 
What would you suggest instead? The point is that we need to use $(CC_FOR_BUILD) for building this.

tools/tools.mk
ifranzki: hm, right, that line is probably superfluous at all.

tools/tools.mk
ifranzki: also superfluous.

tools/tools.mk
juergenchrist: How about simply
```
$(CC_FOR_BUILD) $(CFLAGS_FOR_BUILD) $(LDFLAGS_FOR_BUILD) -o $@
```

configure.ac
```diff
@@ -595,6 +597,7 @@ CFLAGS+=" -std=c99 -Werror=implicit-int -Werror=implicit-function-declaration -W
 
 # At this point, CFLAGS is set to something sensible
 AC_PROG_CC
+AX_PROG_CC_FOR_BUILD
```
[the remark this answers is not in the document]
ifranzki: I did not need to install anything to use this. So looks like the distros include this by default?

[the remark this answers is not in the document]
ifranzki: Although I think we should be allowed to include this file into our package, I added a note to README.md about how to install it or where to download it in case it is not available anyway. I case this causes trouble with distributors later on, we can still decide to include it into the m4 directory.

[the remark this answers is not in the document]
ifranzki: also added it a s build requirement to the example RPM spec file.

juergenchrist (approved):

mhartmay:
`Tested-by: Marc Hartmayer <marc@linux.ibm.com>`

PR #639, opened 2023-01-20
EP11: Correct precondition checks for protected key usage and creation

ifranzki:
Using an existing and valid protected key attached to a key object is not dependent on the session state. Only if the protected key must be (re-)created (secure key to protected key transformation), a proper session state is required.

PKCS#11-UG states:
_"A session can be a read/write (R/W) session or a read-only (R/O) session. Read/write and read-only refer to the access to token objects, not to session objects. In both session types, an application can create, read, write and destroy session objects, and read token objects. However, only in a read/write session can an application create, modify, and destroy token objects."_

Allow the transformation of a key object into a protected key for session objects also in a read-only session. As per above, session objects can be updated in read-only session.

Furthermore PKCS#11-UG states for SO-RW sessions (CKS_RW_SO_FUNCTIONS):
_The Security Officer has been authenticated to the token. The application has read/write access only to public objects on the token, not to private objects._

Thus, in a SO read/write session, private key objects can not be transformed into a protected key.

### DIFF
--- a/usr/lib/ep11_stdll/ep11_specific.c
+++ b/usr/lib/ep11_stdll/ep11_specific.c
@@ -1004,15 +1004,28 @@ done:
 
 /**
  * Returns true if the session is ok for creating protected keys, false
- * otherwise. The session must be read/write and not public, because the
- * protected key is a private/secret key attribute.
+ * otherwise. The session must be read/write for token objects, and not public
+ * nor SO for private objects.
  */
-static CK_BBOOL ep11tok_pkey_session_read_write(SESSION *session)
+static CK_BBOOL ep11tok_pkey_session_ok_for_obj(SESSION *session,
+                                                OBJECT *key_obj)
 {
-    if (session->session_info.flags & CKF_RW_SESSION)
-        return CK_TRUE;
+    if (object_is_token_object(key_obj) &&
+        (session->session_info.flags & CKF_RW_SESSION) == 0)
+        return CK_FALSE;
 
-    return CK_FALSE;
+    if (object_is_private(key_obj)) {
+        switch (session->session_info.state) {
+        case CKS_RO_PUBLIC_SESSION:
+        case CKS_RW_PUBLIC_SESSION:
+        case CKS_RW_SO_FUNCTIONS:
+            return CK_FALSE;
+        default:
+            break;
+        }
+    }
+
+    return CK_TRUE;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -1083,16 +1096,19 @@ CK_RV ep11tok_pkey_check(STDLL_TokData_t *tokdata, SESSION *session,
...
+
         if (template_attribute_get_non_empty(key_obj->template,
                                              CKA_IBM_OPAQUE_PKEY,
                                              &opaque_attr) != CKR_OK ||
             !ep11tok_pkey_is_valid(tokdata, key_obj)) {
             /* this key has either no pkey attr, or it is not valid,
-             * try to create one */
+             * try to create one, if the session state allows it */
+            if (!ep11tok_pkey_session_ok_for_obj(session, key_obj))
+                goto done;
```
jschmidb, line 1110: Why did you move this check down? You could do the check above in line 1086 as before and would not call template_attribute_get_non_empty if the session is not ok for the obj.

ifranzki: _Using an existing and valid protected key attached to a key object is not dependent on the session state._

Thats why I moved it down. You only need to check the session state if the protected key needs to be re-derived, not when it is already available and valid.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -1004,15 +1004,28 @@ static CK_RV ep11tok_pkey_update(STDLL_TokData_t *tokdata, SESSION *session,
...
 {
-    if (session->session_info.flags & CKF_RW_SESSION)
-        return CK_TRUE;
+    if (object_is_token_object(key_obj) &&
+        (session->session_info.flags & CKF_RW_SESSION) == 0)
+        return CK_FALSE;
 
-    return CK_FALSE;
+    if (object_is_private(key_obj)) {
+        switch (session->session_info.state) {
```
jschmidb, line 1018: Would this check "object_is_private" also apply to AES keys? Do they also have the CKA_PRIVATE attribute? But yes, we saw that there is no routine like "is_secret_key()" in object.c. Just to mention ...

ifranzki: Yes, attribute CKA_PRIVATE is defined by the PKCS#11 standard for all key objects (actually for all storage objects, where keys belong to).

There might be a misunderstanding: CKA_PRIVATE does **not** mean that its a private key in the sense of a private RSA key vs. a RSA public key. CKA_PRIVATE just means that this object is private to the USER pin, and thus is stored encrypted with the token's master key.

ifranzki: so even an RSA public key object may have CKA_PRIVATE=TRUE (although it does not make too much sense, since it does not contain any sensitive information).

jschmidb (approved):

vkanjur (approved):

PR #640, opened 2023-01-24
EP11: Support Dilithium round 3 variants and Kyber

ifranzki:


usr/lib/ep11_stdll/ep11.h
```diff
@@ -863,38 +727,46 @@ typedef enum {
...
 //---  importer PK types  ----------------------------------------------------
 typedef enum {
 	XCP_IMPRKEY_RSA_2048    = 0,
 	XCP_IMPRKEY_RSA_4096    = 1,
-	XCP_IMPRKEY_EC_P256     = 2,    /* EC, NIST P-256        */
-	XCP_IMPRKEY_EC_P521     = 3,    /* EC, NIST P-521        */
-	XCP_IMPRKEY_EC_BP256r   = 4,    /* EC, Brainpool BP-256r */
-	XCP_IMPRKEY_EC_BP320r   = 5,    /* EC, Brainpool BP-320r */
-	XCP_IMPRKEY_EC_BP512r   = 6,    /* EC, Brainpool BP-512r */
+	XCP_IMPRKEY_EC_P256     = 2,    /* EC, NIST P-256                     */
```
jschmidb, line 750: Formatting? tabs?

ifranzki: We take the ep11 header file(s) as is, no change (except the license header).

usr/lib/common/key.c
```diff
@@ -2656,6 +2656,117 @@ CK_RV rsa_priv_unwrap_get_data(TEMPLATE *tmpl,
...
+        *mode_attr = CKA_IBM_DILITHIUM_MODE;
+        *oids = dilithium_oids;
+        break;
+    case CKM_IBM_KYBER:
+        *keyform_attr = CKA_IBM_KYBER_KEYFORM;
+        *mode_attr = CKA_IBM_KYBER_MODE;
+        *oids = kyber_oids;
+        break;
+    default:
+        TRACE_ERROR("Unsupported mechanims\n");
```
jschmidb, line 2682: Print the mech type in the TRACE_ERROR?

testcases/crypto/dilithium_func.c
```diff
@@ -34,14 +35,48 @@ typedef struct signVerifyParam {
...
+
+const _variant_info variants[] = {
+    { "DEFAULT (DILITHIUM_R2_65)", 0, NULL, 0 },
+    { "DILITHIUM_R2_65", CK_IBM_DILITHIUM_KEYFORM_ROUND2_65,
+      dilithium_r2_65, dilithium_r2_65_len },
+    { "DILITHIUM_R2_87", CK_IBM_DILITHIUM_KEYFORM_ROUND2_87,
+      dilithium_r2_87, dilithium_r2_87_len },
+    { "DILITHIUM_R3_44", CK_IBM_DILITHIUM_KEYFORM_ROUND3_44,
+      dilithium_r3_44, dilithium_r3_44_len },
+    { "DILITHIUM_R3_56", CK_IBM_DILITHIUM_KEYFORM_ROUND3_65,
```
jschmidb, line 72: "DILITHIUM_R3_65" ?

usr/lib/common/asn1.c
```diff
@@ -4400,3 +4400,566 @@ CK_RV ber_decode_IBM_DilithiumPrivateKey(CK_BYTE *data,
...
+     *       SEQUENCE (2 elem)
+     *          BIT STRING  -> pk
+     */
+    rc = ber_encode_SEQUENCE(FALSE, data, data_len, buf3, total_len);
+    if (rc != CKR_OK)
+        TRACE_ERROR("%s ber_encode_Seq failed with rc=0x%lx\n", __func__, rc);
+
+error:
+
+    if (buf)
```
jschmidb, line 4549: As always ... the if is not technically necessary, because free() does a NULL check

usr/lib/common/key.c
```diff
@@ -3032,6 +3038,240 @@ CK_RV ibm_dilithium_priv_unwrap(TEMPLATE *tmpl, CK_BYTE *data,
...
+                            CK_BBOOL length_only,
+                            CK_BYTE **data, CK_ULONG *data_len)
+{
+    switch (keytype) {
+    case CKK_IBM_PQC_DILITHIUM:
+        return ibm_dilithium_publ_get_spki(tmpl, length_only, data, data_len);
+    case CKK_IBM_PQC_KYBER:
+        return ibm_kyber_publ_get_spki(tmpl, length_only, data, data_len);
+    default:
+        TRACE_DEVEL("Kytype 0x%lx not supported.\n", keytype);
```
jschmidb, line 3223: Key type

usr/lib/common/key.c
```diff
@@ -3032,6 +3038,240 @@ CK_RV ibm_dilithium_priv_unwrap(TEMPLATE *tmpl, CK_BYTE *data,
...
+        return CKR_KEY_TYPE_INCONSISTENT;
```
jschmidb, line 3224: Hmm interesting ... in a previous commit there was a similar case of a key type not being supported. There you returned CKR_MECHANISM_INVALID (in policy.c line 207/208). I searched for a "key type unsupported" rc, but such an rc does not exist. Here you decided to use CKR_KEY_TYPE_INCONSISTENT, which is not exactly what we would need ;-)

usr/lib/common/key.c
ifranzki: Should only be called for Dilithium or Kyber, but sure, printing the mech makes sense.

testcases/crypto/dilithium_func.c
ifranzki: good catch! will fix

usr/lib/common/asn1.c
ifranzki: yes, I know, but I tend to have the if anyway, makes it clearer that I expect it be NULL....

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -5214,6 +5214,126 @@ CK_RV token_specific_aes_cmac(STDLL_TokData_t *tokdata,
...
+                                            CK_MECHANISM *mech,
+                                            struct EP11_KYBER_MECH *mech_ep11,
+                                            OBJECT **secret_key_obj)
+{
+    CK_IBM_KYBER_PARAMS *kyber_params;
+    CK_RV rc;
+
+    kyber_params = mech->pParameter;
+    if (mech->ulParameterLen != sizeof(CK_IBM_KYBER_PARAMS)) {
+        TRACE_ERROR("Mechanism parameter length not as epxeted\n");
```
jschmidb, line 5232: expected

usr/lib/common/key.c
ifranzki: fixed (in multiple occurances)

usr/lib/common/key.c
ifranzki: well, the caller of this function expects to deal with a key type of Dilithium or Kyber, so if it is not one of that, the key type is inconsistent with the operation performed. CKR_MECHANISM_INVALID would be totally wrong here. 

Yes for policy.c line 207/208 it should probably also be CKR_KEY_TYPE_INCONSISTENT.

testcases/common/common.c
```diff
@@ -735,6 +735,125 @@ CK_RV create_DilithiumPublicKey(CK_SESSION_HANDLE session,
     return rc;
 }
 
+/** Create an IBM Dilithium private key using private values **/
```
jschmidb, line 738: IBM Kyber private key

testcases/common/common.c
```diff
@@ -735,6 +735,125 @@ CK_RV create_DilithiumPublicKey(CK_SESSION_HANDLE session,
...
+        else if (is_rejected_by_policy(rc, session))
+            rc = CKR_POLICY_VIOLATION;
+        else
+            testcase_error("C_CreateObject rc=%s", p11_get_ckr(rc));
+    }
+
+    return rc;
+}
+
+/** Create an IBM Kyber public key using   **/
```
jschmidb, line 804: using public values ?

testcases/crypto/kyber_func.c
```diff
@@ -0,0 +1,1176 @@
...
+    CK_ULONG keyform;
+    const CK_BYTE *oid;
+    CK_ULONG oid_len;
+} _variant_info;
+
+const _variant_info variants[] = {
+    { "DEFAULT (KYBER_R2_1024)", 0, NULL, 0 },
+    { "KYBER_R2_768", CK_IBM_KYBER_KEYFORM_ROUND2_768,
+      kyber_r2_768, kyber_r2_768_len },
+    { "KYBER_R2_1024", CK_IBM_KYBER_KEYFORM_ROUND2_768,
```
jschmidb, line 50: CK_IBM_KYBER_KEYFORM_ROUND2_1024 ?

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: fixed

testcases/common/common.c
ifranzki: fixed

testcases/common/common.c
ifranzki: fixed

testcases/crypto/kyber_func.c
ifranzki: oups, yes.... fixed.

jschmidb (approved):

PR #642, opened 2023-01-25
EP11: Fix setting unknown CPs to ON

ifranzki:
The very last control point must also be applied from the queried bits to the combined bits. Otherwise the very last control point is always treated as being ON, although it might be OFF, and this can lead to mechanisms being used that are disabled by that control point.

Fixes https://github.com/opencryptoki/opencryptoki/commit/97248f73495695436f11fafd74c2ec41a5a6f796

This caused failures in the CI last night.


jschmidb (approved):
lgtm

PR #643, opened 2023-01-25
EP11: Enable AES XTS mechanism for protected keys

vkanjur:


usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -2870,6 +2880,43 @@ static int get_curve_type_from_template(TEMPLATE *tmpl)
...
+
+CK_RV ep11tok_aes_xts_pkey_check(STDLL_TokData_t *tokdata, CK_ATTRIBUTE_PTR attrs,
+                                 CK_ULONG attrs_len)
+{
+    ep11_private_data_t *ep11_data = tokdata->private_data;
+
+    if (ep11tok_pkey_option_disabled(tokdata) == TRUE ||
+        ep11_data->msa_level < 4 ||
+        ep11tok_is_mechanism_supported(tokdata, CKM_IBM_CPACF_WRAP) != CKR_OK ||
+        ep11tok_is_mechanism_supported(tokdata, CKM_AES_KEY_GEN) != CKR_OK) {
```
ifranzki, line 3014: Instead of checking these 4 preconditions here, you could simply use
`if (ep11tok_is_mechanism_supported(tokdata, CKM_AES_XTS_KEY_GEN) == CKR_OK) ....` which already checks the same things (you introduces this only a few commits later, but that should not matter).

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -2917,6 +2922,175 @@ CK_RV ep11tok_aes_xts_pkey_check(STDLL_TokData_t *tokdata, CK_ATTRIBUTE_PTR attr
...
+    CK_RV rc;
+    CK_ATTRIBUTE_PTR p_attrs = NULL;
+    CK_ULONG attrs_len = 0;
+    CK_ATTRIBUTE_PTR new_p_attrs = NULL;
+    CK_ULONG new_attrs_len = 0;
+    CK_ATTRIBUTE *chk_attr = NULL, *attr = NULL;
+    unsigned char *ep11_pin_blob = NULL;
+    CK_ULONG ep11_pin_blob_len = 0;
+    ep11_session_t *ep11_session = (ep11_session_t*) sess->private_data;
+
```
ifranzki, line 3052: Should we maybe also check here if XTS is supported at all? I.e. if msa level is correct, and pkey is not disabled? Otherwise we import an XTS key but can't use it.

Maybe add
`if (ep11tok_is_mechanism_supported(tokdata, CKM_AES_XTS_KEY) != CKR_OK) ....`
here and return CKR_KEY_TYPE_INCONSISTENT if the XTS mech is not supported.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -9126,7 +9245,14 @@ static CK_RV ep11_ende_crypt_init(STDLL_TokData_t * tokdata, SESSION * session,
         free(ep11_state);
         goto done;
     case CKR_FUNCTION_NOT_SUPPORTED:
-        /* fallback into ep11 path */
+        /* if mechanism is AES XTS, return error
+         * else fallback to ep11 path */
```
ifranzki, line 9271: Style: can you put this comment into one line? 

Multi line comments should be like this:
```
/*
 * comment ...
 */
```

testcases/crypto/aes_func.c
```diff
@@ -69,6 +69,11 @@ CK_RV do_EncryptDecryptAES(struct generated_test_suite_info *tsuite)
         goto testcase_cleanup;
     }
 
+    /* Skip tests if pkey = false, but the slot doesn't support CKM_AES_XTS*/
```
ifranzki, line 72: Style: Add a blank please before the closing `*/`
Same for all places further down.

testcases/crypto/aes_func.c
```diff
@@ -1561,6 +1606,14 @@ CK_RV do_WrapUnwrapRSA(struct generated_test_suite_info * tsuite)
         goto testcase_cleanup;
     }
 
+    if (!wrap_supported(slot_id, tsuite->mech)) {
+        testsuite_skip(3, "Slot %u doesn't support %s (0x%x)",
```
ifranzki, line 1610: Maybe make the text a bit more precise:
`"Slot %u doesn't support wrapping with %s (0x%x)"`
(it may support the mechanisms as such, but just not wrapping with it).

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -13448,11 +13458,27 @@ CK_RV token_specific_set_attribute_values(STDLL_TokData_t *tokdata,
...
+                            __func__, rc);
+                free(ibm_opaque_attr);
+                goto out;
+            }
+        }
         rc = template_update_attribute(new_tmpl, ibm_opaque_attr);
         if (rc != CKR_OK) {
             TRACE_ERROR("%s template_update_attribute failed rc=0x%lx\n",
                         __func__, rc);
-            free(ibm_opaque_attr);
```
ifranzki, line 13006: Why was this line removed? I think this should still be freed here.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -6034,6 +6035,20 @@ CK_RV ep11tok_derive_key(STDLL_TokData_t *tokdata, SESSION *session,
...
     }
+
+    rc = template_attribute_get_ulong(base_key_obj->template, CKA_KEY_TYPE, &keytype);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("Invalid key type attribute\n");
+        goto error;
+    }
+
+    if (mech->mechanism == CKM_AES_XTS || keytype == CKK_AES_XTS) {
+        TRACE_ERROR("%s Key derive not supported\n",
```
ifranzki, line 6046: Please make the trace message a bit more procise, like 
"%s Key derive with AES-XTS is not supported\n"

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -6057,6 +6072,13 @@ CK_RV ep11tok_derive_key(STDLL_TokData_t *tokdata, SESSION *session,
         goto error;
     }
 
+    if (ktype == CKK_AES_XTS) {
+        TRACE_ERROR("%s Derived keytype not supported\n",
```
ifranzki, line 6076: also here:
"%s Deriving an AES-XTS key is not supported\n"

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -9444,6 +9467,19 @@ CK_RV ep11tok_wrap_key(STDLL_TokData_t * tokdata, SESSION * session,
         return rc;
     }
 
+    rc = template_attribute_get_ulong(wrap_key_obj->template, CKA_KEY_TYPE, &ktype);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("Invalid key type attribute\n");
+        return rc;
```
ifranzki, line 9473: this must be `goto done;` instead of return, otherwise we leak the wrap_key_obj.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -9444,6 +9467,19 @@ CK_RV ep11tok_wrap_key(STDLL_TokData_t * tokdata, SESSION * session,
...
+    }
+
+    if (mech->mechanism == CKM_AES_XTS || ktype == CKK_AES_XTS) {
+        TRACE_ERROR("%s Key wrap not supported\n",
```
ifranzki, line 9477: "%s Key wrap with AES-XTS is not supported\n"

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -9459,6 +9495,18 @@ CK_RV ep11tok_wrap_key(STDLL_TokData_t * tokdata, SESSION * session,
                     rc);
         goto done;
     }
+    rc = template_attribute_get_ulong(key_obj->template, CKA_KEY_TYPE, &ktype);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("Invalid key type attribute\n");
+        return rc;
```
ifranzki, line 9501: this must be goto done; instead of return, otherwise we leak the wrap_key_obj.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -9459,6 +9495,18 @@ CK_RV ep11tok_wrap_key(STDLL_TokData_t * tokdata, SESSION * session,
...
+    }
+
+    if (ktype == CKK_AES_XTS) {
+        TRACE_ERROR("%s Key wrap not supported\n",
```
ifranzki, line 9505: "%s Wrapping an AES-XTS key is not supported\n"

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -9587,6 +9636,20 @@ CK_RV ep11tok_unwrap_key(STDLL_TokData_t * tokdata, SESSION * session,
                     __func__, rc);
         return rc;
     }
+
+    rc = template_attribute_get_ulong(kobj->template, CKA_KEY_TYPE, &keytype);
+        if (rc != CKR_OK) {
+            TRACE_ERROR("Invalid key type attribute\n");
+            return rc;
```
ifranzki, line 9643: this must be goto done; instead of return, otherwise we leak the kobj.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -9587,6 +9636,20 @@ CK_RV ep11tok_unwrap_key(STDLL_TokData_t * tokdata, SESSION * session,
...
+        }
+
+    if (mech->mechanism == CKM_AES_XTS || keytype == CKK_AES_XTS) {
+        TRACE_ERROR("%s Key wrap not supported\n",
```
ifranzki, line 9647: "%s Key unwrap with AES-XTS is not supported\n"

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -9698,6 +9761,13 @@ CK_RV ep11tok_unwrap_key(STDLL_TokData_t * tokdata, SESSION * session,
         goto error;
     }
 
+    if (ktype == CKK_AES_XTS) {
+        TRACE_ERROR("%s Key unwrap not supported\n",
```
ifranzki, line 9765: "%s Unwrapping an AES-XTS key is not supported\n"

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -598,7 +598,8 @@ static CK_RV get_ep11_target_for_apqn(uint_32 adapter, uint_32 domain,
                                       target_t *target, uint64_t flags);
 static void free_ep11_target_for_apqn(target_t target);
 static CK_RV update_ep11_attrs_from_blob(STDLL_TokData_t *tokdata,
-                                         SESSION * session, TEMPLATE *tmpl);
+                                         SESSION * session, TEMPLATE *tmpl,
```
ifranzki, line 601: Syle: please make this `SESSION *session`. I know it was that before, but while you changing this line, lets correct it.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -4243,6 +4304,40 @@ CK_RV ep11tok_generate_key(STDLL_TokData_t * tokdata, SESSION * session,
...
+                                   ep11_pin_blob, ep11_pin_blob_len, blob2,
+                                   &blobsize2, csum, &csum_len,
+                                   target_info->target);
+        RETRY_END(rc, tokdata, session)
+        if (rc != CKR_OK) {
+            rc = ep11_error_to_pkcs11_error(rc, session);
+            TRACE_ERROR("%s m_GenerateKey rc=0x%lx mech='%s' attrs_len=0x%lx\n",
+                        __func__, rc, ep11_get_ckm(tokdata, mech->mechanism),
+                        attrs_len);
+            goto done;
```
ifranzki, line 4605: This should be `goto error;`, not `goto done;`.
Also above in line 4295. This was wrong before already, but lets correct it at both places please.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -2917,6 +2922,175 @@ CK_RV ep11tok_aes_xts_pkey_check(STDLL_TokData_t *tokdata, CK_ATTRIBUTE_PTR attr
...
+        TRACE_INFO("%s encrypt ksize=0x%lx clen=0x%lx rc=0x%lx\n", __func__,
+                   attr->ulValueLen / 2, clen, rc);
+    }
+
+    rc = check_key_attributes(tokdata, CKK_AES, CKO_SECRET_KEY, p_attrs,
+                              attrs_len, &new_p_attrs, &new_attrs_len, -1);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("%s AES XTS check private key attributes failed with "
+                    "rc=0x%lx\n", __func__, rc);
+        return rc;
```
ifranzki, line 3102: This should be `goto import_aes_xts_key_end;` instead of the return.
This was already wrong in `rawkey_2_blob()` where you copied it from ;-)

If you like, add an additional commit to this PR fixing this in `rawkey_2_blob()`.
The commit title would be for example `EP11: Fix potential memory leak when importing symmetric keys`.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -2917,6 +2922,175 @@ CK_RV ep11tok_aes_xts_pkey_check(STDLL_TokData_t *tokdata, CK_ATTRIBUTE_PTR attr
...
+    /* tell ep11 the attributes the user specified */
+    rc = build_ep11_attrs(tokdata, aes_xts_key_obj->template, &p_attrs,
+                          &attrs_len, CKK_AES, CKO_SECRET_KEY, -1, &mech);
+    if (rc != CKR_OK)
+        goto import_aes_xts_key_end;
+
+    rc = ep11tok_aes_xts_pkey_check(tokdata, p_attrs, attrs_len);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("%s EP11 AES XTS is not supported: rc=0x%lx\n",
+                    __func__, rc);
```
ifranzki, line 3069: does this fit into one line ?

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -2917,6 +2922,175 @@ CK_RV ep11tok_aes_xts_pkey_check(STDLL_TokData_t *tokdata, CK_ATTRIBUTE_PTR attr
...
+                             blob, blob_size, csum, &cslen, target_info->target);
+    RETRY_END(rc, tokdata, sess)
+
+    if (rc != CKR_OK) {
+        rc = ep11_error_to_pkcs11_error(rc, sess);
+        TRACE_ERROR("%s unwrap blen=%zd rc=0x%lx\n", __func__, *blob_size, rc);
+        goto import_aes_xts_key_end;
+    } else {
+        TRACE_INFO("%s unwrap blen=%zd rc=0x%lx\n", __func__, *blob_size,
+                   rc);
```
ifranzki, line 3126: does this fit into one line ?

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -2917,6 +2922,175 @@ CK_RV ep11tok_aes_xts_pkey_check(STDLL_TokData_t *tokdata, CK_ATTRIBUTE_PTR attr
...
+                             ep11_data->raw2key_wrap_blob_l, NULL, ~0,
+                             ep11_pin_blob, ep11_pin_blob_len, &mech,
+                             new_p_attrs, new_attrs_len, blob + *blob_size,
+                             &blob_size2, csum, &cslen, target_info->target);
+    RETRY_END(rc, tokdata, sess)
+
+    if (rc != CKR_OK) {
+        rc = ep11_error_to_pkcs11_error(rc, sess);
+        TRACE_ERROR("%s unwrap blen=%zd rc=0x%lx\n", __func__, blob_size2,
+                    rc);
```
ifranzki, line 3173: does this fit into one line ?

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -2917,6 +2922,175 @@ CK_RV ep11tok_aes_xts_pkey_check(STDLL_TokData_t *tokdata, CK_ATTRIBUTE_PTR attr
...
+    } else {
+        TRACE_INFO("%s unwrap blen=%zd rc=0x%lx\n", __func__, blob_size2,
+                   rc);
```
ifranzki, line 3176: does this fit into one line ?

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: and there is a `goto import_aes_xts_key_end;` missing in the error case.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -4142,6 +4316,16 @@ CK_RV token_specific_object_add(STDLL_TokData_t * tokdata, SESSION * sess,
                    __func__, rc, blobsize);
```
ifranzki, line 4419: please remove this empty line while you are at it.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -2870,6 +2880,43 @@ static int get_curve_type_from_template(TEMPLATE *tmpl)
     return curve_type;
 }
 
+CK_BBOOL ep11_check_pkey_attrs(CK_ATTRIBUTE_PTR attrs,
```
jschmidb, line 2990: Naming: currently all the protected key related routines in ep11_specific.c are grouped together as ep11tok_pkey_xxxx. I would rename this routine to ep11tok_pkey_check_attrs. Also the routine below: ep11tok_pkey_aes_xts_check or ep11tok_pkey_check_aes_xts. And I would copy them into the block of the other ep11tok_pkey_xxx routines.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -2917,6 +2922,175 @@ CK_RV ep11tok_aes_xts_pkey_check(STDLL_TokData_t *tokdata, CK_ATTRIBUTE_PTR attr
...
+                    "rc=0x%lx\n", __func__, rc);
+        return rc;
+    }
+
+    trace_attributes(__func__, "Import sym.:", new_p_attrs, new_attrs_len);
+
+    ep11_get_pin_blob(ep11_session, object_is_session_object(aes_xts_key_obj),
+                      &ep11_pin_blob, &ep11_pin_blob_len);
+
+    /* the encrypted key is decrypted and a blob is build,
```
jschmidb, line 3110: is built

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -2917,6 +2922,175 @@ CK_RV ep11tok_aes_xts_pkey_check(STDLL_TokData_t *tokdata, CK_ATTRIBUTE_PTR attr
...
+        TRACE_INFO("%s encrypt ksize=0x%lx clen=0x%lx rc=0x%lx\n", __func__,
+                   attr->ulValueLen / 2, clen, rc);
+    }
+
+    trace_attributes(__func__, "Import sym.:", new_p_attrs, new_attrs_len);
+
+    /* update the remaining buffer size in blob */
+    blob_size2 = blob_size2 - *blob_size;
+
+    /* the encrypted key is decrypted and a blob is build,
```
jschmidb, line 3158: is built

usr/lib/common/pkey_utils.c
```diff
@@ -750,6 +754,170 @@ CK_RV pkey_aes_cmac(OBJECT *key_obj, CK_BYTE *message,
...
+
+    switch (keylen) {
+    case 32:
+        fc = ENCRYPTED_AES_XTS_128;
+        break;
+    case 64:
+        fc = ENCRYPTED_AES_XTS_256;
+        break;
+    default:
+        return 0;
```
jschmidb, line 773: Just make sure that the 0 is handled as an error in calling routines. 0 is a valid function code (all the query functions, e.g. KM_QUERY).

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -902,9 +943,52 @@ static CK_RV ep11tok_pkey_skey2pkey(STDLL_TokData_t *tokdata, SESSION *session,
         goto done;
     }
 
+    if (wk->token_size > sizeof(wrapped_key)) {
+        TRACE_ERROR("Wrapped key size too big\n");
```
jschmidb, line 947: "token size too big"? or "wrapped key size too small"? Or simply, as the return code says: "Buffer too small"?

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -10412,6 +10429,19 @@ CK_RV ep11tok_is_mechanism_supported(STDLL_TokData_t *tokdata,
             goto out;
         }
         break;
+
+    case CKM_AES_XTS:
+    case CKM_AES_XTS_KEY_GEN:
+        if (ep11tok_pkey_option_disabled(tokdata) != FALSE ||
```
jschmidb, line 10505: I would check for '== TRUE' here

ifranzki:
I noticed that testcases/misc_tests/p11sak_test.sh fails when run against the EP11 token (`SLOT=4 PKCS11_USER_PIN=<your-pin> ./p11sak_test.sh`) because it attempts to generate an AES-XTS key, but it gets `CKR_TEMPLATE_INCONSISTENT`, because the attributes are not set so that protected keys are allowed. This is using the default EP11 token config, thus `PKEY_MODE DEFAULT`. With `PKEY_MODE ENABLE4NONEXTR` it would certaily work, but this requires manual user actions before running the testcase.

Unfortunately the current version of p11sak does not allow to set the CKA_IBM_PROTKEY_EXTRACTABLE attribute during key-gen. I have updates for p11sak in the pipe that will allow to set that....

For now, lets exclude XTS tests when running against the EP11 token.

Please include the following update into you testcase commit, or add a new commit with this change.

```
 testcases/misc_tests/p11sak_test.sh | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

diff --git a/testcases/misc_tests/p11sak_test.sh b/testcases/misc_tests/p11sak_test.sh
index ae7fd9dc..49a5a75a 100755
--- a/testcases/misc_tests/p11sak_test.sh
+++ b/testcases/misc_tests/p11sak_test.sh
@@ -76,7 +76,7 @@ p11sak generate-key aes 128 --slot $SLOT --pin $PKCS11_USER_PIN --label p11sak-a
 p11sak generate-key aes 192 --slot $SLOT --pin $PKCS11_USER_PIN --label p11sak-aes-192
 p11sak generate-key aes 256 --slot $SLOT --pin $PKCS11_USER_PIN --label p11sak-aes-256
 # aes-xts [128 | 256]
-if [[ -n $( pkcsconf -m -c $SLOT | grep CKM_AES_XTS_KEY_GEN) ]]; then
+if [[ -n $( pkcsconf -m -c $SLOT | grep CKM_AES_XTS_KEY_GEN) && -z $( pkcsconf -t -c $SLOT | grep "Model: EP11") ]]; then
 	p11sak generate-key aes-xts 128 --slot $SLOT --pin $PKCS11_USER_PIN --label p11sak-aes-xts-128
 	p11sak generate-key aes-xts 256 --slot $SLOT --pin $PKCS11_USER_PIN --label p11sak-aes-xts-256
 else
@@ -385,7 +385,7 @@ else
 	status=1
 fi
 
-if [[ -n $( pkcsconf -m -c $SLOT | grep CKM_AES_XTS_KEY_GEN) ]]; then
+if [[ -n $( pkcsconf -m -c $SLOT | grep CKM_AES_XTS_KEY_GEN) && -z $( pkcsconf -t -c $SLOT | grep "Model: EP11") ]]; then
 	# check AES-XTS 128
 	grep -q 'p11sak-aes-xts-128' $P11SAK_AES_XTS_PRE
 	rc=$?
```

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: or just 
```
if (ep11tok_pkey_option_disabled(tokdata) ||
....
```

testcases/crypto/aes_func.c
```diff
@@ -218,6 +223,12 @@ CK_RV do_EncryptDecryptUpdateAES(struct generated_test_suite_info * tsuite)
         goto testcase_cleanup;
     }
 
+    /* Skip tests if pkey = false, but the slot doesn't support CKM_AES_XTS*/
+    if (is_ep11_token(slot_id) && tsuite->mech.mechanism == CKM_AES_XTS && pkey == FALSE) {
+        testcase_skip("Slot doesn't support CKM_AES_XTS for non protected keys\n");
```
jschmidb, line 228: I would write: "Slot supports AES-XTS only for protected keys.\n" Also at the other places.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -1259,6 +1260,38 @@ CK_BBOOL ep11tok_pkey_usage_ok(STDLL_TokData_t *tokdata, SESSION *session,
...
+
+CK_RV ep11tok_pkey_check_aes_xts(STDLL_TokData_t *tokdata, CK_ATTRIBUTE_PTR attrs,
+                                 CK_ULONG attrs_len)
+{
+    if (ep11tok_is_mechanism_supported(tokdata, CKM_AES_XTS_KEY_GEN) != CKR_OK) {
+        TRACE_ERROR("%s\n", ock_err(ERR_MECHANISM_INVALID));
+        return CKR_MECHANISM_INVALID;
+    }
+
+    if (ep11tok_pkey_check_attrs(attrs, attrs_len, CKA_EXTRACTABLE) ||
```
ifranzki, line 1389: Is there a reason why function `ep11tok_pkey_check_aes_xts()` gets the attribute array and number of attributes, and not simply the key object (`OBJECT *key_obj`)? At both places where `ep11tok_pkey_check_aes_xts()` is called (key gen and import), the to-be-created object is already available (`object_mgr_create_skel()` has already been done). 

If `ep11tok_pkey_check_aes_xts()` would get the OBJECT, you would not need function `ep11tok_pkey_check_attrs()`  at all, but simply could use the following (like it is done in `ep11tok_pkey_check()` ):

```
        if (object_is_extractable(key_obj) ||
            !object_is_pkey_extractable(key_obj) ||
            object_is_attr_bound(key_obj))
....
```

This will result to the same behavior as today, but its a little less code, and even more important, the code is consistent to the already existing pkey related code.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -2912,6 +2917,177 @@ static int get_curve_type_from_template(TEMPLATE *tmpl)
...
+        return rc;
+    }
+
+    /* tell ep11 the attributes the user specified */
+    rc = build_ep11_attrs(tokdata, aes_xts_key_obj->template, &p_attrs,
+                          &attrs_len, CKK_AES, CKO_SECRET_KEY, -1, &mech);
+    if (rc != CKR_OK)
+        goto import_aes_xts_key_end;
+
+    rc = ep11tok_pkey_check_aes_xts(tokdata, p_attrs, attrs_len);
```
ifranzki, line 3067: Function ep11tok_pkey_check_aes_xts() checks if CKM_AES_XTS_KEY_GEN is supported, which is somewhat superfluous for importing. Import checks for CKM_AES_XTS being supported above already (which is correct).

Maybe we can remove the check for CKM_AES_XTS_KEY_GEN being supported from `ep11tok_pkey_check_aes_xts()` and just check for the respective mechanism being supported in the key gen function and import function individually? That way function `ep11tok_pkey_check_aes_xts()` is pretty much superfluous as well, since it only consist of 
```
        if (object_is_extractable(key_obj) ||
            !object_is_pkey_extractable(key_obj) ||
            object_is_attr_bound(key_obj))
....
```
(after the change of my previous comment).

ifranzki: Or alternatively pass the mechanism to check to `ep11tok_pkey_check_aes_xts()`, then you can use that function from keygen as well as from import, and you don't need the mechanism check in import separately....

vkanjur:
@ifranzki , @jschmidb : I have worked on your comments. Also added two more commits.

testcases/misc_tests/tok2tok_transport.c
```diff
@@ -664,6 +664,12 @@ CK_RV do_wrap_key_test(struct wrapped_mech_info *tsuite,
 
     case CKM_AES_KEY_GEN:
     case CKM_AES_XTS_KEY_GEN:
+        if ((is_ep11_token(slot_id1) || is_ep11_token(slot_id2))
+            && tsuite->wrapped_key_gen_mech.mechanism == CKM_AES_XTS_KEY_GEN){
+            testcase_skip("Skipping as AES XTS is supported only with protected keys");
+            rc = CKR_OK;
+        goto testcase_cleanup;
```
ifranzki, line 671: Minor style issues (sorry that I did not notice this before):
- the `goto` should be indented.
- blank missing before `{` in line 668:  `){`
- I would usually put the `&&` at the end of the previous line.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -10387,6 +10404,18 @@ CK_RV ep11tok_is_mechanism_supported(STDLL_TokData_t *tokdata,
             goto out;
         }
         break;
+
+    case CKM_AES_XTS:
+    case CKM_AES_XTS_KEY_GEN:
+        if (ep11tok_pkey_option_disabled(tokdata) || ep11_data->msa_level < 4 ||
+            ep11tok_is_mechanism_supported(tokdata, CKM_IBM_CPACF_WRAP) != CKR_OK ||
+            ep11tok_is_mechanism_supported(tokdata, CKM_AES_KEY_GEN) != CKR_OK) {
```
jschmidb, line 10477: Maybe nothing for this pr, but we could change the logic of ep11tok_is_mechanism_supported() to return either 'true' or 'false', as the function name suggests. Currently the function either returns 0 if the mech is supported or CKR_MECHANISM_INVALID if it's not. So already only two return values. It would make the usage of the function more self-explaining.

ifranzki: Right, nothing for this PR. The intention of the return code of ep11tok_is_mechanism_supported() was that it could further be returned directly by the caller, as it is for example done in ep11tok_get_mechanism_info().

And actually, ep11tok_is_mechanism_supported() does return other return codes than CKR_MECHANISM_INVALID, namely CKR_FUNCTION_FAILED. Returning a bool would not allow to distinguish between a real error case and the 'not-supported' case.

jschmidb (approved):
lgtm

ifranzki (approved):
looks good now.

PR #645, opened 2023-01-31
EP11: Fix set attribute values function for Data or Certificate objects.

vkanjur:
Data or Certificate objects do not have a key type(CKA_KEY_TYPE). So, Key type should be fetched after the switch case to handle this.

Signed-off-by: Vishnudatha Kanjur <kanjur@ibm.com>


ifranzki (approved):

jschmidb:
lgtm.
You may want to add a blank char in the commit msg: key type(CKA_KEY_TYPE)

jschmidb (approved):

PR #648, opened 2023-02-02
ICSF token: Support CKM_TLS_PRE_MASTER_KEY_GEN and CKM_TLS_KEY_AND_MAC_DERIVE

ifranzki:
The code already did mostly support those mechanisms, just a few fixes/changes were required to make them really work.
Also add testcases to test these mechanisms.


testcases/crypto/ssl3_func.c
```diff
@@ -637,6 +637,209 @@ CK_RV do_SSL3_MultipleKeysDerive(CK_SESSION_HANDLE session)
     return rc;
 }
 
+CK_RV do_TLS_PreMasterKeyGen(CK_SESSION_HANDLE session)
+{
+    CK_MECHANISM mech;
+    CK_VERSION version;
+    CK_OBJECT_HANDLE h_key;
```
jschmidb, line 644: CK_OBJECT_HANDLE h_key = CK_INVALID_HANDLE;   ?

testcases/crypto/ssl3_func.c
```diff
@@ -637,6 +637,209 @@ CK_RV do_SSL3_MultipleKeysDerive(CK_SESSION_HANDLE session)
...
+        testcase_error("C_DestroyObject() failed");
+
+done:
+    return rc;
+}
+
+CK_RV do_TLS_MultipleKeysDerive(CK_SESSION_HANDLE session)
+{
+    CK_MECHANISM mech;
+    CK_OBJECT_HANDLE h_pm_secret;
```
jschmidb, line 688: also here: initialize with "invalid handle" ?

testcases/crypto/ssl3_func.c
```diff
@@ -637,6 +637,209 @@ CK_RV do_SSL3_MultipleKeysDerive(CK_SESSION_HANDLE session)
...
+    CK_OBJECT_HANDLE h_key;
+    CK_RV rc = CKR_OK;
+    CK_SLOT_ID slot_id = SLOT_ID;
+
+    testcase_begin("starting do_TLS_PreMasterKeyGen...\n");
+
+    version.major = 3;
+    version.minor = 3;
+
+    mech.mechanism = CKM_TLS_PRE_MASTER_KEY_GEN;
```
jschmidb, line 653: In the first commit, you corrected the definition for CKM_SSL3_PRE_MASTER_KEY_GEN, so looks like this never worked before and was never tested. Would it make sense to also add a test for this, i.e. in addition to CKM_TLS_PRE_MASTER_KEY_GEN?

ifranzki: I just corrected the flags of CKM_SSL3_PRE_MASTER_KEY_GEN in the mech table, but those flags are not really used anywhere. So this is just a cosmetic change.
And tests for CKM_SSL3_PRE_MASTER_KEY_GEN are already in ssl3_func.c.

testcases/crypto/ssl3_func.c
ifranzki: well, this function is an exact copy from above, jus the mechanism is changed. I tend to leave it as is, otherwise I would need to change it in all functions in that file.

testcases/crypto/ssl3_func.c
ifranzki: this function is an exact copy from above, jus the mechanism is changed. I tend to leave it as is, otherwise I would need to change it in all functions in that file.

jschmidb (approved):

PR #652, opened 2023-02-06
configure: Fix install location of tmpfiles.d config file

ifranzki:
Don't prefix with destination directory, as this would produce an invalid directory. This would for example cause rpmbuild to fail.
Autoconf prefixes it with the destination directory automatically.


vkanjur (approved):

jschmidb (approved):
lgtm ... besides of the Travis error. But this single line cannot cause the Travis errors.

PR #653, opened 2023-02-06
ICSF: misc fixes

ifranzki:
For wrapping/unwrapping with a symmetric wrapping key using mechanisms CKM_DES_CBC_PAD, CKM_DES3_CBC_PAD, or CKM_AES_CBC_PAD the IV (i.e. the mechanism parameter) must be passed to ICSF as well.

### DIFF
--- a/usr/lib/icsf_stdll/icsf.c
+++ b/usr/lib/icsf_stdll/icsf.c
@@ -3269,6 +3269,7 @@ int icsf_derive_key(LDAP * ld, int *reason, CK_MECHANISM_PTR mech,
             version = (CK_VERSION_PTR) (&params->pVersion);
             version->major = bvParam.bv_val[0];
             version->minor = bvParam.bv_val[1];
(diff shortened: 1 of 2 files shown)


jschmidb (approved):
lgtm. So this never worked before?

ifranzki:
> lgtm. So this never worked before?

Yep. Well, it worked in that sense that it wrapped the key with an unknown (all zero?) IV, and of course was also able to unwrap that key with the same unknown (all zero?) IV. But transferring the wrapped key to another place and there unwrapping it with the desired IV would simply result in a wrong key ....

vkanjur (approved):

ifranzki:
Found some more bugs in the ICSF token, please take a look at the additional commits as well.

jschmidb (approved):
Yes, looks reasonable.

vkanjur (approved):

PR #654, opened 2023-02-07
Fix compile error "error: initializer element is not constant"

ifranzki:
Some compiler versions report error "error: initializer element is not constant" for the PQC initializers.

Use sizeof() directly in the initializer, instead of declaring a global variable holding the size. Such a global variable is not always considered as a constant, although it is declared as const.


jschmidb (approved):
lgtm

vkanjur (approved):

PR #657, opened 2023-02-09
policy: Fix update of mechanism info for AES-XTS mechanisms

ifranzki:
AES-XTS keys use 2 keys, and thus the AES-XTS mechanisms reports the minimum and maximum key size in the mechanism info as 32 - 64 bytes (256 - 512 bits). However, the strength of such an AES-XTS key is just the half of its size, i.e. a 256 bit AES-XTS key has a strength of 128 bits only. Likewise a 512 bit AES-XTS key has a strength of 256 bits.

While the checking of the key strength when using an AES-XTS key is done correctly, the supported minimum key size in the mechanism list is not updated according to the minimum strength policy setting. So, if a minimum strength of 256 bits is configured, then the minimum key size of the AES-XTS mechanisms must be adjusted to 64 bytes (512 bits).

### DIFF
--- a/usr/lib/api/policy.c
+++ b/usr/lib/api/policy.c
@@ -558,7 +558,7 @@ static CK_RV policy_update_modexp(struct policy_private *pp,
 
 static CK_RV policy_update_symmetric(struct policy_private *pp,
                                      CK_MECHANISM_INFO_PTR info,
-                                     CK_BBOOL isbytes)
+                                     CK_BBOOL isbytes, CK_BBOOL isaesxts)
 {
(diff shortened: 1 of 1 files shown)


jschmidb:
Minor typo in commit msg:
AES-XTS keys use 2 keys, and thus the AES-XTS mechanisms reports   <- report
Otherwise lgtm.

jschmidb (approved):

vkanjur (approved):

ifranzki:
typo in commit msg fixed.

PR #658, opened 2023-02-13
pkcsconf: Ignore CKR_TOKEN_NOT_PRESENT when displaying info

ifranzki:
Certain options cause pkcsconf to also list slots that have no token present, i.e. those options that list slots (versus tokens). If options that list slots are combined with options that list tokens, do not fail when attempting to display token information on a slot that has no token present.


jschmidb (approved):
lgtm

vkanjur (approved):

PR #660, opened 2023-02-14
SOFT/ICA: Add min/max key sizes to HMAC mechanism info

ifranzki:
Add the minimum and maximum key sizes for the HMAC mechanisms in their mechanism info. The key sizes are according to the key sizes of the CKM_GENERIC_SECRET_KEY_GEN key generation mechanism, adjusted to the minimum key sizes for the digests.


jschmidb (approved):
lgtm

vkanjur (approved):

PR #661, opened 2023-02-20
ICA: Calculate private exponent when generating RSA keys

ifranzki:
The ICA token uses libica to generate RSA keys in CRT format. That way, the public key components n and e, as well as the private CTR components p, q, dp, dq, and qInverse are available, but not the private exponent d.

Calculate the private exponent d from the modulus n, the public exponent e, and the private key components p and q as follows:

ϕ(n) = (p − 1) * (q − 1) = n − p − q + 1. Then d = e ^−1 mod ϕ(n)

That way, an RSA key generated by the ICA token has all private key components, the mod-expo components as well as the CRT components.

### DIFF
--- a/usr/lib/ica_s390_stdll/ica_specific.c
+++ b/usr/lib/ica_s390_stdll/ica_specific.c
@@ -1848,6 +1848,81 @@ err_crtkey:
     return NULL;
 }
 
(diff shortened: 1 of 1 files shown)


jschmidb (approved):
lgtm

vkanjur (approved):

PR #663, opened 2023-02-21
CCA: Support EC public key import with compressed EC point

ifranzki:
The CCA host library needs an uncompressed EC point in the key value structure when importing an EC public key. Uncompress the EC point using OpenSSL in case CKA_EC_POINT contains an EC point in compressed form.


jschmidb (approved):
lgtm

PR #664, opened 2023-02-22
ICA: Fix private exponent calculation

ifranzki:
The 'p' component in the libica CRT private key is padded with 8 bytes. 
Correct the conversion of the 'p' component into an OpenSSL BIGNUM by skipping the first 8 bytes of padding.

Unfortunately this is no where documented, one needs to 'know' that there is an 8 bytes padding for 'p', 'dp', and 'qInverse'.

This fixes an error during private exponent calculation at RSA key generation when BN_mod_inverse() returns an error, reporting that no inverse exists, because the 'p' component is wrong.

Fixes: b9ae1f6efbef ("ICA: Calculate private exponent when generating RSA keys")

@jschmidb Maybe you could add information about this 8 bytes padding to the libica header file and/or manual?


jschmidb (approved):

jschmidb:
Will update libica docs! Thanks for the hint.

PR #670, opened 2023-03-14
COMMON: Create HSM_MK_CHANGE directory in RPM and at pkcsslotd startup

ifranzki:
Create the /var/lib/opencryptoki/HSM_MK_CHANGE/ directory during RPM install and check for it also during pkcsslotd startup.


ifranzki:
> Is it intended to leave this directory after a make uninstall?

Yes this is intended. We also keep the token directories (basically we keep everything under /var/lib/oprncryptoki), since this is the current state of the tokens, including the token keys. We leave it to the user to remove these directories if he/she really wants to. Automatically removing them might cause a loss of all the keys. One might want to update its OCK version and should not loose all of its token data while doing so.

jschmidb (approved):

PR #672, opened 2023-03-15
hsm_mk_change: Try to open existing lock file before creating it

ifranzki:
If the OCK_HSM_MK_CHANGE_LOCK_FILE has already been created by another user, just open it. Only if it is not existent, try to create it and set the permissions.

While open(... , O_CREAT | O_RDONLY) also works for an already existing file, setting the permissions may fail if the file is owned by another user. However, when the file already exists, we can assume that the permissions are set correctly already.


jschmidb:
Ah yes, this is the same construction as in libica. Lgtm.

jschmidb (approved):

PR #675, opened 2023-03-20
EP11: Fix statistics counting for PKEY-based operations

ifranzki:
For crypto operations performed via protected key, the corresponding counter for strength 0 is incremented, instead of the the counter for the desired key strength.

When a crypto request is performed via protected key, the key object is released and set to NULL when ep11tok_pkey_check() returnes CKR_OK, indicating that the operation can be performed via protected key.

When the statistics counter is to be incremented at the 'done' label, the INC_COUNTER macro gets the now NULL key object, and thus falls back to the 'no-key-available' strength index specified as very last parameter.

Get and remember the key's strength index prior to releasing the key object, so that  it can be passed to the INC_COUNTER macro as 'no-key-available' strength index in case of the operation was performed via protected key. That way, the counter for the desired strength is incremented.


jschmidb (approved):

vkanjur (approved):

ifranzki:
Fixed typos in commit message (no code change).

PR #676, opened 2023-03-22
CCA pkey option

jschmidb:
This pr introduces protected key support for the CCA token.


usr/lib/cca_stdll/cca_specific.c
```diff
@@ -2974,6 +2974,304 @@ static CK_BBOOL ccatok_pkey_option_disabled(STDLL_TokData_t *tokdata)
...
+                                         &skey_attr) != CKR_OK) {
+        TRACE_ERROR("This key has no blob: should not occur!\n");
+        ret = CKR_FUNCTION_FAILED;
+        goto done;
+    }
+
+    /* Transform the secure key into a protected key */
+    ret = ccatok_pkey_skey2pkey(tokdata, skey_attr, &pkey_attr);
+    if (ret != CKR_OK) {
+        TRACE_ERROR("protected key creation failed with rc=0x%lx\n",ret);
```
ifranzki, line 3053: Here you must take care about the case when it is failing because a concurrent MK change is active, and none of the APQNs have the desired MK to convert the blobs to protected keys. In this case the ioctl will return ENODEV I guess, but this will lead to CKR_FUNCTION_FAILED here. 
In such a case (quite likely during MK changes), you should NOT just fail, but fall back to secure key usage.

ifranzki: If we think into the future about AES-XTS support, that has no secure key fallback, we might want to better handle the MK change case, and add code to retry the conversion with the reenciphered blobs. But lets keep that in mind for now, and maybe add it on top of this patch queue.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -5046,9 +5056,13 @@ CK_RV token_specific_aes_ecb(STDLL_TokData_t * tokdata,
     unsigned char exit_data[1];
     CK_BYTE *local_out = out_data;
     CK_ATTRIBUTE *attr = NULL;
-    CK_MECHANISM mech = { CKM_AES_ECB, NULL, 0 };
     long int key_len;
     CK_RV rc;
+#ifndef NO_PKEY
+    CK_MECHANISM mech = { CKM_AES_ECB, NULL, 0 };
+#else
+    UNUSED(session);
```
ifranzki, line 5087: hm, I would make 2 #ifdef blocks here. One for variable declaration, and one for the UNUSED. 
UNUSED really does not belong to variable declaration, and hsould have ablank line between variaböles and UNUSED.

Same for the other functions below.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -3276,6 +3281,7 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
  *                     End of CCA protected key option
  *
  ******************************************************************************/
```
ifranzki, line 3306: How about the PKEY config option if NO_PKEY is defined? I would say we should still support it, but only accept the DISABLED setting. Also, in this case the default of course is DISABLED.

Also add a note to the config file description that protected keys are only available on the s390x (Linux on IBM Z) platform

configure.ac
```diff
@@ -245,17 +245,24 @@ dnl ---
...
+	  CFLAGS="$CFLAGS -m31 -DNO_PKEY"
 	  ;;
      *ppc64* | *x86_64*)
+	  CFLAGS="$CFLAGS -DNO_PKEY"
 	  ;;
      *ppc* | i*86*)
-	  CFLAGS="$CFLAGS -m32"
+	  CFLAGS="$CFLAGS -m32 -DNO_PKEY"
 	  ;;
+	 *)
```
ifranzki, line 263: indention seems wrong here. Tabs vs spaces ?

configure.ac
```diff
@@ -471,20 +478,23 @@ fi
...
+	if test "x$enable_pkey" = "xyes"; then 
+		AC_CHECK_HEADER([asm/pkey.h], [], [
+			enable_ccatok=no
 	])
+	fi
 else
 	enable_ccatok=no
 fi
 AM_CONDITIONAL([ENABLE_CCATOK], [test "x$enable_ccatok" = "xyes"])
+AM_CONDITIONAL([NO_PKEY], [test "x$enable_pkey" != "xyes"])
```
ifranzki, line 497: I would move this above to below the `case $target in` block. Thats where the value of enable_pkey is set.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -2722,6 +2722,29 @@ typedef struct {
     enum cca_token_type keytype;
 } pkey_wrap_handler_data_t;
 
+/*
+ * On older kernels, the PKEY_KBLOB2PROTK3 ioctl is not yet available, which is
+ * no problem at runtime: protected key support is then just not available.
```
ifranzki, line 2727: > protected key support is then just not available.

Are you sure that you current code behaves so? If the ioctl fails, all ends with CKR_FUNCTION_FAILED and no secure key fallback is taken.

Also on each key usage it would try again and again to use the ioctl and fail again.
So you should detect during init if the kernel supports that ioctl or not.
Is this maybe done when getting the firmware MKVP ?

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -2584,6 +2585,116 @@ CK_RV cca_load_config_file(STDLL_TokData_t *tokdata, char *conf_name)
     return rc;
 }
 
+/*******************************************************************************
```
ifranzki, line 2601: please remove this 'Begin CCA protected key option' comment. This does not add anything. We see from the function name and its usage for what it is.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -2695,6 +2703,282 @@ static CK_RV ccatok_pkey_add_attrs(STDLL_TokData_t * tokdata, TEMPLATE *template
...
+    struct cca_private_data *cca_data = tokdata->private_data;
+
+    if (cca_data->pkey_mode == PKEY_MODE_DISABLED)
+        return CK_TRUE;
+
+    return CK_FALSE;
+}
+
+/*
+ *                     End of CCA protected key option
```
ifranzki, line 3304: remove this "End of CCA protected key option" comment, too

configure.ac
jschmidb: Yes, this is already a mixture of tabs and spaces ... now tried to at least add the new lines in the same way. Done.

configure.ac
jschmidb: Done.

usr/lib/cca_stdll/cca_specific.c
jschmidb: > Is this maybe done when getting the firmware MKVP ?

Yes, routine ccatok_pkey_get_firmware_wkvp sets cca_data->pkey_wrap_supported = 1; if it could successfully determine the wkvp. This indicates if pkey support is available.

usr/lib/cca_stdll/cca_specific.c
jschmidb: Done.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -2695,6 +2703,282 @@ static CK_RV ccatok_pkey_add_attrs(STDLL_TokData_t * tokdata, TEMPLATE *template
...
+    long kek_identifier_1_length = 0;
+    long kek_identifier_2_length = 0;
+    long the_2nd_key_token_length = 0;
+
+    /* Build CPACF-exportable AES internal cipher key token */
+    rule_array_count = 5;
+    memcpy(rule_array, "INTERNAL" "NO-KEY  " "AES     " "CIPHER  " "XPRTCPAC",
+           rule_array_count * CCA_KEYWORD_SIZE);
+
+    dll_CSNBKTB2(&return_code, &reason_code, &exit_data_len, exit_data,
```
ifranzki, line 2921: So I guess this will fail if it is sent to a crypto adapter that does not support the XPRTCPAC keyword, or if the host library does not support it. So this is an implicit check for the require versions.

However: What if multiple CCA adapters are in the system, and DEV-ANY is used, and not all of the adapters are on the desired level? 

Also You create an AES key with XPRTCPAC, this is actually supported with CEX6C already. But EC with XPRTCPAC needs CEX7C I guess..... So this check here will actually check for CEX6C, but not for CEX7C.

I guess you will need to add an explicit version check of all APQNs used by this token. Iterate over them using cca_iterate_adapters() and check their FW version/level. 

Only if ALL adapters are at least CEX6C, then XPRTCPAC can be used for AES keys.
Only if ALL adapters are at least CEX7C, then XPRTCPAC can be used for ECC keys.

So you might want to differentiate between AES and EC key support of protected key, or yo just enable PKEY only if all are CEX7Cs. 

You must make sure to not attempt to create a key with the XPRTCPAC  keyword, if not all adapters selectable for this token supports it.

ifranzki: I would actually also add an explicit check for the CCA holt library version to be 7.x.x. or later. You can use field `version` in struct cca_private_data, it contains version, release, modification level of the used host library already.

For AES keys, version 6.3.27 would be sufficient, but I guess we can just require 7.0.0 or later.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -2773,6 +3057,32 @@ CK_RV token_specific_init(STDLL_TokData_t * tokdata, CK_SLOT_ID SlotNumber,
     if (rc != CKR_OK)
         goto error;
 
+    cca_private->msa_level = get_msa_level();
```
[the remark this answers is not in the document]
ifranzki: fd 0 is stdin, so if stdin is closed during `C_Initialize`, it is quite clear that `pkcsconf -I` fails to prompt for the pin and this leads to the travis failures.

usr/lib/cca_stdll/cca_specific.c
jschmidb: Done.

usr/lib/cca_stdll/cca_specific.c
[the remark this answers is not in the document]
jschmidb: Ah thanks! Makes sense. I'm adding it right after cca_private is assigned:
```
    tokdata->private_data = cca_private;
    cca_private->pkeyfd = -1;
```

There is a 'goto error' above, before even allocating cca_private, but this should be ok.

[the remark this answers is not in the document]
jschmidb: OK, will do ...

usr/lib/cca_stdll/cca_specific.c
jschmidb: Yes, I didn't put reading the pkey_mode option out of the config file into a #ifndef NO_PKEY block, so the option is read in any case. But as all functions that evaluate 'cca_data->pkey_mode' are enclosed in #ifndef NO_PKEY it doesn't matter what the value is.  But I can set the value to DISABLED in cca_config_set_pkey_mode when NO_PKEY is defined. 

config file description is added.

ifranzki: And please add code to not accept anything else than DISABLED if NO_PKEY is defined. We don't want that user think they have it enabled, but it is ignored.

usr/lib/cca_stdll/cca_specific.c
ifranzki: Please also check the adapter versions as mentioned in my first comment on this place.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -2974,6 +2974,304 @@ static CK_BBOOL ccatok_pkey_option_disabled(STDLL_TokData_t *tokdata)
...
+ *            path is ok.
+ *
+ *         all others:
+ *            An internal error occurred and it was possibly attempted to create
+ *            a protected key for the object. In this case, the key blob, or
+ *            any other attr, might be no longer valid in a parallel environment
+ *            and the ep11 fallback is not possible anymore. The calling
+ *            function shall return with an error in this case.
+ */
+CK_RV ccatok_pkey_check(STDLL_TokData_t *tokdata, SESSION *session,
```
ifranzki, line 3148: I guess this function can also be static. You are not using it anywhere outside of cca_specific.c.

testcases/crypto/ec_func.c
```diff
@@ -392,7 +392,7 @@ CK_RV run_DeriveECDHKey(void)
     testcase_user_login();
 
     /* Skip tests if pkey = true, but the slot doesn't support protected keys*/
-    if (pkey && !is_ep11_token(SLOT_ID)) {
+    if (pkey && !is_ep11_token(SLOT_ID) && !is_cca_token(SLOT_ID)) {
```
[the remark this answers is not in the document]
jschmidb: Yep, makes sense. Will update  create_ECPrivateKey() and generate_EC_KeyPair() ... 

Will also check ci_config.sh ...

usr/lib/cca_stdll/cca_specific.c
jschmidb: > Please also check the adapter versions as mentioned in my first comment on this place

See new commit: add check for minimum card version

ifranzki:
In commit description of commit "CCA pkey option: new config option PKEY_MODE":
....only supported on s390x (Linux on Z) -> Linux on **IBM** Z

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -2690,6 +2698,285 @@ static CK_RV ccatok_pkey_add_attrs(STDLL_TokData_t * tokdata, TEMPLATE *template
...
+    long key_name_2_length = 0;
+    long user_data_1_length = 0;
+    long user_data_2_length = 0;
+    long kek_identifier_1_length = 0;
+    long kek_identifier_2_length = 0;
+    long the_2nd_key_token_length = 0;
+
+    /* Check CCA host library version: XPRTCPAC requires min 7.0.0 */
+    if (cca_data->version.ver < 7) {
+        TRACE_ERROR("CCA host lib is %d.%d.%d, but pkey support requires min 7.0.0\n",
```
ifranzki, line 2924: Maybe also issue a SYSLOG message here? 
If it fails because of that, then below code issues syslog message "Warning: Could not get mk_vp, protected key support not available.", but it does not show a reason. So if this message is issued to SYSLOG then the reason would be know. It might not be too seldom that uses run with a CCA host library earlier than version 7.....

ifranzki:
Commit: "CCA pkey option: fix in token_specific_init"
This is not specific to PKEY, so better name it just "CCA: fix in token_specific_init".

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -2922,6 +2927,14 @@ static CK_RV ccatok_pkey_get_firmware_wkvp(STDLL_TokData_t *tokdata)
         return CKR_FUNCTION_NOT_SUPPORTED;
     }
 
+    /* Check minimum card version, XPRTCPAC requires min CEX7C */
+    if (cca_data->min_card_version.ver < 7) {
+        TRACE_ERROR("Minimum card version must be CEX7C, but we only have %d.%d.%d\n",
+                    cca_data->version.ver,cca_data->version.rel,
+                    cca_data->version.mod);
+        return CKR_FUNCTION_NOT_SUPPORTED;
```
ifranzki, line 2935: Syslog message also here?

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -312,6 +312,11 @@ struct cca_private_data {
         unsigned int rel;
         unsigned int mod;
     } version;
+    struct {
+        unsigned int ver;
+        unsigned int rel;
+        unsigned int mod;
+    } min_card_version;
```
ifranzki, line 319: Since we have the second same version struct again here, I would suggest to add a struct definitions just in front of the struct cca_private_data like this:

```
struct cca_version {
        unsigned int ver;	
        unsigned int rel;	
        unsigned int mod;
    };	  
``` 

and use the following inside cca_private_data  (and everywhere else where the version struct is used:
```
struct cca_version cca_lib_version;
struct cca_version min_card_version;
```

While at it, please rename the 'version' field to 'cca_lib_version' or similar, indicating who's version it holds.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -3310,6 +3323,117 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
 }
 #endif /* NO_PKEY */
 
+typedef struct {
+    CK_BBOOL card_level_set;
+    struct {
+        unsigned int ver;
+        unsigned int rel;
+        unsigned int mod;
+    } min_card_version;
```
ifranzki, line 3391: use struct cca_version here as well

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -3310,6 +3323,117 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
...
+} cca_min_card_version_t;
+
+#define CCA_STATCCA_CCA_VERSION_OFFSET         24
+#define CCA_STATCCA_CCA_VERSION_LENGTH          8
+#define MAX_INT                                0xFFFFFFFF
```
ifranzki, line 3396: Move these defines to cca_stdll.h near the other offset defines

And use UINT_MAX from limits.h

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -3310,6 +3323,117 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
...
+/*
+ * Called from within cca_iterate_adapters() handler function, thus no need to
+ * obtain  CCA adapter lock
+ */
+static CK_RV cca_get_adapter_version(cca_min_card_version_t *data)
+{
+    unsigned char rule_array[CCA_RULE_ARRAY_SIZE] = { 0, };
+    long return_code, reason_code, rule_array_count, verb_data_length;
+    unsigned int ver = MAX_INT, rel = MAX_INT, mod = MAX_INT;
+    char ccaversion[8];
```
ifranzki, line 3407: char ccaversion[CCA_STATCCA_CCA_VERSION_LENGTH + 1];

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -3310,6 +3323,117 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
...
+                &rule_array_count, rule_array,
+                &verb_data_length, NULL);
+
+    if (return_code != CCA_SUCCESS) {
+        TRACE_ERROR("CSUACFQ (STATCCA) failed. return:%ld, reason:%ld\n",
+                    return_code, reason_code);
+        return CKR_FUNCTION_FAILED;
+    }
+
+    memcpy(ccaversion, &rule_array[CCA_STATCCA_CCA_VERSION_OFFSET],
```
ifranzki, line 3423: This leaves ccaversion non-zero-terminated ! 
Make sure to add `ccaversion[CCA_STATCCA_CCA_VERSION_LENGTH] = '\0';`

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -3310,6 +3323,117 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
...
+           CCA_STATCCA_CCA_VERSION_LENGTH);
+
+    sscanf(ccaversion, "%d.%d.%02d*", (int *)&ver, (int *)&rel, (int *)&mod);
```
ifranzki, line 3426: error checking ?

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -3310,6 +3323,117 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
...
+
+    if (ver < data->min_card_version.ver) {
+        data->min_card_version.ver = ver;
+        data->min_card_version.rel = rel;
+        data->min_card_version.mod = mod;
+        data->card_level_set = 1;
+    } else if (rel < data->min_card_version.rel) {
```
ifranzki, line 3433: What if ver is > data->min_card_version.ver ? Then you mess up the rel and mod....

`} else if (ver == data->min_card_version.ver && rel < data->min_card_version.rel) {`

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -3310,6 +3323,117 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
...
+        data->min_card_version.rel = rel;
+        data->min_card_version.mod = mod;
+        data->card_level_set = 1;
+    } else if (mod < data->min_card_version.mod) {
```
ifranzki, line 3437: `} else if (ver == data->min_card_version.ver && rel == data->min_card_version.rel && mod < data->min_card_version.mod) {`

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -3310,6 +3323,117 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
...
+                               &card_level_data);
+
+    if (ret != CKR_OK || card_level_data.card_level_set == 0) {
+        TRACE_ERROR("cca_iterate_adapters failed, could not determine min card level.\n");
+        ret = CKR_FUNCTION_FAILED;
+        goto done;
+    }
+
+    /* Update min card level in cca_private_data */
+    memcpy(&cca_private->min_card_version, &card_level_data.min_card_version,
```
ifranzki, line 3486: If you use the `struct cca_version` as mentioned above you can do:
`cca_private->min_card_version = card_level_data.min_card_version;`

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -3389,6 +3513,10 @@ CK_RV token_specific_init(STDLL_TokData_t * tokdata, CK_SLOT_ID SlotNumber,
     if (rc != CKR_OK)
         goto error;
 
+    rc = cca_get_min_card_level(tokdata);
```
ifranzki, line 3575: Call this in cca_handle_apqn_event() as well. 

But then it needs proper locking:

- Add `pthread_rwlock_t min_card_version_rwlock;` to `struct cca_private_data`
- Initialize the lock in `token_specific_init()`: `pthread_rwlock_init(&cca_private->min_card_version_rwlock, NULL);` (plus error checking)
- Free lock in token_specific_final(): `pthread_rwlock_destroy(&cca_private->min_card_version_rwlock);`
- Write-Lock in cca_get_min_card_level() where you copy the version from `card_level_data` to `cca_private->min_card_version`: `pthread_rwlock_wrlock(&cca_private->min_card_version_rwlock);` plus error checking and `pthread_rwlock_unlock(&cca_private->min_card_version_rwlock);` after the copy.  Do not hld the lock while you iterate over the APQNs, just while you are copying the version from `card_level_data` to `cca_private->min_card_version`.
- Read lock in `ccatok_pkey_get_firmware_wkvp() `where you chekc the min-card_version: `pthread_rwlock_rdlock(&cca_private->min_card_version_rwlock);` plus error checking and `pthread_rwlock_unlock(&cca_private->min_card_version_rwlock);`. I.e. its this line that needs the lock:  `if (cca_data->min_card_version.ver < 7) {`
- in `cca_handle_apqn_event()` call `cca_get_min_card_level()` as well.
- in `cca_handle_apqn_event()` check resulting min_card_version and determine the right value of `cca_data->pkey_wrap_supported` again. I.e. if min_card_version < 7, then set pkey_wrap_supported = FALSE. If min_card_version >= 7, then check msa level, etc again and set pkey_wrap_supported = TRUE if all preconditions are meet (no need to get FW-WKVP again here).

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -3321,6 +3321,65 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
 
     return ret;
 }
+
+/*
+ * ECC private key section (X'20'), Key-usage and translation control flag.
+ */
+#define CCA_XPRTCPAC                          0x01
```
ifranzki, line 3328: Move these defines to cca_stdll.h

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -3321,6 +3321,65 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
...
+                                     TEMPLATE * templ,
+                                     enum cca_token_type token_type)
+{
+    CK_ATTRIBUTE *pkey_attr = NULL, *skey_attr = NULL;
+    CK_RV ret;
+
+    UNUSED(tokdata);
+
+    /* OK, if there's no secure key token at all */
+    ret = template_attribute_find(templ, CKA_IBM_OPAQUE, &skey_attr);
```
ifranzki, line 3351: Please use template_attribute_get_non_empty()

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -3321,6 +3321,65 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
...
+    if (ret != TRUE)
+        return CKR_OK;
+
+    /* Also ok, if there's no CKA_IBM_PROTKEY_EXTRACTABLE attribute */
+    ret = template_attribute_find(templ, CKA_IBM_PROTKEY_EXTRACTABLE, &pkey_attr);
```
ifranzki, line 3356: please use template_attribute_get_bool()

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -3321,6 +3321,65 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
...
+
+    if (keyusage & CCA_XPRTCPAC)
+        return CK_TRUE;
+
+    return CK_FALSE;
+}
+
+static CK_RV ccatok_pkey_check_attrs(STDLL_TokData_t *tokdata,
+                                     TEMPLATE * templ,
+                                     enum cca_token_type token_type)
```
ifranzki, line 3343: It does not make much sense that the function obtains the key blob internally, but expects the caller to pass the cca_token_type .... The cca_token_type can only be known by the caller if it has obtained the key blob already and passed it through analyse_cca_key_token().

Either: pass blob and cca_token_type  as parameters, or let  the function obtain everything internally.
I would vote for passing everything as paranmeters.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -3321,6 +3321,65 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
...
+     * At this point, the key has CKA_IBM_PROKEY_EXTRACTABLE = true and it has
+     * a secure key token, so let's check if the secure key token is
+     * CPACF-exportable.
+     */
+    switch (token_type) {
+    case sec_aes_data_key:
+        /* Nothing to do: AES data keys are always CPACF-exportable */
+        break;
+    case sec_ecc_priv_key:
+        if (pkey_op_ec_curve_supported_by_cpacf(templ) &&
```
ifranzki, line 3370: Do we really want to check the curve here? What if at a alter time CPACF supports more curves? 
I guess we can accept to have CKA_IBM_PROTKEY_EXTRACTABLE=TRUE in any case, except the key blob does not have XPRTCPAC. Thats the only real reason why to reject CKA_IBM_PROTKEY_EXTRACTABLE=TRUE. 

We also allow DES or RSA keys to have CKA_IBM_PROTKEY_EXTRACTABLE=TRUE, although we don't support PKEY for those key types.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -2977,6 +2977,90 @@ static CK_BBOOL ccatok_pkey_option_disabled(STDLL_TokData_t *tokdata)
...
+                 * function that it's not AES here. It covers all EC mechs
+                 */
+            break;
+        default:
+            add_pkey_extractable = CK_TRUE;
+            break;
+        }
+
+        if (add_pkey_extractable) {
+            if (!template_attribute_find(tmpl, CKA_IBM_PROTKEY_EXTRACTABLE, &pkey_attr)) {
```
ifranzki, line 3294: Use template_attribute_get_bool

usr/lib/cca_stdll/cca_specific.c
jschmidb: Done.

usr/lib/cca_stdll/cca_specific.c
jschmidb: Done.

usr/lib/cca_stdll/cca_specific.c
ifranzki: Hm, if cca_handle_apqn_event() detects that pkey is now possible, but it has never been possible before, then the FW-WKVP is not available. So you would need to to ccatok_pkey_get_firmware_wkvp() as well. 
But then the FW-WKVP field needs to be lock-protected as well against multithreaded access.
Same for the pkey_wrap_supported field actually....

usr/lib/cca_stdll/cca_specific.c
jschmidb: Done.

usr/lib/cca_stdll/cca_specific.c
jschmidb: Done.

usr/lib/cca_stdll/cca_specific.c
jschmidb: Done.

usr/lib/cca_stdll/cca_specific.c
jschmidb: Done.

usr/lib/cca_stdll/cca_specific.c
jschmidb: Done.

usr/lib/cca_stdll/cca_specific.c
jschmidb: Done.

usr/lib/cca_stdll/cca_specific.c
jschmidb: Done.

usr/lib/cca_stdll/cca_specific.c
ifranzki: See PR https://github.com/opencryptoki/opencryptoki/pull/677 for how this can be done and how pkey_wrap_supported can be set without additional locks in the APQN event routine.

usr/lib/cca_stdll/cca_specific.c
ifranzki: Well, might be ok to use template_attribute_find here, since its just about checking the existence of the attribute here.

jschmidb: Done. We may want to fix token_specific_set_attrs_for_new_object in ep11_specific.c in the same way.

ifranzki: Hm, I think ists better the way it was before.... I don't like the local variable `CK_BBOOL pkey_attr_value` within the block, and also the check `!= CKR_OK || pkey_attr_value == CK_FALSE)` seems wrong. Why checking for the variable being FALSE ?
Lets revert this change to how it was before.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -2858,8 +2882,22 @@ static CK_RV ccatok_pkey_skey2pkey(STDLL_TokData_t *tokdata,
...
-    ret = cca_iterate_adapters(tokdata, ccatok_pkey_sec2prot,
-                               &pkey_wrap_handler_data);
+    while (num_retries < 3600) {
+
+        ret = cca_iterate_adapters(tokdata, ccatok_pkey_sec2prot,
+                                   &pkey_wrap_handler_data);
+
+        if (ret != CKR_OK || !pkey_wrap_handler_data.wrap_was_successful) {
+
+            if (ccatok_pkey_token_is_encr_with_new_mk(tokdata, key_type, mkvp)) {
```
ifranzki, line 2892: Please add the call to `check_expected_mkvp()` above right after the call to `analyse_cca_key_token()` with the following error handling: 

```
    if (check_expected_mkvp(tokdata, keytype, mkvp, &new_mk) != CKR_OK) {
        TRACE_ERROR("%s\n", ock_err(ERR_DEVICE_ERROR));
        rc = CKR_DEVICE_ERROR;
        goto done;
    }
```

We do want to detect if the blob is using the right MKVP (old or new) in any case. 
You get the `new_mk` flag that way right away. You can just check for `new_mk == TRUE` inside the loop instead of calling `ccatok_pkey_mk_change_active()`

Also, instead of calling `ccatok_pkey_mk_change_active()`, please just use `cca_mk_change_find_op_by_keytype(...) != NULL`.
No need for adding any new functions here.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -2858,8 +2882,22 @@ static CK_RV ccatok_pkey_skey2pkey(STDLL_TokData_t *tokdata,
...
+                if (ccatok_pkey_mk_change_active(tokdata, key_type)) {
+                    sleep(1);
+                    num_retries++;
+                    continue;
+                }
+            }
+        }
```
ifranzki, line 2899: You are missing to break the loop if pkey_wrap_handler_data.wrap_was_successful was true. You don't want to loop if the conversion was successful.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -2609,6 +2609,29 @@ CK_RV cca_load_config_file(STDLL_TokData_t *tokdata, char *conf_name)
 }
 
 #ifndef NO_PKEY
+static CK_BBOOL ccatok_pkey_token_is_encr_with_new_mk(STDLL_TokData_t *tokdata,
```
ifranzki, line 2612: No need for these 2 functions, see other comments.

usr/lib/cca_stdll/cca_specific.c
jschmidb: experimentally done, will push again ...

usr/lib/cca_stdll/cca_specific.c
jschmidb: Done.

usr/lib/cca_stdll/cca_specific.c
jschmidb: Hm, yes, I think the idea was to pass the template, because the function also has to get the CKA_IBM_PROTKEY_EXTRACTABLE attribute internally. So we need the template inside ccatok_pkey_check_attrs anyway. Consequently, the outer routine may also provide CKA_IBM_PROTKEY_EXTRACTABLE as a parm. What do you think?

usr/lib/cca_stdll/cca_specific.c
jschmidb: Done.

usr/lib/cca_stdll/cca_specific.c
ifranzki: Well, since the caller does not really need CKA_IBM_PROTKEY_EXTRACTABLE, it might not be intuitiv to let it get that attribute and pass it as param. In contrast to that, the blob is anyway obtained by the caller, so here it is quite intuitive to pass it along.  

So I guess I would still pass the template to ccatok_pkey_check_attrs() so that it can get the CKA_IBM_PROTKEY_EXTRACTABLE itself, but pass the blob to it explicitly, since it is already available at the caller. 

Although this is still a mixture if pass and get internally, I guess this is the easiest way for the caller (you have at least 2 callers, maybe more in the future).

usr/lib/cca_stdll/cca_specific.c
jschmidb: Done.

usr/lib/cca_stdll/cca_specific.c
jschmidb: OK, done.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -3325,6 +3325,54 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
 
     return ret;
 }
+
+static CK_BBOOL ccatok_token_is_cpacf_exportable(const CK_BYTE *token)
```
ifranzki, line 3348: Please also pass the token size to this function, and check if the token is large enough for CCA_ECC_TOKEN_KEYUSAGE_OFFSET.

usr/lib/cca_stdll/cca_specific.c
ifranzki: I guess you can make this commit non-experimental now. The code looks good, so you can merge it into commit "CCA pkey option: determine firmware wkvp" or keep it as separate commit (but then with changed commit title).

usr/lib/cca_stdll/cca_specific.c
jschmidb: Done.

usr/lib/cca_stdll/cca_specific.c
jschmidb: BTW: you always created macros to handle concurrent MK change ops. Should we also introduce two macros here, like BEGIN_RETRY_SEC2PROT / END_RETRY_SEC2PROT ?

usr/lib/cca_stdll/cca_specific.c
jschmidb: Done.

usr/lib/cca_stdll/cca_specific.c
ifranzki: I created such macros for cases where the same logic is needed at many places. 
This here is needed at exact this one and only place, so no need to create a macro.

Actually RETRY_NEW_MK_BLOB_START / RETRY_NEW_MK_BLOB_END is a macro for that matter, but it does not fit here, because you are dealing with the pkey ioctl, not with CCA verbs.

usr/lib/cca_stdll/cca_specific.c
jschmidb: As discussed offline: The CCA lib allows the XPRTCPAC rule_array keyword only for the 5 curves supported by CPACF: p256, p384, p521, ed25519, ed448. Adjusted the check, but still checking the curves here. 
Done.

usr/lib/cca_stdll/cca_specific.c
jschmidb: Done.

ifranzki:
Your last change introduced build errors in Travis on non-s390x platforms:
```
  CC       usr/lib/cca_stdll/opencryptoki_stdll_libpkcs11_cca_la-cca_specific.lo
usr/lib/cca_stdll/cca_specific.c: In function ‘cca_handle_apqn_event’:
usr/lib/cca_stdll/cca_specific.c:9844:17: error: implicit declaration of function ‘ccatok_pkey_option_disabled’ [-Werror=implicit-function-declaration]
 9844 |                !ccatok_pkey_option_disabled(tokdata)) {
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~
usr/lib/cca_stdll/cca_specific.c:9846:14: error: implicit declaration of function ‘ccatok_pkey_get_firmware_wkvp’ [-Werror=implicit-function-declaration]
 9846 |         rc = ccatok_pkey_get_firmware_wkvp(tokdata);
      |              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
make[1]: *** [Makefile:7528: usr/lib/cca_stdll/opencryptoki_stdll_libpkcs11_cca_la-cca_specific.lo] Error 1
make[1]: Leaving directory '/home/travis/build/opencryptoki/opencryptoki'
make: *** [Makefile:4213: all] Error 2
```

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -9823,8 +9823,33 @@ static CK_RV cca_handle_apqn_event(STDLL_TokData_t *tokdata,
...
+    } else if (cca_private->pkey_wrap_supported == 0 &&
+               !ccatok_pkey_option_disabled(tokdata)) {
+        /* get firmware WKVP, this will enable PKEY on success */
+        rc = ccatok_pkey_get_firmware_wkvp(tokdata);
+        if (rc != CKR_OK) {
+            OCK_SYSLOG(LOG_WARNING,
+                "%s: Warning: Could not get wkvp, protected key support not available.\n",
+                __func__);
+            TRACE_WARNING("Could not get wkvp, protected key support not available.\n");
+        }
```
ifranzki, line 9852: This is a good start, but you still need proper locking when updating the min_card_version in cca_get_min_card_level():

- Add `pthread_rwlock_t min_card_version_rwlock;` to `struct cca_private_data`
- Initialize the lock in `token_specific_init()`: `pthread_rwlock_init(&cca_private->min_card_version_rwlock, NULL);` (plus error checking)
- Free lock in token_specific_final(): `pthread_rwlock_destroy(&cca_private->min_card_version_rwlock);`
- Write-Lock in cca_get_min_card_level() where you copy the version from `card_level_data` to `cca_private->min_card_version`: `pthread_rwlock_wrlock(&cca_private->min_card_version_rwlock);` plus error checking and `pthread_rwlock_unlock(&cca_private->min_card_version_rwlock);` after the copy.  Do not hold the lock while you iterate over the APQNs, just while you are copying the version from `card_level_data` to `cca_private->min_card_version`.
- Read lock in `ccatok_pkey_get_firmware_wkvp() `where you check the min-card_version: `pthread_rwlock_rdlock(&cca_private->min_card_version_rwlock);` plus error checking and `pthread_rwlock_unlock(&cca_private->min_card_version_rwlock);`. I.e. its this line that needs the lock:  `if (cca_data->min_card_version.ver < 7) {`
- Read lock in cca_handle_apqn_event() where you check the min-card_version: `pthread_rwlock_rdlock(&cca_private->min_card_version_rwlock);` plus error checking and `pthread_rwlock_unlock(&cca_private->min_card_version_rwlock);`. I.e. its this line that needs the lock:  `if (cca_data->min_card_version.ver < 7) {`
- in ccatok_pkey_get_firmware_wkvp() use __sync_or_and_fetch() to set pkey_wrap_supported to TRUE.

On your discussion points:
1. do we need to check the CCA lib version also? I think not, because
   the lib cannot be exchanged dynamically.
-> no the lib cannot change.

2. In the previous logic, the routine returned CKR_OK when cca_check_mks
   returned something != CKR_OK. I'm now returning this error rc, which
   seems more appropriate
-> hm no, better not. The event handling function did its work properly, and it flagged the token as inconsistent in that case. So no reason to return an error. Returning an error from the event handler function means that the event handling as such failed, which is not the case. So please change it back to return CKR_OK. Also, I would not skip the CCA version checking in that case either. Let it continue just as it was before.


3. RW lock around updating the min card level and the pkey_wrap_supported
   variable?
-> yes, see above. For updating pkey_wrap_supported __sync_or_and_fetch / __sync_and_and_fetch is sufficient.

usr/lib/cca_stdll/cca_specific.c
ifranzki: you still need to change this.

usr/lib/cca_stdll/cca_specific.c
ifranzki: you still need to change this.

usr/lib/cca_stdll/cca_specific.c
jschmidb: Regarding the travis errors, I guess I need #ifndef NO_PKEY around the new code block. Will re-push to check ... 
Also, changed the return above the new code block back to 'return CKR_OK'. 
Other changes will follow ...

ifranzki: Please move the #ifdef after the call to cca_get_min_card_level(), we do want to update the version info in any case, regardless if pkey is supported on the platform or not.

jschmidb: Done.

ifranzki: Looks good. Lets merge this with commit "[CCA pkey option: add check for minimum card version](https://github.com/opencryptoki/opencryptoki/pull/676/commits/5c3b4e5f64814b4afaedff1eb3102215656a2647)".

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -3327,6 +3355,135 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
...
+typedef struct {
+    CK_BBOOL card_level_set;
+    struct cca_version min_card_version;
+} cca_min_card_version_t;
+
+/* return -1 if v1 < v2, 0 if v1 == v2, 1 if v1 > v2 */
+static int compare_cca_version(const struct cca_version *v1,
+                               const struct cca_version *v2)
+{
+
```
ifranzki, line 3444: remove empty line (yes I know it was there also in EP11's version of this function).

testcases/crypto/ec_func.c
```diff
@@ -1081,6 +1082,7 @@ CK_RV run_DeriveECDHKeyKAT(void)
             {CKA_KEY_TYPE, &secret_key_type, sizeof(secret_key_type)},
             {CKA_VALUE_LEN, &(ecdh_tv[i].derived_key_len), sizeof(CK_ULONG)},
             {CKA_SENSITIVE, &false, sizeof(false)},
+            {CKA_IBM_PROTKEY_EXTRACTABLE, &true, sizeof(true)},
```
ifranzki, line 1085: Shouldn't we make this dependent on the pkey mode here as well? Also add CKA_EXTRACTABLE to the template, otherwise this might not even have an effect for EP11.

jschmidb: Done.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -8656,10 +9865,53 @@ static CK_RV cca_handle_apqn_event(STDLL_TokData_t *tokdata,
...
+    /* Re-check after APQN set change if protected key support is available */
+    rc = cca_get_min_card_level(tokdata);
+    if (rc != CKR_OK) {
+        TRACE_WARNING("Could not re-determine min card level, protected key support not available.\n");
+        return rc;
+    }
+
+#ifndef NO_PKEY
+    {
```
ifranzki, line 9881: Please avoid this block and move the min_card_version variable declaration (incl #ifdef around) to the top of the function.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -2545,6 +2607,959 @@ CK_RV cca_load_config_file(STDLL_TokData_t *tokdata, char *conf_name)
...
+        TRACE_ERROR("CCA min_card_version RD-Unlock failed.\n");
+        return CKR_CANT_LOCK;
+    }
+
+    /* Check minimum card version, XPRTCPAC requires min CEX7C */
+    if (min_card_version < 7) {
+        TRACE_ERROR("Minimum card version must be CEX7C, but we only have %d.%d.%d\n",
+                    cca_data->min_card_version.ver,cca_data->min_card_version.rel,
+                    cca_data->min_card_version.mod);
+        OCK_SYSLOG(LOG_ERR, "Minimum card version must be CEX7C, but we only have %d.%d.%d\n",
```
ifranzki, line 2982: Please make this a LOG_WARN, this is really not an error situation. It just means we can't use PKEY.

usr/lib/cca_stdll/cca_specific.c
jschmidb: Yes, this was my initial approach, but when compiled with NO_PKEY we get an "unused variable" warning. So another #ifndef is necessary around the variable .. but will fix.

usr/lib/cca_stdll/cca_specific.c
jschmidb: p.s. will also change the TRACE_ERROR to TRACE_WARNING then

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -2545,6 +2607,959 @@ CK_RV cca_load_config_file(STDLL_TokData_t *tokdata, char *conf_name)
...
+    long kek_identifier_2_length = 0;
+    long the_2nd_key_token_length = 0;
+    unsigned int min_card_version;
+
+    /* Check CCA host library version: XPRTCPAC requires min 7.0.0 */
+    if (cca_data->cca_lib_version.ver < 7) {
+        TRACE_ERROR("CCA host lib is %d.%d.%d, but pkey support requires min 7.0.0\n",
+                    cca_data->cca_lib_version.ver,cca_data->cca_lib_version.rel,
+                    cca_data->cca_lib_version.mod);
+        OCK_SYSLOG(LOG_ERR, "CCA host lib is %d.%d.%d, but pkey support requires min 7.0.0\n",
```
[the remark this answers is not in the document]
jschmidb: Done, syslog and trace.

usr/lib/cca_stdll/cca_specific.c
jschmidb: Done.

PR #682, opened 2023-04-17
pkcsslotd hardening

ifranzki:


usr/sbin/pkcsslotd/mutex.c
```diff
@@ -27,6 +27,9 @@ int CreateXProcLock(void)
     struct group *grp;
     mode_t mode = (S_IRUSR | S_IRGRP);
 
+    if (xplfd == -1)
+        xplfd = open(OCK_API_LOCK_FILE, O_RDONLY);
```
jschmidb, line 31: is the mode parm intentionally left out here?

usr/sbin/pkcsslotd/slotmgr.c
```diff
@@ -113,28 +117,80 @@ int compute_hash(int hash_type, int buf_size, char *buf, char *digest)
...
-        {LOCKDIR_PATH, S_IRWXU | S_IRWXG},
-        {OCK_LOGDIR, S_IRWXU | S_IRWXG},
-        {CONFIG_PATH, S_IRWXU | S_IRWXG},
-        {CONFIG_PATH "/HSM_MK_CHANGE", S_IRWXU | S_IRWXG},
-        {NULL, 0},
+        { "/run/opencryptoki", S_IRWXU | S_IXGRP, 1 },
+        { LOCKDIR_PATH, S_IRWXU | S_IRWXG, 0 },
+        { OCK_LOGDIR, S_IRWXU | S_IRWXG, 0 },
+        { CONFIG_PATH, S_IRWXU | S_IRWXG, 0 },
+        { CONFIG_PATH "/HSM_MK_CHANGE", S_IRWXU | S_IRWXG, 0 },
```
jschmidb, line 225: As this folder name (/HSM_MK_CHANGE) is global in opencryptoki, we could make a constant for it. Or even a constant in a include file that's included by pkcsslotd and opencryptoki itself.

usr/sbin/pkcsslotd/slotmgr.c
```diff
@@ -791,6 +828,9 @@ int main(int argc, char *argv[], char *envp[])
             return EACCES;
     }
 
+    if (drop_capabilities() != 0)
+        return 7;
```
jschmidb, line 836: what means 7? Is there probably an error constant?

usr/sbin/pkcsslotd/mutex.c
ifranzki: Yes, the mode parameter only has a meaning if O_CREAT (or O_TMPFILE ) is specified (i.e. the file is to be created), otherwise it is ignored. So no need to specify it here, since it is using O_RDONLY to try to open an existing file.

Makefile.am
```diff
@@ -162,7 +162,7 @@ endif
 	@echo "Remember you must run ldconfig before using the above settings"
 	@echo "--------------------------------------------------------------"
 	$(MKDIR_P) $(DESTDIR)$(lockdir) $(DESTDIR)$(logdir)
-	$(CHGRP) pkcs11 $(DESTDIR)$(lockdir) $(DESTDIR)$(logdir)
+	$(CHGRP) $(pkcs_group) $(DESTDIR)$(lockdir) $(DESTDIR)$(logdir)
```
jschmidb, line 165: Below in line 175 and 176 is pkcs11 hard coded. Should it also be $(pkcs_group)?

usr/sbin/pkcsslotd/slotmgr.c
ifranzki: Well, the exit codes of pkcsslotd are a mess.... other places also return arbitrary numbers to somehow indicate the place where it happend.... I will leave it as is for now, but that's something I need to clean up generally at a later time.

Makefile.am
ifranzki: In lines 175 and 176 it is the name of a directory that must always be named 'pkcs11', and has nothing to do with the group name.

usr/sbin/pkcsslotd/slotmgr.c
ifranzki: Yea, I should have done that already when introducing the HSM MK change support.
I have added a new commit before the others to change this.

vkanjur (approved):

jschmidb (approved):

PR #683, opened 2023-04-18
Support CKM_PBE_SHA1_DES3_EDE_CBC for EP11

ifranzki:
The EP11 token claims to support CKM_PBE_SHA1_DES3_EDE_CBC, so lets make it really work and add a testcase for it.

### DIFF
--- a/testcases/crypto/des3_func.c
+++ b/testcases/crypto/des3_func.c
@@ -1328,6 +1328,95 @@ testcase_cleanup:
     return rc;
 }
 
+CK_RV do_PBE(void)
+{
+    CK_SESSION_HANDLE session;
+    CK_OBJECT_HANDLE h_key = CK_INVALID_HANDLE;
+    CK_FLAGS flags;
+    CK_RV rc = CKR_OK;
+    CK_SLOT_ID slot_id = SLOT_ID;
+    CK_ULONG user_pin_len;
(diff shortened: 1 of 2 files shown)


testcases/crypto/des3_func.c
```diff
@@ -1328,6 +1328,95 @@ CK_RV do_SignVerifyMAC(struct published_mac_test_suite_info *tsuite)
...
+    clear_len = sizeof(clear);
+    memset(encrypted, 0x00, sizeof(encrypted));
+    encrypted_len = sizeof(encrypted);
+
+    rc = funcs->C_Encrypt(session, clear, clear_len, encrypted, &encrypted_len);
+    if (rc != CKR_OK) {
+        testcase_error("C_Encrypt rc=%s", p11_get_ckr(rc));
+        goto error;
+    }
+
```
jschmidb, line 1403: Should we probably also check if decrypt returns the original plaintext? (C_DecryptInit + C_Decrypt)

ifranzki: Well, this encrypt is to check if the created key is 'usable' at all. I would say that performing just encryption is sufficient for verifying this.

jschmidb: ok, so this is why you also only test the DES3 mech. Let's assume that the DES2 mech works as well.

jschmidb (approved):

PR #685, opened 2023-04-19
Remove support for --disable-locks (transactional memory)

ifranzki:
The use of transactional memory has never really worked for openCryptoki. Also, many code places use locks, even with '--disable-locks'. Option '--enable-locks' has been the default so far.

To retain backward compatibility, continue to accept the '--disable-locks' option during configuration, but ignore it if it is specified. A warning message is issued if --disable-locks' is still used.

With this change, there is no longer a need to maintain 2 copies of the btree.c and sess_mgr.c source (btree.c and lock_btree.c, sess_mgr.c and lock_sess_mgr.c), which contained the same code except the differences for supporting transactional memory versus locks. Now only the version with locks is retained, but it is renamed to not contain the 'lock_' prefix any more.

**Note to reviewer:** You do see changes in files btree.c and sess_mgr.c in this commit. This are no real changes but just because I first removed btree.c and sess_mgr.c and then renamed lock_btree.c and lock_sess_mgr.c to btree.c and sess_mgr.c. So it appears as if I changed them, but I did not touch any line of those files, just renamed them. So no need to review the changes in these 2 files.


.travis.yml
```diff
@@ -11,44 +11,44 @@ before_install:
...
           os: linux
           compiler: clang
-          env: CONFIG_OPTS="--enable-swtok --enable-icsftok --enable-ccatok --enable-testcases --enable-locks --with-systemd" CFLAGS="-O3 -Werror -DDEBUG"
+          env: CONFIG_OPTS="--enable-swtok --enable-icsftok --enable-ccatok --enable-testcases --with-systemd" CFLAGS="-O3 -Werror -DDEBUG"
         - name: "linux-x86-gcc-tm"
           os: linux
           compiler: gcc
-          env: CONFIG_OPTS="--enable-swtok --enable-icsftok --enable-ccatok --enable-testcases --disable-locks --with-systemd" CFLAGS="-O3 -Wno-clobbered -Werror"
-        - name: "linux-ppc64le-clang-locks"
+          env: CONFIG_OPTS="--enable-swtok --enable-icsftok --enable-ccatok --enable-testcases --with-systemd" CFLAGS="-O3 -Wno-clobbered -Werror"
```
jschmidb, line 21: As we know that tm is no more available, we could remove the "xxx-tm" jobs from Travis. However, there is one difference between the previous "-locks" job compared to -tm:
-Werror versus -Wno-clobbered -Werror
Are you intentionally leave the -tm job because of this?

configure.ac
```diff
@@ -183,7 +181,7 @@ AC_ARG_ENABLE([pkcsep11_session],
 
 dnl --- locking support
 AC_ARG_ENABLE([locks],
-	AS_HELP_STRING([--disable-locks],[build opencryptoki with transactional memory instead of locks]))
+	AS_HELP_STRING([--disable-locks],[This option is no longer supported, if it is specified, it is ignored.]))
```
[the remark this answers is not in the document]
ifranzki: will use your version.

.travis.yml
ifranzki: well, I guess I should remove the '-tm' from the names of the jobs. But I want to keep these jobs in general, because they are compiling with gcc versus the others use clang. I would like to cover both compilers as they sometimes behave differently in regards to warnings.

vkanjur (approved):

PR #687, opened 2023-05-02
Updates for making RSA operations constant time

ifranzki:


usr/lib/common/mech_openssl.c
```diff
@@ -37,6 +37,101 @@
 #include <openssl/param_build.h>
 #endif
 
+void openssl_free_ex_data(OBJECT *obj, void *ex_data, size_t ex_data_len)
+{
+    struct openssl_ex_data *data = ex_data;
+
+    if (ex_data == NULL || ex_data_len < sizeof(struct openssl_ex_data))
```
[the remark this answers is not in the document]
ifranzki: Actually yes. These mech_openssl routines are used by 2 tokens: 1. the soft token and 2. the ica token as software fallback, when libica does not provide a mechanism or fails to perform it. 
The soft token attaches just `struct openssl_ex_data` to an object, but the ICA token attaches `ica_ex_data_t` which contains `struct openssl_ex_data` as first member. The ICA token must be prepared to allow the mech_openssl routines to perform the software fallback.
To free such ICA attached data, callback function `ica_free_ex_data()` is used, which calls `openssl_free_ex_data()` as very last thing. So here, the data length is actually larger than just `struct openssl_ex_data`. Thus the check for  `ex_data_len < sizeof(struct openssl_ex_data)` to allow larger data.

usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -2413,132 +2509,116 @@ static CK_RV ica_specific_rsa_decrypt(STDLL_TokData_t *tokdata,
...
+                                                           priv_exp);
+            if (ex_data->modexpoKey == NULL) {
+                TRACE_ERROR("rsa_convert_mod_expo_key failed\n");
+                rc = CKR_FUNCTION_FAILED;
+                goto done;
+            }
+            /* in_data must be in big endian format. Size in bits must not
+             * exceed the bit length of the key, and size in bytes must
+             * be the same */
+            // FIXME: we're not cheking the size in bits of in_data
```
vkanjur, line 2961: Do we need to remove the FIXME comment? Or is it intended to be left in the code?

usr/lib/common/mech_openssl.c
```diff
@@ -37,6 +37,101 @@
...
+        obj->ex_data_free(obj, ex_data, ex_data_len);
+    return CKR_OK;
+}
+
+/*
+ * Gets the attached OpenSSL ex_data from the key object. If no ex_data is
+ * attached yet, then a WRITE lock is obtained, the ex_data is allocated in the
+ * specified size, and the ex_data is returned. If the ex_data is already
+ * attached, the need_wr_lock routine is called to determine if a WRTE lock is
+ * needed. If the need_wr_lock routine return TRUE, a WRITE lock is obtained,
```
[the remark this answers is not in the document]
ifranzki: will fix

usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -484,6 +510,9 @@ static CK_BBOOL ica_need_wr_lock_rsa_privkey(OBJECT *obj, void *ex_data,
     if (ex_data == NULL || ex_data_len < sizeof(ica_private_data_t))
         return FALSE;
 
+    if (data->blinding == NULL || data->blinding_mont_ctx == NULL)
+        return TRUE;
```
vkanjur, line 514: Do we need to return TRUE? Or is it FALSE like above?

usr/lib/ica_s390_stdll/ica_specific.c
ifranzki: Yea, I just moved that comment with the other lines into the block.... 
But, yes, lets remove the FIXME here. 
There are other similar FIXMEs in the code, I leave them as is since I am not touching those lines.

usr/lib/ica_s390_stdll/ica_specific.c
ifranzki: We need to return TRUE, because when blinding or blinding_mont_ctx  is NULL, then we need to setup blinding first, and this requires a write lock, since we are changing the attached data (i.e. set the blinding and blinding_mont_ctx fields). 

Similar when modexpoKey and crtKey are both NULL, then we also need a write lock because we are updating one of them.

usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -2410,6 +2439,362 @@ CK_RV token_specific_rsa_generate_keypair(STDLL_TokData_t *tokdata,
...
+     * - OpenSSL 1.1.1u:
+     *   https://github.com/openssl/openssl/commit/3f499b24f3bcd66db022074f7e8b4f6ee266a3ae
+     * - OpenSSL 3.0.9:
+     *   https://github.com/openssl/openssl/commit/a00d757d9ca212994625d1a02c81cc5edd27e13b
+     * - OpenSSl 3.1.1:
+     *   https://github.com/openssl/openssl/commit/550a16247e899363ef973aa08623f9b19bb636fb
+     */
+    rc = BN_BLINDING_invert_ex(bn_data, unblind, ex_data->blinding, bn_ctx);
+    if (rc != 1) {
+        TRACE_ERROR("BN_BLINDING_innvert_ex failed\n");
```
vkanjur, line 2757: Typo BN_BLINDING_innvert_ex -> BN_BLINDING_invert_ex

ifranzki: will fix

usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -2588,11 +2976,42 @@ static CK_RV ica_specific_rsa_decrypt(STDLL_TokData_t *tokdata,
...
+    if (bn_ctx == NULL) {
+        TRACE_ERROR("BN_CTX_new failed\n");
+        rc = CKR_FUNCTION_FAILED;
+        goto done;
+    }
+
+    if (ex_data->blinding == NULL || ex_data->blinding_mont_ctx == NULL) {
+        rc = ica_blinding_setup(tokdata, key_obj, ex_data, bn_ctx);
+        if (rc != CKR_OK) {
+            TRACE_ERROR("ica_setup_blinding failed\n");
```
vkanjur, line 2989: Typo ica_setup_blinding -> ica_blinding_setup

ifranzki: you have good eyes ;-) 
will fix

usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -2616,9 +3035,24 @@ static CK_RV ica_specific_rsa_decrypt(STDLL_TokData_t *tokdata,
         break;
     }
 
+    rc = ica_blinding_invert(tokdata, key_obj, ex_data, bn_ctx,
+                             buff + in_data_len, out_data, in_data_len,
+                             unblind);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("ica_blinding_convert\n");
```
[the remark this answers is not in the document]
ifranzki: will fix

vkanjur:
Do we need testcases update for const time?

ifranzki:
> Do we need testcases update for const time?

Function wise it should (must?) be transparent, so the existing testcases check that it still works as before.
However, I am thinking of adding some test using dudect to test for constant-timeness at a later time. However, those run pretty long, not sure if this is doable in the CI.....

jschmidb (approved):

vkanjur (approved):

vkanjur:
Also we found some minor Typos in commit msgs.
1)common: Allow to attach external data to an object
This data only lives in memory, and is not saved to the disk when the object
stored (i.e. for token objects). -> is stored (i.e. for token objects).

2)soft/ica: Cache OpenSSL PKEY object as data attached to key object
alyways -> always

PR #695, opened 2023-05-25
Refactoring ep11_specific.c

ifranzki:
Move code for EP11 session handling and concurrent MK change into separate source files. 

There should be no logic change, just moving code around. Some definitions need to be moved to the header file, and some variables can no longer be static.


jschmidb (approved):
lgtm

vkanjur (approved):

PR #700, opened 2023-08-02
CCA: Don't remember serialno of adapter as it can change during runtime

ifranzki:
The CCA token queries the serial number of the current CCA adapter during token initialization. This serial number is then later used to select the adapter during cca_iterate_adapters().

Unfortunately, the underlying adapter may change during runtime, and thus the serialno may change as well. Query the current serialno freshly within cca_iterate_adapters() and not depend on a remembered, and now possibly incorrect serialno.

### DIFF
--- a/usr/lib/cca_stdll/cca_specific.c
+++ b/usr/lib/cca_stdll/cca_specific.c
@@ -330,7 +330,6 @@ struct cca_private_data {
     unsigned int num_usagedoms;
     unsigned short usage_domains[256];
     CK_BBOOL inconsistent;
-    char serialno[CCA_SERIALNO_LENGTH + 1];
     struct cca_mk_change_op mk_change_ops[CCA_NUM_MK_TYPES];
     char token_config_filename[PATH_MAX];
     int pkey_mode;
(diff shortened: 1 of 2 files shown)


jschmidb:
Vishnu and I are just discussing this pr. One thought is: can we get this problem also in icatoken? And, as icatoken uses libica, which can use accelerators and CCA coprocessors, it may happen that an APQN comes online again even with a different type (CEXnA -> CEXnC and vice versa). Also to be considered: accelerators don't have a serialnr property in the sysfs.

ifranzki:
>  One thought is: can we get this problem also in icatoken?

Not directly, as libica does not care about the serial number of the adapters. 

The CCA token only cares about the serial number because of CCA's 'strange' way of addressing the adapters (CRPxx vs APQNs, i.e. card+domain), and the serial number is the only vehicle to match the APQNs known by Linux/sysfs with the cards as known by the CCA host library (CRPxx). 

However, what libica (and maybe other applications using crypt cards as well) must always be prepared for is that adapters and domains can come and go during runtime. This also includes that an adapter changes its type, i.e. the old adapter goes, and a new one with another type comes at the same address. 

However, as long as you don't explicitly address an APQN, but let the kernel look for an adapter that can handle the request, you should be fine. Libca does not address an adapter explicitly, it just passes a request to the kernel and lets it look for an adapter that can handle it. So libica should be fine.

vkanjur (approved):

jschmidb (approved):
Approving per Vishnus test results. Thanks Vishnu!

PR #704, opened 2023-08-11
CCA: Handle concurrent MK change for AES-XTS keys special cases

ifranzki:
@vkanjur As discussed some time ago, here is the code change to handle MK changes in-between key1 and key2 of an XTS key. 
We agreed that I will add that code on top of your PR. 

---------------
When the new CCA AES master key is activated after the first part of an AES-XTS key has been generated or imported, but before the second part of the AES-XTS key is generated or imported, special handling is required.

Also, when the first part of an AES-XTS key has been generated or imported while the new CCA AES master key was already activated, but before generating or importing the second part of the AES-XTS key the APQN with the new CCA AES master key gets offline, and another APQN which still has the old CCA AES master key is used to generate or import the second part of the AES-XTS key, both key parts needs to be recreated (now both with the old CCA AES master key).

When importing an AES-XTS from an opaque key blob, then both parts must use the same master key (i.e. the keys contain the same master key verification pattern). If that is not the case, then the import fails with CKR_ATTRIBUTE_VALUE_INVALID. This check was already present in the code.

### DIFF
--- a/usr/lib/cca_stdll/cca_specific.c
+++ b/usr/lib/cca_stdll/cca_specific.c
@@ -445,9 +445,10 @@ static CK_RV cca_reencipher_sec_key(STDLL_TokData_t *tokdata,
                                     CK_ULONG sec_key_len, CK_BBOOL from_old);
 
 static CK_RV cca_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl,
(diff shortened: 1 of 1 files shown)


jschmidb (approved):
lgtm

vkanjur (approved):

PR #705, opened 2023-08-14
Ensure that the 2 key parts of an AES-XTS key are not the same

ifranzki:


usr/lib/cca_stdll/cca_specific.c
```diff
@@ -3674,6 +3674,17 @@ CK_RV token_specific_aes_xts_key_gen(STDLL_TokData_t *tokdata, TEMPLATE *tmpl,
         goto retry;
     }
 
+    /*
+     * Compare the encrypted key material to ensure that the 2 key parts are
+     * not the same.
+     * A CCA AES-DATA key blob contains the encrypted key material at
+     * offset 16, with a length of 32 bytes.
```
jschmidb, line 3681: Is this always the case? Are AES-256 XTS keys not supported by CCA?

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -2844,6 +2844,13 @@ static CK_RV import_aes_xts_key(STDLL_TokData_t *tokdata, SESSION *sess,
         return rc;
     }
 
+    if (memcmp(attr->pValue,
+               ((CK_BYTE *)attr->pValue) + attr->ulValueLen / 2,
+               attr->ulValueLen / 2) == 0) {
```
jschmidb, line 2849: Is at this point already checked that attr->ulValueLen is an even number and also is an allowed even number?

usr/lib/cca_stdll/cca_specific.c
ifranzki: Remember that CCA does not support XTS at all. Tats why we build an XTS key out of 2 single AES keys.

A single CCA AES key token always have an 32 bytes encrypted key, regardless of the key-size of the underlying AES key. Simply because the AES-master key is an AES-256 bit key, so the resulting encrypted key is always 32 bytes.

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: Yes, aes_validate_attribute() in key.c checks for that during attribute validation, which was run before.

jschmidb (approved):
ok, looks good then.

vkanjur (approved):

PR #706, opened 2023-08-15
CCA: Add missing \n to trace message

ifranzki:


ifranzki:
The travis failures for x86 are unrelated....

jschmidb (approved):
Looks good. While at it, did you check if there are other places with missing newlines?

ifranzki:
> did you check if there are other places with missing newlines?

Well, I did not see any in the trace files I looked at..... But I can't be 100% sure that I haven't missed one :-)

PR #707, opened 2023-08-15
CCA: Move CCA MK change code into separate source file

ifranzki:
No logic change, just moving existing code into a new source file cca_mkchange.c. This requires that certain definitions are moved into cca_stdll.h, and some functions can no longer be static.


usr/lib/cca_stdll/cca_stdll.h
```diff
@@ -190,4 +200,225 @@ enum cca_ktc_type {
...
+ * Macros to enclose any usage of CCA verbs.
+ * Obtains a READ lock on the CCA adapter lock, if a DOM-ANY configuration is
+ * used. This prevents CCA adapter usage concurrent to another thread performing
+ * Domain selection processing. Domain selection works on process scope, so
+ * it would influence all threads that currently use CCA verbs.
+ */
+#define USE_CCA_ADAPTER_START(tokdata, return_code, reason_code)             \
+                do {                                                         \
+                    if (((struct cca_private_data *)                         \
+                                      (tokdata)->private_data)->dom_any) {  \
```
jschmidb, line 341: while at it, you may move the ending \ one col right

ifranzki: done

PR #709, opened 2023-08-18
TESTCASES: Remove EC key objects after test step

ifranzki:
Remove the created EC keys when the test step is complete, don't leave them hanging around.


jschmidb:
Ah right, the first key pair was not removed! Good catch!

jschmidb (approved):

ifranzki:
> Ah right, the first key pair was not removed! Good catch!

actually all keys except those from the last curve type were not removed. This left 44 EC private keys in the repository (the public keys are session objects, but the private key are token objects - for whatever reason).

jschmidb:
Yes, routine generate_EC_KeyPair in testcases/common/common.c has CKA_TOKEN in its privateKeyTemplate. Perhaps because of the cca_ecc_export_import_tests.

PR #710, opened 2023-08-18
some more p11sak fixes

ifranzki:


usr/sbin/p11sak/p11sak.c
```diff
@@ -6911,16 +6911,25 @@ static CK_RV p11sak_extract_x509_pk(const struct p11sak_objtype *certtype,
...
     }
 
     /* If no new ID is specified, try to use ID from certificate */
     if (opt_new_id == NULL) {
         rc = get_attribute(cert, &id_attr);
-        if (rc == CKR_OK)
+        if (rc == CKR_OK) {
             rc = add_attribute(CKA_ID, id_attr.pValue, id_attr.ulValueLen, attrs, num_attrs);
+            if (rc != CKR_OK) {
+                warnx("Failed to add attributes for extracted certificate's public key.");
```
jschmidb, line 6926: Yep, makes sense. Just a nit: here we goto done with the rc from the failing function, while below in the old code at 6929 we explicitly set rc =  CKR_FUNCTION_FAILED before the goto done.

ifranzki: right, thats because the statements before line 6929 add up the return codes. So if several add_attribute() calls return different non-zero return codes (rather unlikely, but possible), then the resulting sum of the return code is most likely not a valid return code. So thats why we set CKR_FUNCTION_FAILED if the sum is non zero.

jschmidb (approved):

Issue #697, opened 2023-06-02
p11sak: Add support for exporting/importing Edwards/Montgomery keys (EP11 token)


ifranzki:
Currently only true EC key can be exported or imported. The EP11 token supports certain Edwards/Montgomery curves as EC key (CKK_EC). However, OpenSSL distinguishes between EC and ECX keys and needs different code to export/import those. Because of this, it currently fails if one tries to export or import an ECX key. 

Most likely, ECX keys can only be imported or exported when using OpenSSL 3.0 or later, because OpenSSL 1.1.1 does not have a way to get or set the key material of a an ECX key. OpenSSL 3.0 should allow this using the EVP_PKEY_fromdata/todata functions with ECX specific parameters.

Note: PKCS#11 v3 defines distinct key types for Edwards and Montgomery keys: CKK_EC_EDWARDS and CKK_EC_MONTGOMERY. Those are currently not supported by opencryptoki.  Supporting those will certainly also have a hit on p11sak.

PR #711, opened 2023-08-22
p11sak: Support import and export of Edwards and Montgomery keys

ifranzki:
Only since OpenSSL 3.0 there exists a public API to access the key components of Edwards and Montgomery keys. So import and export of Edwards and Montgomery keys is only supported with OpenSSL 3.0 or later.

Closes: https://github.com/opencryptoki/opencryptoki/issues/697


testcases/misc_tests/p11sak_test.sh
```diff
@@ -297,6 +297,16 @@ p11sak import-key ec private --slot $SLOT --pin $PKCS11_USER_PIN --label "import
...
 p11sak import-key ec public --slot $SLOT --pin $PKCS11_USER_PIN --label "import-ec-secp521r1-public" --file $DIR/ec-key-secp521r1.pem --attr sX
 RC_P11SAK_IMPORT=$((RC_P11SAK_IMPORT + $?))
+# edwards/montgomery
+if [[ -n $(openssl version | grep "OpenSSL 3.") && -n $( pkcsconf -t -c $SLOT | grep "Model: EP11") ]]; then
+	p11sak import-key ec private --slot $SLOT --pin $PKCS11_USER_PIN --label "import-ec-ed25519-private" --file $DIR/ed25519-private-key.pem --attr sX
+	RC_P11SAK_IMPORT=$((RC_P11SAK_IMPORT + $?))
+	p11sak import-key ec public --slot $SLOT --pin $PKCS11_USER_PIN --label "import-ec-ed25519-public" --file $DIR/ed25519-public-key.pem --attr sX
+	RC_P11SAK_IMPORT=$((RC_P11SAK_IMPORT + $?))
+else
+	echo "Skip importing edwards/montgomery keys, OpenSSL version not supporting it or not EP11 token"
```
jschmidb, line 307: Do you intentionally leave out tests for ed448? 
Also not sure if there's a difference between ed448 and x448 in the created key material.

ifranzki: Yes, I decided to test only one out of the 4 possibilities with p11sak:

- ed25519
- ed448
- x25519
- x448

It all runs through the same code in p11sak, only within OpenSSL it may make a difference. 
To test the OCK token support of those 4 curves, we already have tests in testcases_crypto/ec_tests.c.

PR #712, opened 2023-08-22
More p11sak fixes

ifranzki:
An RSA-PSS key can be generated via:
`openssl genpkey -algorithm rsa-pss -pkeyopt rsa_keygen_bits:2048  -out rsapss.pem`
Generating a certificate with this key resuts in a certificate that contains an RSA-PSS key.


jschmidb (approved):
lgtm

PR #713, opened 2023-08-28
Ensure OpenSSL config is loaded before creating own library context

ifranzki:
Ensure OpenSSL is initialized and the OpenSSL config is loaded BEFORE creating OpenCryptoki's own library context. Otherwise the OpenSSL config is loaded later, which may cause that all configured providers are also loaded into the library context. We intentionally want to use only the default and legacy provider for OpenCryptoki's software crypto operations. So loading other providers into our context may cause that those other providers are used instead. In worst case this may cause a recursion, e.g. when a PKCS#11 provider is loaded routing the crypto calls back to OpenCryptoki.

### DIFF
--- a/usr/lib/api/api_interface.c
+++ b/usr/lib/api/api_interface.c
@@ -3104,7 +3104,13 @@ CK_RV C_Initialize(CK_VOID_PTR pVoid)
      *   -> PKCS11-provider -> Opencryptoki -> ...
      * Explicitly using the 'default' provider only for Opencrypoki's OpenSSL
      * usage breaks this loop.
+     * Ensure OpenSSL is initialized and the OpenSSL config is loaded
+     * BEFORE creating the library context. Otherwise the OpenSSL config
+     * is loaded later, which may cause that all configured providers
+     * are also loaded into the library context. We need to make sure that
+     * only the default and legacy provider is loaded in the library context.
      */
+    OPENSSL_init_crypto(OPENSSL_INIT_LOAD_CONFIG, NULL);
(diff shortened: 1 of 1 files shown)


usr/lib/api/api_interface.c
```diff
@@ -3104,7 +3104,13 @@ CK_RV C_Initialize(CK_VOID_PTR pVoid)
      *   -> PKCS11-provider -> Opencryptoki -> ...
      * Explicitly using the 'default' provider only for Opencrypoki's OpenSSL
      * usage breaks this loop.
+     * Ensure OpenSSL is initialized and the OpenSSL config is loaded
+     * BEFORE creating the library context. Otherwise the OpenSSL config
+     * is loaded later, which may cause that all configured providers
+     * are also loaded into the library context. We need to make sure that
+     * only the default and legacy provider is loaded in the library context.
      */
+    OPENSSL_init_crypto(OPENSSL_INIT_LOAD_CONFIG, NULL);
```
jschmidb, line 3113: Yes, lgtm, same as for libica.
One thought here: as this fix goes into the api layer, it affects all tokens. And all tokens, except ICSF, in fact use openssl. Just let's make sure the fix doesn't introduce a problem for the ICSF token.

ifranzki: Actually the OpenCryptoki common code als uses OpenSSL itself (pin encryption, etc), so basically all tokens use OpenSSL anyway, explicitly or implicitly.

jschmidb (approved):

PR #714, opened 2023-09-11
Misc fixes

ifranzki:


usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -1550,6 +1550,16 @@ static CK_RV ab_unwrap_update_template(STDLL_TokData_t * tokdata,
         return CKR_TEMPLATE_INCONSISTENT;
     }
     for (i = 0; i < sizeof(attrs) / sizeof(CK_ATTRIBUTE); ++i) {
+        if (attrs[i].type == CKA_VALUE_LEN) {
+            switch (keytype) {
+            case CKK_GENERIC_SECRET:
```
jschmidb, line 1555: What's about CKK_DES2 and CKK_DES3? They show up in ep11_supported_mech_list()

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -8129,6 +8129,13 @@ CK_RV ep11tok_generate_key_pair(STDLL_TokData_t * tokdata, SESSION * sess,
...
     }
 
+    rc = update_ep11_attrs_from_blob(tokdata, sess, public_key_obj, FALSE);
+    if (rc != CKR_OK) {
+        TRACE_ERROR("%s update_ep11_attrs_from_blob failed with rc=0x%lx\n",
+                    __func__, rc);
+        goto error;
+    }
+
     /* Copy CKA_MODULUS and CKA_PUBLIC_EXPONENT attributes from
```
jschmidb, line 8139: ok, yes, that was missing. Not all attrs handled by update_ep11_attrs_from_blob() may be applicable for public keys.

jschmidb: ... but this is handled by checking each attr value for CK_UNAVAILABLE_INFORMATION. OK.

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: DES keys are fixed in length, and the length only depends on the key type (DES, DES2 or DES3). Thus, the PKCS#11 standard does not define the CKA_VALUE_LEN attribute for such keys.  The only key types where CKA_VALUE_LEN is defined per PKCS#11 standard is AES (i.e. 128, 192, 256) and generic secret.

jschmidb (approved):

PR #716, opened 2023-09-11
COMMON: Fix coverty findings - unused variable assignment

ifranzki:


usr/lib/common/data_obj.c
```diff
@@ -55,11 +55,6 @@ CK_RV data_object_set_default_attributes(TEMPLATE *tmpl, CK_ULONG mode)
     CK_ATTRIBUTE *objid_attr = NULL;
     CK_RV rc;
 
-    // satisfy the compiler
-    //
-    if (mode)
```
jschmidb, line 60: Interesting ... shouldn't we have something like UNUSED(mode) ?

PR #717, opened 2023-09-12
'make distcheck' fixes

ifranzki:
With these changes, a `make distcheck` works and the tarball contains all desired files.


Makefile.am
```diff
@@ -183,6 +183,7 @@ if ENABLE_CCATOK
 	if test -d $(DESTDIR)/$(libdir)/opencryptoki/stdll; then \
 		cd $(DESTDIR)/$(libdir)/opencryptoki/stdll && \
 		rm -f PKCS11_CCA.so; fi
+	rm -f $(DESTDIR)$(sysconfdir)/opencryptoki/ccatok.conf
```
jschmidb, line 186: Are we then removing /etc/opencryptoki/ccatok.conf ?
What if the user customized the conf file? I see we are already removing the ep11tok.conf file. So yes, let's make this consistent, but I'm not sure what's right ;-)

jschmidb:
A minor thing in the commit msg of the first commit:

> Also include several extra files into EXTRA_DIST to make sure they are distributed as well-.

(you may want to remove the dash at the end ... )

Makefile.am
ifranzki: Yes, we are removing all files that we installed with `make install`. 

Please note that `make uninstall` is not the same as an RPM uninstall. An RPM uninstall usually takes care about modified config files during install (don't overwrite if modified), and uninstall (keep if modified). 

Please also note that as part of `make distcheck` it not only builds the package, but also installs it via `make install` into a temporary location, and then uninstalls it via `make uninstall`. It then checks if all files there were installed got removed during uninstall. If not, then `make distcheck` fails. So this tells me that `make uninstall` must remove all files that are installed with `make install`.

ifranzki:
> A minor thing in the commit msg of the first commit:
> 
> > Also include several extra files into EXTRA_DIST to make sure they are distributed as well-.
> 
> (you may want to remove the dash at the end ... )

fixed

testcases/misc_tests/misc_tests.mk
```diff
@@ -10,6 +10,38 @@ noinst_PROGRAMS +=							\
...
+	testcases/misc_tests/p11sak_ecp384cert.crt			\
+	testcases/misc_tests/p11sak_ecp521cert.crt			\
+	testcases/misc_tests/p11sak_rsa2048cert.crt			\
+	testcases/misc_tests/p11sak_rsa4096cert.crt			\
+	testcases/misc_tests/migration.sh				\
+	testcases/misc_tests/p11sak_test.sh				\
+	testcases/misc_tests/pkcsconf_test.sh				\
+	testcases/misc_tests/spinlock_tests.sh				\
+	testcases/misc_tests/spinlock_child.sh				\
+	testcases/misc_tests/tok_obj2.c
```
jschmidb, line 43: why tok_obj2, but not tok_obj? And why the C-source at all?

ifranzki: `tok_obj2` seems to be a copy of `tok_obj`, but is not built or used anywhere, for whatever reason. Initially I added it here to get it distributed, too, but I guess it would make more sens to just remove that file. Will do tat in a separate commit.

jschmidb (approved):

PR #719, opened 2023-09-15
ICSF token fixes

ifranzki:


usr/lib/icsf_stdll/icsf.c
```diff
@@ -906,8 +932,30 @@ static int icsf_ber_put_attribute_list(BerElement * ber, CK_ATTRIBUTE * attrs,
                                        CK_ULONG attrs_len)
 {
     size_t i;
+    CK_OBJECT_CLASS class = CK_UNAVAILABLE_INFORMATION;
+
+    find_ulong_attribute(attrs, attrs_len, CKA_CLASS, &class);
```
jschmidb, line 937: it's a bit confusing that routine find_ulong_attribute() has a rc, but the rc is not checked. However, looks like the class value is set in all cases, in some just the preset value is still set after calling the routine.

usr/lib/icsf_stdll/icsf.c
```diff
@@ -876,14 +876,40 @@ static int is_numeric_attr(CK_ULONG type)
...
+    case CKA_JAVA_MIDP_SECURITY_DOMAIN:
         return 1;
     }
     return 0;
 }
 
+static CK_RV find_ulong_attribute(CK_ATTRIBUTE *attrs, CK_ULONG attrs_len,
+                                  CK_ATTRIBUTE_TYPE type, CK_ULONG *value)
+{
+    CK_ULONG i;
```
jschmidb, line 896: size_t instead of CK_ULONG?

usr/lib/icsf_stdll/icsf.c
ifranzki: Right, if the function fails, then the preset value is still set and the switch below goes into the default case.

usr/lib/icsf_stdll/icsf.c
ifranzki: Why ? `i` is used to iterate over the attributes up to `attrs_len`. `attrs_len` is also a CK_ULONG, so `i` should be the same type.

usr/lib/icsf_stdll/icsf.c
jschmidb: But then, why not make the function return void?

usr/lib/icsf_stdll/icsf.c
jschmidb: ok good

usr/lib/icsf_stdll/icsf.c
ifranzki: `find_ulong_attribute()` is pretty much a copy from `find_bbool_attribute()` in usr/lib/common/utility.c. I first wanted to put `find_ulong_attribute()` also into usr/lib/commno/utility.c, but module icsf.c where usr/lib/commno/utility.c. is used is also lined into the usr/sbin/pkcsicsf/pkcsicsf tool, and there utility.c is not included. Trying to include it lead to further problems (other missing functions, etc), so I decided to make `find_ulong_attribute`  a static function in icsf.c directly. 

We do have functions `get_ulong_attribute_by_type()` and friends already in attributes.c, but including that in the usr/sbin/pkcsicsf/pkcsicsf tool causes similar problems.

jschmidb: ok

jschmidb (approved):

PR #720, opened 2023-09-18
CCA/EP11: add info about AES-XTS to token config files

jschmidb:
For both tokens, AES-XTS is only supported via protected keys. Other operation modes, like ECB and CBC, are just accelerated via the PKEY_MODE option.

### DIFF
--- a/usr/lib/cca_stdll/ccatok.conf
+++ b/usr/lib/cca_stdll/ccatok.conf
@@ -31,6 +31,10 @@ version cca-0
 # IBM specific boolean attribute CKA_IBM_PROTKEY_EXTRACTABLE must be true to
 # make a key eligible for protected key support. 
 #
+# AES-XTS related mechanisms are only available if the PKEY_MODE option is not
+# disabled and additional hardware and firmware prerequisites are met. AES-XTS
+# is not supported via the CCA coprocessor itself.
+#
 #    PKEY_MODE = DISABLED | DEFAULT | ENABLED
 #
 #        DISABLED       : Protected key support disabled. All keys are used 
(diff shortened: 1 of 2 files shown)


usr/lib/cca_stdll/ccatok.conf
```diff
@@ -31,6 +31,10 @@ version cca-0
 # IBM specific boolean attribute CKA_IBM_PROTKEY_EXTRACTABLE must be true to
 # make a key eligible for protected key support. 
 #
+# AES-XTS related mechanisms are only available if the PKEY_MODE option is 
+# enabled and additional hardware and firmware prerequisites are met. AES-XTS
```
ifranzki, line 35: should we use upper case `ENABLED` here, as the option value is `ENABLED` ?

usr/lib/ep11_stdll/ep11tok.conf
```diff
@@ -81,6 +81,10 @@
 # and CKA_EXTRACTABLE cannot both be true. The default value of
 # CKA_IBM_PROTKEY_EXTRACTABLE is false.
 #
+# AES-XTS related mechanisms are only available if the PKEY_MODE option is 
+# enabled and additional hardware and firmware prerequisites are met. AES-XTS
```
ifranzki, line 85: maybe change 'is enabled' to 'is not DISABLED' ? The other 2 values mean enabled, right ?

usr/lib/cca_stdll/ccatok.conf
jschmidb: Hmm, good question ... I intentionally used "enabled" here to make it general. So we can use the same text also for ep11 where we have different forms of "enabled". OK, we may use "not disabled" as you suggest below for ep11.

ifranzki: Yes using 'not disabled' for both is probably the best, and is still general.

jschmidb: ok, text changed.

ifranzki (approved):

PR #722, opened 2023-09-18
TESTCASES: cca_export_import_test did leave RSA and EC keys over

jschmidb:
A total of 17 keys were left over in the CCA token repository when running the cca_export_import_test.

### DIFF
--- a/testcases/misc_tests/cca_export_import_test.c
+++ b/testcases/misc_tests/cca_export_import_test.c
@@ -359,7 +359,7 @@ static CK_RV cca_des_data_export_import_tests(void)
     CK_BYTE key[] = { 0x01, 0x23, 0x45, 0x67, 0x89, 0xab, 0xcd, 0xef };
     CK_BYTE iv[] = { 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08 };
     CK_MECHANISM mech = { CKM_DES_CBC, iv, sizeof(iv) };
-    CK_OBJECT_HANDLE hkey, hikey;
+    CK_OBJECT_HANDLE hkey = CK_INVALID_HANDLE, hikey = CK_INVALID_HANDLE;
     CK_BYTE data[80], encdata1[80], encdata2[80];
     CK_ULONG ilen, len;
     CK_BYTE *opaquekey = NULL;
@@ -456,6 +456,10 @@ static CK_RV cca_des_data_export_import_tests(void)
 testcase_cleanup:
     testcase_close_session();
     free(opaquekey);
+    if (hkey != CK_INVALID_HANDLE)
+        funcs->C_DestroyObject(session, hkey);
+    if (hikey != CK_INVALID_HANDLE)
+        funcs->C_DestroyObject(session, hikey);
 out:
     return rc;
 }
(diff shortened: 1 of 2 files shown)


testcases/misc_tests/cca_export_import_test.c
```diff
@@ -1259,6 +1259,11 @@ static CK_RV cca_rsa_export_import_tests(void)
         priv_opaquekey = NULL;
         free(publ_opaquekey);
         publ_opaquekey = NULL;
+
+        if (publ_key != CK_INVALID_HANDLE)
+            funcs->C_DestroyObject(session, publ_key);
+        if (priv_key != CK_INVALID_HANDLE)
+            funcs->C_DestroyObject(session, priv_key);
```
ifranzki, line 1266: Shouldn't also the `imp_priv_key` and `imp_publ_key` keys that are created in line 1155 and 1177 be destroyed here? Those are session objects, so you won't see them in the repository afterwards, but when we are at it, lets destroy them, too.

ifranzki:
How about the other functions that import keys, like `cca_des_data_export_import_tests()`, `cca_des3_data_export_import_tests()`, `cca_aes_data_export_import_tests()`, `cca_aes_xts_data_export_import_tests()`, `cca_hmac_data_export_import_tests()`? They also leak key objects....

testcases/misc_tests/cca_export_import_test.c
jschmidb: Yes, ok, will do ...

jschmidb:
Good question, running the tests does not leave any token objs in the repo after destroying the RSA and EC keys. I'll check ...

ifranzki:
> Good question, running the tests does not leave any token objs in the repo after destroying the RSA and EC keys. I'll check ...

Those are session objects, so they will go away when the session closes. But still, lets destroy the objects.

testcases/misc_tests/cca_export_import_test.c
ifranzki: You still miss to destroy imp_priv_key and imp_publ_key in cca_rsa_export_import_tests(). You are destroying it in cca_ecc_export_import_tests(), but still not in cca_rsa_export_import_tests().

jschmidb: Yep, corrected.

testcases/misc_tests/cca_export_import_test.c
```diff
@@ -1259,6 +1280,15 @@ static CK_RV cca_rsa_export_import_tests(void)
         priv_opaquekey = NULL;
         free(publ_opaquekey);
         publ_opaquekey = NULL;
+
+        if (publ_key != CK_INVALID_HANDLE)
+            funcs->C_DestroyObject(session, publ_key);
```
ifranzki, line 1285: Hm, minor issue for those cases that run inside a loop.

If the first iteration works fine, and thus destroyed the objects, then the handles are still left at at a now invalid handle value. If the next iteration fails and goes to the error label early, then it tries to destroy already destroyed objects.  Should not really hurt, C_DestroyObject will silently fail ....

jschmidb: Hmm, yes that's right ... similar to setting buffers to NULL after a free to avoid double-free's, we could add a xxx_key = CK_INVALID_HANDLE after the C_DestroyObject ...

jschmidb: Done

ifranzki (approved):

PR #724, opened 2023-10-09
p11sak: Correct help texts for attributes

ifranzki:
Include attributes CKA_IBM_PROTKEY_EXTRACTABLE ('K') and CKA_IBM_PROTKEY_NEVER_EXTRACTABLE ('Z') into the list of attributes that can be used for filtering keys.

However, only CKA_IBM_PROTKEY_EXTRACTABLE ('K') can be used for setting attributes.

CKA_TRUSTED can be used for setting attributes, but it can only be set to TRUE by the SO user.


jschmidb:
lgtm. Just to make sure we don't forget anything: the manpage already contains CKA_IBM_PROTKEY_EXTRACTABLE and CKA_IBM_PROTKEY_NEVER_EXTRACTABLE in the list of attributes. So no change necessary there?

ifranzki:
> So no change necessary there?

I would say no change needed in the man page. The man page does not list the possible attributes per command, only globally as part of the `--attr` option description.

jschmidb (approved):

PR #725, opened 2023-10-23
tmpfiles.conf: Do not create token specific dirs inside /var/lock/opencryptoki/

ifranzki:
Do not let tmpfiles.d create the token specific directories inside /var/lock/opencryptoki/. It is not known during build time which tokens will exist later during run time. A user may configure any number of tokens with any names in /etc/opencryptoki/opencryptoki.conf.

The API layer will create any non-existing token specific lock directories inside /var/lock/opencryptoki/ during C_Initialize() as needed.

### DIFF
--- a/misc/misc.mk
+++ b/misc/misc.mk
@@ -1,25 +1,3 @@
-TOKENS = swtok
-
-if ENABLE_ICATOK
-TOKENS += lite
-endif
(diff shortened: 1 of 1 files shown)


jschmidb (approved):
lgtm

PR #726, opened 2023-10-25
EP11: Also check for deconfigured and check-stopped cards

ifranzki:
When checking if a card or an APQN is online, not only check the 'online' sysfs attribute, but also check the 'config' and 'chkstop' attribute. Cards and APQNs in check-stopped or deconfigured state can still be reported as online via the sysfs attribute, although they are not available to be used for the EP11 token.

In case the 2 additional sysfs attributes are not available in sysfs, then don't fail, but rely on the 'online' attribute only.

### DIFF
--- a/usr/lib/ep11_stdll/ep11_specific.c
+++ b/usr/lib/ep11_stdll/ep11_specific.c
@@ -12535,6 +12535,18 @@ CK_BBOOL is_apqn_online(uint_32 card, uint_32 domain)
     if (strcmp(buf, "1") != 0)
         return CK_FALSE;
 
+    sprintf(fname, "%s/card%02x/%02x.%04x/config", SYSFS_DEVICES_AP,
+            card, card, domain);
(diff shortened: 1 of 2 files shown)


jschmidb:
lgtm so far. But I didn't find a similar logic in the CCA token. Why is there no 'online' check for CCA? 
As this commit is just for EP11, I would state this early in the commit msg, like:
"When checking if **an EP11** card or APQN is online ... "

ifranzki:
> lgtm so far. But I didn't find a similar logic in the CCA token. Why is there no 'online' check for CCA?

In the CCA token this is done implicitly by calling CCA verbs and looking at the return codes. For CCA it is a bit tricky to map the APQN as defined by card and domain to the CCA view of an adapter which is denoted by "CRPxx". Therefore things work different in the CCA token.

> "When checking if an EP11 card or APQN is online ... "

Done.

PR #728, opened 2023-11-10
TPM: Fix issue caused by uninitialized variables

dex:
Set NULL to OpenSSL BIGNUM pointers as initial value. Otherwise it will crash in OpenSSL libary when invoking openssl_get_modulus_and_prime function.


ifranzki (approved):
I guess it is only needed to set them to NULL for the OpenSSL 3.0 case with EVP_PKEY_get_bn_param(), which crashes if the target BIGNUM pointer contains garbage. 

For the OpenSSL 1.1.1 case RSA_get0_key/RSA_get0_factors just return internal pointers.
But it certainly does not hurt to also set them to NULL in the OpenSSL 1.1.1 case.

dex:
Thanks for the reviewing. Yes, as you mentioned it won't crash against OpenSSL 1.1.1.

ifranzki:
Thanks for you contribution.

Issue #731, opened 2023-12-11
opencryptoki vulnerable to the Marvin Attack


kkaarreell:
I've verified that opencryptoki is vulnerable to the [Marvin Attack](https://people.redhat.com/~hkario/marvin/)—a timing variant of the well-known Bleichenbacher attack.

I've executed the test on RHEL-9, using opencryptoki opencryptoki-3.21.0, on a ICA token and openssl-3.0.7-18.el9_2 . I will attach the reproducer shortly. The testing and analysis has been executed using the [marvin-toolkit](https://github.com/tomato42/marvin-toolkit/tree/master/example/crypto%2B%2B).

The test summary:
```
tlsfuzzer analyse.py version 5 analysis
Sign test mean p-value: 0.1115, median p-value: 2.313e-15, min p-value: 2.419e-286
Friedman test (chisquare approximation) for all samples
p-value: 0.0
Worst pair: 0(header_only), 5(valid_0)
Mean of differences: 1.47574e+05s, 95% CI: -1.57720e+06s, 2.066035e+06s (±1.822e+06s)
Median of differences: -3.76000e-07s, 95% CI: -3.98000e-07s, -3.570000e-07s (±2.050e-08s)
Trimmed mean (5%) of differences: -5.02983e-07s, 95% CI: -2.09351e-06s, 9.483105e-07s (±1.521e-06s)
Trimmed mean (25%) of differences: -4.14615e-07s, 95% CI: -9.93803e-07s, 1.096255e-07s (±5.517e-07s)
Trimmed mean (45%) of differences: -3.74689e-07s, 95% CI: -3.96070e-07s, -3.536652e-07s (±2.120e-08s)
Trimean of differences: -6.80000e-07s, 95% CI: -4.63625e-06s, 2.836875e-06s (±3.737e-06s)
Layperson explanation: Definite side-channel detected, implementation is VULNERABLE
For detailed report see rsa2048_repeat/report.csv
```
According to Hubert Kario:
> I can confirm that the layperson explanation is correct: there
> is a detectable side-channel. Not only is the correctness of the
> plaintext is leaking, the length of the message is leaking too.
> 
> That being said, the fact that no_header_with_payload_48 and
> signature_padding_8 is similar to the valid_0 and valid_48 suggests
> that there may be a bug in the depadding code...

I am attaching an archive with the report and will kindly ask [tomato42](https://github.com/tomato42) to provide more detailed explanation.

kkaarreell:
[...]

PR #732, opened 2023-12-12
Fixes for Marvin attack

ifranzki:
Take the implementation of OpenSSL function RSA_padding_check_PKCS1_type_2() in crypto/rsa/rsa_pk1.c instead of ossl_rsa_padding_check_PKCS1_type_2(), since the latter leaks the message size.

Fixes: https://github.com/opencryptoki/opencryptoki/issues/731


jschmidb (approved):
lgtm

PR #733, opened 2023-12-15
TESTCASE: Correct test run counting for digest test

ifranzki:
When running the digest tests against the EP11 token, the number of passed tests is greater than the number of ran tests:

  `Total=849, Ran=365, Passed=389, Failed=0, Skipped=484, Errors=0`

Fix this by adding a call to testcase_new_assertion() at the right place to ensure the ran counter is increased before a test is counted as being passed or failed.

This is only seen with the EP11 token because only the EP11 token currently supports mechanisms CKM_SHAnnn_KEY_DERIVATION. The tests for those mechanisms are missing the testcase_new_assertion() to increase the ran counter.

### DIFF
--- a/testcases/crypto/digest_func.c
+++ b/testcases/crypto/digest_func.c
@@ -1715,10 +1715,12 @@ CK_RV do_SHA_derive_key(void)
                     continue;
                 }
 
+                testcase_new_assertion();
                 testcase_fail("Derive key with %s keytype=0x%lx, value_len=%lu: rc: %s",
                               mech_to_str(derive_mechs[i].mechanism),
                               key_type, key_len, p11_get_ckr(rc));
             } else {
+                testcase_new_assertion();
                 testcase_pass("Derive key with %s keytype=0x%lx, value_len=%lu",
                               mech_to_str(derive_mechs[i].mechanism),
                               key_type, key_len);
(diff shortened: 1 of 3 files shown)


chrisRund:
Realised the dilithium_tests located in the crypto/ folder (dilithium_func.c) run into a similar issue.

ifranzki:
> Realised the dilithium_tests located in the crypto/ folder (dilithium_func.c) run into a similar issue.

Not for me:
`Total=208, Ran=208, Passed=208, Failed=0, Skipped=0, Errors=0`

Can you tell which CEX levels you have active?

chrisRund:
> 
> Can you tell which CEX levels you have active?

I am using opencryptoki 3.21 on the EP11 token which is backed by both CEX7P and CEX8P cards.
There is no change in counting when explicitly running with either CEX7P or CEX8P cards.

`Total=128, Ran=96, Passed=64, Failed=0, Skipped=32, Errors=0 `

ifranzki:
OK found it. Its because when you have a CEX7 in the set of adapters, then only some Dilithium variants are supported, and the others are skipped. The counting is wrong when skips happen.

chrisRund (approved):
Tested on the opencrytoki 3.21 version code and LGTM

PR #734, opened 2024-01-08
ICA: Enable external AES-GCM IV when libica is in FIPS mode

ifranzki:
When the system is in FIPS mode, newer libica versions may prevent AES-GCM from being used with an external IV. FIPS requires that the AES-GCM IV is created libica internally via an approved random source.

The ICA token can not support the internal generation of the AES-GCM IV, because the PKCS#11 mechanism CKM_AES_GCM does not allow this.

Enable the use of external AES-GCM IVs for libica, if the used libica library supports this. Newer libica versions support to allow external AES-GCM IVs via function ica_allow_external_gcm_iv_in_fips_mode().


jschmidb (approved):
lgtm

PR #735, opened 2024-01-09
TESTCASE: Correct test run counting for the events test case

ifranzki:
When running the events test the number of passed tests is 10, but the 'total' and 'ran' tests is 1.

  Total=1, Ran=1, Passed=10, Failed=0, Skipped=0, Errors=0

Fix this by adding a call to testcase_new_assertion() at the right places to ensure the ran counter is increased before a test is counted as being passed or failed.

### DIFF
--- a/testcases/misc_tests/events.c
+++ b/testcases/misc_tests/events.c
@@ -75,6 +75,8 @@ int main(int argc, char **argv)
     }
     testcase_pass("send_event (simple, one-shot)");
 
+    testcase_new_assertion();
+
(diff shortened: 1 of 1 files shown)


jschmidb (approved):
lgtm. 
Funny, I thought after all these years we already caught all these places ;-)

ifranzki:
> Funny, I thought after all these years we already caught all these places ;-)

I am sure we still have missed some :-)

chrisRund (approved):
LGTM

PR #736, opened 2024-01-16
Fixes for protecting buffers containing sensitive data

ifranzki:
For OpenSSL BIGNUMs containing sensitive private key material, use BN_secure_new() to allocate the BIGNUM on OpenSSLs secure heap.

Also clean any buffers that might contain sensitive data before freeing it.

### DIFF
--- a/usr/sbin/p11sak/p11sak.c
+++ b/usr/sbin/p11sak/p11sak.c
@@ -3943,8 +3943,10 @@ static void free_attributes(CK_ATTRIBUTE *attrs, CK_ULONG num_attrs)
         return;
 
     for (i = 0; i < num_attrs; i++) {
-        if (attrs[i].pValue != NULL)
+        if (attrs[i].pValue != NULL) {
+            OPENSSL_cleanse(attrs[i].pValue, attrs[i].ulValueLen);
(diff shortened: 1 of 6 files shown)


usr/sbin/p11sak/p11sak.c
```diff
@@ -9235,7 +9325,10 @@ static CK_RV p11sak_export_sym_key(const struct p11sak_objtype *keytype,
     }
 
 done:
-    free(data);
+    if (data != NULL) {
+        OPENSSL_cleanse(data, data_len);
+        free(data);
```
jschmidb, line 9330: Could you probably use OPENSSL_clear_free instead of the two calls?

ifranzki: No. the `data` has been allocated with plan `malloc`, not with `OPENSSL_malloc`, so it must also be freed with plain `free`, and NOT with `OPENSSL_[clear_]free`. You can not mix those!

jschmidb: ok, looks good then, will approve.

jschmidb (approved):

PR #739, opened 2024-01-26
TESTCASES: ec_tests: Support ECDH KATs for secure key tokens

ifranzki:
For secure key tokens, the derived secret can not be compared with the expected key value for known answer tests (KATs), because the derived secret is a secure key that never reveals its key value in clear.

However, the test can calculate an HMAC of some know data using the derived key as well as with the expected key value from the KAT. If both keys are the same, then the calculated HMAC must also be the same.

### DIFF
--- a/testcases/crypto/ec_func.c
+++ b/testcases/crypto/ec_func.c
@@ -15,6 +15,8 @@
 #include <string.h>
 #include <memory.h>
 
+#include <openssl/hmac.h>
+
 #include "pkcs11types.h"
 #include "regress.h"
 #include "common.c"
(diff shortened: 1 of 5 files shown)


testcases/crypto/ec_func.c
```diff
@@ -362,6 +364,50 @@ CK_RV run_HMACSign(CK_SESSION_HANDLE session, CK_OBJECT_HANDLE h_key,
...
+    default:
+        testcase_notice("Mechanism %s is not supported. Skipping key check",
+                        p11_get_ckm(&mechtable_funcs, hmac_mech));
+        *mac_len = 0;
+        return CKR_OK;
+    }
+
+    out_len = *mac_len;
+    if (HMAC(md, key, key_len, data, sizeof(data), mac, &out_len) == NULL) {
+        testcase_notice("HMAC (OpneSSL) failed");
```
jschmidb, line 403: OpenSSL

testcases/crypto/ec_func.c
```diff
@@ -1207,51 +1296,104 @@ CK_RV run_DeriveECDHKeyKAT(void)
...
+
+            rc = run_HMACSign(session, secret_keyA, ecdh_tv[i].derived_key_len,
+                              CKM_SHA_1_HMAC, mac1, &mac1_len);
+            if (rc != CKR_OK) {
+                testcase_fail("HMAC for derived key #1 failed: %s",
+                              p11_get_ckr(rc));
+                goto testcase_cleanup;
+            }
+
+            if (mac1_len != 0 && /* skip chck if mac can't be calcualted */
```
jschmidb, line 1374: skip check if mac can't be calculated

jschmidb: ... also below

testcases/crypto/ec_func.c
ifranzki: fixed

testcases/crypto/ec_func.c
ifranzki: fixed both

jschmidb (approved):
lgtm

ifranzki:
Added another commit to support compressed or hybrid EC points with EP11 as public data with ECDH. This also removes the ugly checking in the testcases for such EC points.
@jschmidb please take one more look.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -6282,6 +6282,8 @@ CK_RV ep11tok_derive_key(STDLL_TokData_t *tokdata, SESSION *session,
          * The EP11 host library only accepts the raw form, thus convert
          * it to the raw format if the caller specified it in the DER-encoded
          * form.
+         * Also, the EP11 host library only accepts EC points is uncompressed
```
jschmidb, line 6285: in compressed form

usr/lib/common/mech_ec.c
```diff
@@ -1530,3 +1530,73 @@ int ec_point_from_public_data(const CK_BYTE *data, CK_ULONG data_len,
...
+        break;
+    }
+
+    if (rc != CKR_OK && *allocated) {
+        free(*ec_point);
+        *ec_point = NULL;
+        *ec_point_len = 0;
+        *allocated = FALSE;
+    }
+    if (ec_point2 != NULL)
```
jschmidb, line 1598: Hmm, in line 1581 *ec_point = ec_point2;
If something goes wrong, i.e. rc != CKR_OK, *ec_point gets freed, which is ok.
But also in the good case, in line 1599, ec_point2 is freed, which means *ec_point is freed. Is this intended?

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: fixed

usr/lib/common/mech_ec.c
ifranzki: > But also in the good case, in line 1599, ec_point2 is freed

No it is not. ec_point2 is set to NULL in line 1584, so it won't be freed.

jschmidb (approved):

PR #741, opened 2024-01-29
common: Fix compile error/warning when compiled with GCC 14

ifranzki:
usr/lib/common/mech_openssl.c:2425:51:
   error: passing argument 5 of 'EVP_PKEY_get_octet_string_param' from
          incompatible pointer type

Change ecpoint_len to type size_t instead of CK_ULONG to match the expected type used by EVP_PKEY_get_octet_string_param(). Also, for the further use of that variable it is OK to use size_t.

### DIFF
--- a/testcases/crypto/dilithium_func.c
+++ b/testcases/crypto/dilithium_func.c
@@ -104,7 +104,7 @@ CK_RV run_SignVerifyDilithium(CK_SESSION_HANDLE session,
         }
     }
 
-    data = calloc(sizeof(CK_BYTE), inputlen);
+    data = calloc(inputlen, sizeof(CK_BYTE));
     if (data == NULL) {
         testcase_error("Can't allocate memory for %lu bytes",
                        sizeof(CK_BYTE) * inputlen);
@@ -128,7 +128,7 @@ CK_RV run_SignVerifyDilithium(CK_SESSION_HANDLE session,
         testcase_error("C_Sign rc=%s", p11_get_ckr(rc));
         goto testcase_cleanup;
     }
-    signature = calloc(sizeof(CK_BYTE), signaturelen);
+    signature = calloc(signaturelen, sizeof(CK_BYTE));
     if (signature == NULL) {
         testcase_error("Can't allocate memory for %lu bytes",
                        sizeof(CK_BYTE) * signaturelen);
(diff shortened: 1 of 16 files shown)


ifranzki:
@sharkcz OK, fixed several more places. 
I have added all this to the one commit, so please check again if now all places are fixed.
Unfortunately I can't reproduce myself, because `size_t` seems to be an `unsigned long int` on my systems, while in your case its a `unsigned int`.  Are you compiling on a 32 bit system?

sharkcz:
yes, technically it is a 32-bit system, it's a 32-bit chroot/container on top of a x86_64 host

I will give it another try and let you know.

ifranzki:
Just FYI: At some point in time we will remove support for 32 bit systems, also see https://github.com/opencryptoki/opencryptoki/issues/174

sharkcz:
> Just FYI: At some point in time we will remove support for 32 bit systems, also see #174

ack, it is already encouraged in Fedora to remove leaf package for i686

on the other hand the warnings from a 32-bit system allow to write better/safer code

sharkcz:
now it fails at `icsf_stdll`, please see https://kojipkgs.fedoraproject.org//work/tasks/6083/112536083/build.log

ifranzki:
> now it fails at `icsf_stdll`, please see https://kojipkgs.fedoraproject.org//work/tasks/8858/112548858/build.log

Hmm, are you sure you applied the complete commit?  All the places that still fail have been fixed in the commit, e.g. see https://github.com/opencryptoki/opencryptoki/pull/741/commits/94d5ff15acaf8e85103825a4252ddcff3281b3f1#diff-d474f6728105facd0bb6500ab896ed1e6e8681e540e92a211b7a49565ef5d21c

ifranzki:
Ah right, forgot to update the header.... Fixed that and some more findings shown in your log.
Give it another try please with the updated commit.

sharkcz:
The last version looks good, no errors, just 2 formatting warnings, please see https://kojipkgs.fedoraproject.org//work/tasks/7336/112607336/build.log

ifranzki:
Fixed those 2 warnings as well.

sharkcz:
Awesome, thanks a lot, not a single warning on any platform, see https://koji.fedoraproject.org/koji/taskinfo?taskID=112610096

sharkcz (approved):
LGTM

PR #744, opened 2024-02-06
opencryptoki (master branch) cannot be build on s390x

ngothan:
opencryptoki (master branch) cannot be build on s390x, it aborts with the following error:

usr/lib/ep11_stdll/ep11_specific.c: In function 'ep11tok_decrypt_single': usr/lib/ep11_stdll/ep11_specific.c:9750:27: error: 'ctx' undeclared (first use in this function)
 9750 |     if (!is_rsa_mechanism(ctx->mech.mechanism)) {
      |                           ^~~
usr/lib/ep11_stdll/ep11_specific.c:9750:27: note: each undeclared identifier is reported only once for each function it appears in
make[1]: *** [Makefile:7719: usr/lib/ep11_stdll/opencryptoki_stdll_libpkcs11_ep11_la-ep11_specific.lo] Error 1

The proposed patch resolves this issue.

### DIFF
--- a/usr/lib/ep11_stdll/ep11_specific.c
+++ b/usr/lib/ep11_stdll/ep11_specific.c
@@ -9977,7 +9977,7 @@ CK_RV ep11tok_decrypt_final(STDLL_TokData_t * tokdata, SESSION * session,
     rc = constant_time_select(constant_time_eq(rc, CKR_OK),
                               ep11_error_to_pkcs11_error(rc, session),
                               rc);
-    if (!is_rsa_mechanism(ctx->mech.mechanism)) {
+    if (!is_rsa_mechanism((&session->decr_ctx)->mech.mechanism)) {
         if (rc != CKR_OK) {
             TRACE_ERROR("%s rc=0x%lx\n", __func__, rc);
         } else {


ifranzki (changes requested):
Hm, I don't really understand what's going on here. 

First of all it does build on my system. It also builds in the Travis CI on s390x (see https://app.travis-ci.com/github/opencryptoki/opencryptoki/jobs/617379451).

Furthermore, the compile error mentions function `ep11tok_decrypt_single`, but your fix is in function `ep11tok_decrypt_final` ???

And lastly, I do not see any reason why `ctx` would be undeclared. `ctx` is declared above and is used multiple times just a few lines above just fine.

The line numbers also do not seem to match the current master.... Line 9750 is a totally different line on master.

Can you please double check and maybe ensure that you really use the code from master?

ngothan:
I checked again and realized that I did not use source from master branch, but used opencryptoki-3.22.0.tar.gz with 3 additional patches (for Marvin Attack)

 * https://github.com/opencryptoki/opencryptoki/commit/2ea019ee2b09f15724d808382d53baca03403288
 * https://github.com/opencryptoki/opencryptoki/commit/7ffc0e135b4d923d686be536aa7bf69405a360a1
 * https://github.com/opencryptoki/opencryptoki/commit/c26e049bf40d656bc51429bad190b82fbf63f0c7

The source with 3 additional patches cannot be built on s390x, but source from master branch is built without any problems.
I seems the third patch caused this issue which was fixed in master branch.

I apologize for that!

ifranzki:
Ok, good to know that you were able to sort it out.
I am closing this PR then.

PR #747, opened 2024-02-09
COMMON: Fix implicit rejection with RSA keys with empty CKA_PRIVATE_EXPONENT

ifranzki:
An RSA key object that has no CKA_PRIVATE_EXPONENT may either don't have that attribute at all, or may have an empty CKA_PRIVATE_EXPONENT attribute. Both situations should be handed the same, and the private exponent of the key needs to be calculated from the other key components.

Note that RSA key objects generated with a current soft or ICA token will always have a valid CKA_PRIVATE_EXPONENT attribute, since this is provided during key generation.


jschmidb (approved):
lgtm

PR #748, opened 2024-02-12
ICA: Fix length check for ICA ex-data

ifranzki:
The data attached to an object of the ICA token is of size sizeof(ica_ex_data_t), not sizeof(ica_private_data_t).


jschmidb (approved):
Ah right ... at line 462 the length check is correct, but at these 3 places not. 
Lgtm.

PR #750, opened 2024-03-06
EP11 pkey option: misc fixes and enhancements 

jschmidb:
This pull request adds the following fixes and enhancements:

- Add two new parameter values ENABLE4EXTR and ENABLE4ALL to the ep11token PKEY_MODE config option. Older ep11 card firmware enforces the restriction that keys can not have CKA_EXTRACTABLE=true and CKA_IBM_PROTKEY_EXTRACTABLE=true at the same time. With newer card firmware this restriction is removed and a new control point is introduced to allow checking for this feature.
- On 32-bit s390 platforms, the pkey related assembler code parts won't compile. Therefore, add NO_PKEY compile switches where necessary. The NO_PKEY compile switch is already handled in configure.ac.
- The XCP_MFL_PROBE flag is no longer needed for host libs > 4.0, so add a check for this and don't specify the flag for newer libraries.

### DIFF
--- a/usr/lib/ep11_stdll/ep11_specific.c
+++ b/usr/lib/ep11_stdll/ep11_specific.c
@@ -1080,6 +1080,26 @@ static CK_BBOOL ep11tok_pkey_session_ok_for_obj(SESSION *session,
     return CK_TRUE;
 }
 
+/*
+ * Returns true if the given key object is eligible to get a protected key
+ * attribute, false otherwise.
+ */
+CK_BBOOL ep11tok_pkey_obj_eligible_for_pkey_support(ep11_private_data_t *ep11_data,
+                                                    OBJECT *key_obj)
+{
+    if (object_is_attr_bound(key_obj) || !ep11_data->pkey_wrap_supported ||
+        !object_is_pkey_extractable(key_obj)) {
+        return CK_FALSE;
+    }
+
+    if (!ep11_data->pkey_combined_extract_supported &&
+        object_is_extractable(key_obj)) {
+        return CK_FALSE;
+    }
+
+    return CK_TRUE;
+}
+
 /**
  * Checks if the preconditions for using the related protected key of
  * the given secure key object are met. The caller of this routine must
@@ -1135,6 +1155,8 @@ CK_RV ep11tok_pkey_check(STDLL_TokData_t *tokdata, SESSION *session,
         break;
     case PKEY_MODE_DEFAULT:
     case PKEY_MODE_ENABLE4NONEXTR:
+    case PKEY_MODE_ENABLE4EXTR:
+    case PKEY_MODE_ENABLE4ALL:
         /* Use existing pkeys, re-create invalid pkeys, and also create new
          * pkeys for secret/private keys that do not already have one. EC
          * public keys that are pkey-extractable, can always be used via CPACF
@@ -1149,12 +1171,8 @@ CK_RV ep11tok_pkey_check(STDLL_TokData_t *tokdata, SESSION *session,
         if (ep11tok_pkey_get_firmware_mk_vp(tokdata, session) != CKR_OK)
             goto done;
 
-        if (object_is_extractable(key_obj) ||
-            !object_is_pkey_extractable(key_obj) ||
-            object_is_attr_bound(key_obj) ||
-            !ep11_data->pkey_wrap_supported) {
+        if (!ep11tok_pkey_obj_eligible_for_pkey_support(ep11_data, key_obj))
             goto done;
-        }
 
         if (template_attribute_get_non_empty(key_obj->template,
                                              CKA_IBM_OPAQUE_PKEY,
--- a/usr/lib/ep11_stdll/ep11_specific.c
+++ b/usr/lib/ep11_stdll/ep11_specific.c
@@ -1369,11 +1369,6 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
                 add_pkey_extractable = CK_TRUE;
             break;
         }
-        if (add_pkey_extractable) {
-            ret = ep11tok_pkey_add_protkey_attr_to_tmpl(tmpl);
-            if (ret != CKR_OK)
-                goto done;
-        }
         break;
     case PKEY_MODE_ENABLE4EXTR:
         /* If the application did not specify CKA_IBM_PROTKEY_EXTRACTABLE in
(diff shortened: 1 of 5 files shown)


usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -2838,6 +2855,7 @@ CK_RV ep11tok_init(STDLL_TokData_t * tokdata, CK_SLOT_ID SlotNumber,
             goto error;
     }
 
+#ifndef NO_PKEY
```
ifranzki, line 2865: This also excludes the pthread_mutex_init for raw2key_wrap_blob_mutex which is still needed, even without PKEY.
So I guess there needs to be an #endif after the MSA lines, and a new #ifndef before the pthread_mutex_init for pkey_mutex.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -535,13 +535,20 @@ static CK_RV ep11tok_pkey_wrap_handler(uint_32 adapter, uint_32 domain,
...
 
     if (data->wrap_was_successful)
         goto done;
 
     ep11_data = data->tokdata->private_data;
 
-    ret = get_ep11_target_for_apqn(adapter, domain, &target, XCP_MFL_PROBE);
+    if (compare_ck_version(&ep11_data->ep11_lib_version, &ver40) > 0)
+        flags = XCP_MFL_PROBE;
+    else
```
ifranzki, line 548: init flags to 0 at the variable declaration, then you can omit the lese case here.

ifranzki: and use use `flags |= XCP_MFL_PROBE;` in line 547

ifranzki:
Travis errors look unrelated. Seems like the ppc64 systems won't boot properly at the moment.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: OK ... but why not leave the get_msa_level() also for non pkey compiles? So we would add the #ifndef NO_PKEY just before if (!ep11tok_pkey_option_disabled(tokdata)

ifranzki: get_msa_level() calls s390_stfle() and this uses inline assembler. So I guess this also wont compile on 32 bit. Furthermore, get_msa_level() is in pkey_utils.c which is excluded for 32 bit builds.

jschmidb: ok, done.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -1431,6 +1436,18 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
 
     return ret;
 }
+#else
```
ifranzki, line 1444: The ifdef of this else excludes a bit too much. Function `token_specific_set_attrs_for_new_object()` must stay there, it not handles PKEY stuff, but also other stuff (e.g. FORCE_SENSITIVE handling). You should ifdef out only the PKEY related stuff from `token_specific_set_attrs_for_new_object()` but keep the rest.
Also make sure that you don't exclude function `token_specific_set_attrs_for_new_object` in `tok_struct.h`.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -3179,7 +3199,11 @@ static CK_RV import_aes_xts_key(STDLL_TokData_t *tokdata, SESSION *sess,
     if (rc != CKR_OK)
         goto import_aes_xts_key_end;
 
+#ifndef NO_PKEY
     rc = ep11tok_pkey_check_aes_xts(tokdata, aes_xts_key_obj, CKM_AES_XTS);
+#else
+    rc = CKR_FUNCTION_NOT_SUPPORTED;
```
ifranzki, line 3210: Lets add a TRACE statement here in the else case that XTS is disabled due to no PKEY support.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -4821,7 +4849,11 @@ CK_RV ep11tok_generate_key(STDLL_TokData_t * tokdata, SESSION * session,
 
     if (mech->mechanism == CKM_AES_XTS_KEY_GEN) {
         xts = TRUE;
+#ifndef NO_PKEY
         rc = ep11tok_pkey_check_aes_xts(tokdata, key_obj, mech->mechanism);
+#else
+        rc = CKR_FUNCTION_NOT_SUPPORTED;
```
ifranzki, line 4860: Trace statement here as well please.

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: hm, well, the trace in the error handling below is clear enough.... So ignore this comment please.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -1314,23 +1342,62 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
                 add_pkey_extractable = CK_TRUE;
             break;
         }
-
         if (add_pkey_extractable) {
```
ifranzki, line 1375: Could we move that if block to after the switch statement? It appears in pretty much all cases.
If it would be behind the switch, it would just look at add_pkey_extractable as set by the switch cases and do its work or not, dependent on add_pkey_extractable.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -1332,6 +1338,7 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
         }
     }
 
+#ifndef NO_PKEY
```
ifranzki, line 1346: We now have compile warnings when NO_PKEY is defined:
```
usr/lib/ep11_stdll/ep11_specific.c: In function ‘token_specific_set_attrs_for_new_object’:
usr/lib/ep11_stdll/ep11_specific.c:1316:14: warning: unused variable ‘add_pkey_extractable’ [-Wunused-variable]
 1316 |     CK_BBOOL add_pkey_extractable = CK_FALSE;
      |              ^~~~~~~~~~~~~~~~~~~~
usr/lib/ep11_stdll/ep11_specific.c:1315:14: warning: unused variable ‘extractable’ [-Wunused-variable]
 1315 |     CK_BBOOL extractable, sensitive, btrue = CK_TRUE;
      |              ^~~~~~~~~~~
usr/lib/ep11_stdll/ep11_specific.c:1314:19: warning: unused variable ‘ecp_attr’ [-Wunused-variable]
 1314 |     CK_ATTRIBUTE *ecp_attr = NULL, *sensitive_attr = NULL;
      |                   ^~~~~~~~
```

ifranzki (approved):
lgtm

PR #751, opened 2024-03-08
CCA: Support Koblitz curve secp256k1

ifranzki:
The Koblitz curve secp256k1 has a different curve type value (0x03) in the CCA Key value structure for an ECC key pair.

CCA supports Koblitz curve secp256k1 since version 7.2. Using that curve with an older CCA level causes the CCA verbs to fail with 8 / 6015 (An ECC curve type is invalid or its usage is inconsistent). This error is translated into CKR_CURVE_NOT_SUPPORTED by the CCA token.


ifranzki:
@akulhalli This should not cause any conflicts with your PR, but you might want to give it a try on AIX anyway. I would assume that CCA on AIX behaves the same as in IBM Z in regards of this curve.

testcases/crypto/ec_func.c
```diff
@@ -2536,7 +2531,7 @@ struct btc_test {
 };
 
 static const struct btc_test btc_tests[] = {
-    { .ec = { &secp256k1, sizeof(secp256k1), CK_FALSE, CURVE_PRIME,
+    { .ec = { &secp256k1, sizeof(secp256k1), CK_FALSE, CURVE_KOBLITZ,
```
jschmidb, line 2534: Also change the curve type in the next block below (?):
```
 { .ec = { &secp256k1, sizeof(secp256k1), CK_FALSE, CURVE_PRIME,
              CURVE256_LENGTH, "secp256k1" },
  ...
```

ifranzki: Oups, yes, of course .... Although it really does not make any difference for BTC tests.

jschmidb (approved):
lgtm

akulhalli:
@ifranzki you can go ahead and merge this PR. Some of my refactoring changes have broken the daemon on AIX right now - I don't have a current master build that's known to be working for me to test these changes.

I'll fix any conflicts/issues in a separate commit in my PR.

ifranzki:
Had to add CCA version checking for that curve. In case the CCA token uses DEV-ANY or DOM-ANY mode (not possible on AIX), it must be ensured that all configured adapters have the required version. Otherwise some cases will work, and some cases won't, dependent on to which adapter the request gets dispatched. 
@jschmidb can you please take a look again?

jschmidb:
> @jschmidb can you please take a look again?

still looks good.

PR #752, opened 2024-03-12
Pkey bugfixes

jschmidb:
Provides some bugfixes related to ep11 pkey support.

### DIFF
--- a/usr/lib/ep11_stdll/ep11_specific.c
+++ b/usr/lib/ep11_stdll/ep11_specific.c
@@ -12007,7 +12007,8 @@ CK_RV ep11tok_is_mechanism_supported(STDLL_TokData_t *tokdata,
 
     case CKM_AES_XTS:
     case CKM_AES_XTS_KEY_GEN:
-        if (ep11tok_pkey_option_disabled(tokdata) || ep11_data->msa_level < 4 ||
+        if ((ep11_data->pkey_wrap_support_checked && !ep11_data->pkey_wrap_supported) ||
+            ep11tok_pkey_option_disabled(tokdata) || ep11_data->msa_level < 4 ||
             ep11tok_is_mechanism_supported(tokdata, CKM_IBM_CPACF_WRAP) != CKR_OK ||
             ep11tok_is_mechanism_supported(tokdata, CKM_AES_KEY_GEN) != CKR_OK) {
             TRACE_INFO("%s Mech '%s' not suppported\n", __func__,
--- a/usr/lib/ep11_stdll/ep11_specific.c
+++ b/usr/lib/ep11_stdll/ep11_specific.c
@@ -10664,7 +10664,7 @@ static CK_RV ep11_ende_crypt_init(STDLL_TokData_t * tokdata, SESSION * session,
     case CKR_FUNCTION_NOT_SUPPORTED:
         /* if mechanism is AES XTS, return error else fallback to ep11 path */
         if (mech->mechanism == CKM_AES_XTS) {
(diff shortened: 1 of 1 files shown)


usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -12007,7 +12007,8 @@ CK_RV ep11tok_is_mechanism_supported(STDLL_TokData_t *tokdata,
 
     case CKM_AES_XTS:
     case CKM_AES_XTS_KEY_GEN:
-        if (ep11tok_pkey_option_disabled(tokdata) || ep11_data->msa_level < 4 ||
+        if (!ep11_data->pkey_wrap_supported ||
```
ifranzki, line 12011: I guess my suggestion to add this check here was a bit too quick.....

`pkey_wrap_supported` is set in `ep11tok_pkey_get_firmware_mk_vp()` which is normally called during token initialization. However, when FIPS-session mode is active, it can't be called during token initialization because it needs a logged in session. `ep11tok_pkey_get_firmware_mk_vp()` needs to generate a temporary secure key, but in FIPS-session mode every key must be session bound. And this is only possible when a logged in session is available.  Therefore, in FIPS-session mode,  `ep11tok_pkey_get_firmware_mk_vp()` is called lazy, i.e. at the places where a protected key might be needed. At these places a session is available.

So this means, that when `ep11tok_is_mechanism_supported()` is called, the `ep11tok_pkey_get_firmware_mk_vp()` might not always have been called already, and thus `pkey_wrap_supported` was not set yet (and thus defaults to FALSE). So it would then never report the XTS mechs as supported.

One idea would be to call `ep11tok_pkey_get_firmware_mk_vp()` within `ep11tok_is_mechanism_supported()`, right before the check, but that does not work either, because `ep11tok_is_mechanism_supported()` does not have a session. 

So I guess the best is to change the check to the following:

`if ((ep11_data->pkey_wrap_support_checked && !ep11_data->pkey_wrap_supported) || ....`

So if `ep11tok_pkey_get_firmware_mk_vp()` was run before, then `ep11_data->pkey_wrap_support_checked` is set, and we can evaluate `pkey_wrap_supported`, otherwise, we simply don't know if PKEY really works, but we assume that it does.

At the end, this check for `pkey_wrap_supported` only checks, that even though all prereqs for PKEY are given, it might still have failed to produce a protected key verification pattern (for whatever reasons). It is certainly rather unlikely that this happens..... In our case it has failed due to the bug with the XCP_MFL_PROBE flag version checking. 

The prereqs for PKEY are already checked in the if statement in `ep11tok_is_mechanism_supported()`, like checking the MSA level, and checking if CKM_IBM_CPACF_WRAP is supported (checks card versions). So we can assume hat PKEY works if those prereqs are given.

ifranzki:
While you are at it, could you please add another commit to remove the german comment `// Einziger Unterschied: extractable, statt !extractable` in ep11_specific.c line 1395 ?

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: OK, done. 
Should we probably add a comment to the commit, or even in the code, about these relations?

ifranzki: I would suggest to add some text to the commit message explaining this. 

Something like:

If it is already known that the PKEY wrap is not supported or not functioning (for whatever reason), then don't report the XTS mechanisms as supported.

jschmidb: OK, done.

ifranzki (approved):

PR #753, opened 2024-03-13
COMMON: Don't clear context_len before calling context_free_func

ifranzki:
The context_len field of an encrypt/decrypt, sign/verify, or digest context should not be set to zero before the context_free_func callback is called. This callback gets the context_len as argument, so setting it to zero before calling the callback may cause problems inside the callback.

Currently, none of the callbacks used makes use of the context length, so this does not actually cause any problems. However, future callbacks might rely on the context length being passed correctly.


jschmidb (approved):
lgtm

PR #755, opened 2024-03-15
p11sak fixes and a minor CCA code simplification

ifranzki:


jschmidb (approved):
lgtm. Yes, replacing the huge switch stmt by the loop makes much sense.

PR #756, opened 2024-03-20
EP11: Reject combined extract attribute settings if it is not supported

ifranzki:
In case the control point setting of the adapters do not allow that attributes CKA_EXTRACTABLE and CKA_IBM_PROTKEY_EXTRACTABLE are both true, then reject this with CKR_ATTRIBUTE_VALUE_INVALID.

The EP11 code would reject that with CKR_FUNCTION_CANCELED, which for EP11 it means that it violates an internal policy (i.e. control point settings), but in PKCS#11 this return code has a totally different meaning. So reject such situations explicitly with the correct return code.

### DIFF
--- a/usr/lib/ep11_stdll/ep11_specific.c
+++ b/usr/lib/ep11_stdll/ep11_specific.c
@@ -1094,20 +1094,23 @@ static CK_BBOOL ep11tok_pkey_session_ok_for_obj(SESSION *session,
  * Returns true if the given key object is eligible to get a protected key
  * attribute, false otherwise.
  */
-CK_BBOOL ep11tok_pkey_obj_eligible_for_pkey_support(ep11_private_data_t *ep11_data,
-                                                    OBJECT *key_obj)
+static CK_RV ep11tok_pkey_obj_eligible_for_pkey_support(
+                                                ep11_private_data_t *ep11_data,
+                                                OBJECT *key_obj)
 {
     if (object_is_attr_bound(key_obj) || !ep11_data->pkey_wrap_supported ||
         !object_is_pkey_extractable(key_obj)) {
-        return CK_FALSE;
+        return CKR_FUNCTION_NOT_SUPPORTED;
     }
 
     if (!ep11_data->pkey_combined_extract_supported &&
         object_is_extractable(key_obj)) {
-        return CK_FALSE;
+        TRACE_ERROR("Combined extract not supported, but CKA_EXTRACTABLE "
+                    "and CKA_IBM_PROTKEY_EXTRACTABLE are both TRUE\n");
+        return CKR_TEMPLATE_INCONSISTENT;
     }
 
-    return CK_TRUE;
+    return CKR_OK;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


jschmidb:
Looks good. But we only handle the case where an application tries to create a new object with both attrs true. The case where an existing token obj with both attrs true is just used, is handled in routine ep11tok_pkey_check, where routine ep11tok_pkey_obj_eligible_for_pkey_support would return false, and the rc in ep11tok_pkey_check would remain CKR_FUNCTION_NOT_SUPPORTED. Should we probably initialise this rc with CKR_ATTRIBUTE_VALUE_INVALID to make it consistent?

ifranzki:
> Should we probably initialise this rc with CKR_ATTRIBUTE_VALUE_INVALID to make it consistent?

No, CKR_FUNCTION_NOT_SUPPORTED is correct there, because it tells the caller that it can't use protected keys, so the caller will use secure keys instead. This is what we want.

ifranzki:
Hm well, the secure key usage will then probably also fail.....
So maybe your idea isn't that bad. Returning CKR_ATTRIBUTE_VALUE_INVALID in that case will cause the operation to be rejected entirely... Not sure if CKR_ATTRIBUTE_VALUE_INVALID is the right return code though... Maybe CKR_KEY_FUNCTION_NOT_PERMITTED or CKR_KEY_UNEXTRACTABLE ?

ifranzki:
Hm, or maybe CKR_TEMPLATE_INCONSISTENT ? Maybe the check in token_specific_set_attrs_for_new_object should also return CKR_TEMPLATE_INCONSISTENT ?

jschmidb:
The calling functions all check for CKR_FUNCTION_NOT_SUPPORTED and fall back into the ep11 path in this case. In all other cases the function fails entirely. So yes, let's use another rc, where it doesn't matter which one we use from a logic point of view. It's more like giving the application the right info.

ifranzki:
There is one place where ep11tok_pkey_check() is called from ep11tok_pkey_usage_ok(). There it just checks for CKR_OK. So returning a bad return code would not make a difference. It would not make it fail either...

ifranzki:
@jschmidb please take another look. Now both places should be handled.

jschmidb (approved):
perfect!

ifranzki:
There are still cases where we won't detect such key usage which have both attrs TRUE:
SC_Decrypt() calls ep11tok_pkey_usage_ok() which returns FALSE in that case, so it takes the ep11tok_decrypt_single() path. This will then cause EP11 to fail with CKR_FUNCTION_CANCELED...

Same happens for Encrypt, Sign, Verify. Basically everyone that calls ep11tok_pkey_usage_ok(). 
I guess I have to change ep11tok_pkey_usage_ok() to also return a CKR value instead of a bool to be able to catch the combined situation there as well.

jschmidb:
Ah right, here the rc values get lost ...

ifranzki:
Fixed these remaining cases, too. Changed ep11tok_pkey_usage_ok() to return a CK_RV now. 
@jschmidb Please check carefully if the logic is still the same as before....

ifranzki:
@jschmidb any further feedback? Or are we ready to merge ?

jschmidb:
ok, I reviewed it again and see no issues. So let's go ahead and merge. I will then rebase my commit with the new test for combined extract and open a pr for that.

PR #758, opened 2024-04-02
p11sak: Don't copy an empty CKA_ID with extract-pubkey

ifranzki:
Copying an empty (zero length) CKA_ID attribute does not make much sense, because the default CKA_ID value is empty anyway.

As reported by @akulhalli in https://github.com/opencryptoki/opencryptoki/pull/749#issuecomment-2031165150


jschmidb (approved):
Yes, makes sense. There's a second place in p11sak_extract_x509_pk where this check is already coded like this.

akulhalli:
Wouldn't this skip adding zero-length attributes? I'm not very familiar with the PKCS#11 interface - does it require some attributes to always be included in the template, even if it is zero-length (like `CKA_ID`?)

akulhalli (approved):
Ah, never mind. I misread the location of the hunk - this change is only for the `CKA_ID` attribute.

lgtm

PR #760, opened 2024-04-05
p11sak: Add support to export a key or certificate as URI-PEM

ifranzki:
The pkcs11-provider from https://github.com/latchset/pkcs11-provider supports PEM files containing the PKCS#11 URI instead of the key or certificate material. Enhance p11sak to allow to export such URI-PEM files, optionally including the PKCS#11 user pin via `pin-value` (insecure) or `pin-source` query attribute of the URI.

@jschmidb Please give it a try if such a URI-PEM (with or without the pin) generated by p11sak really works with the pkcs11-provider.


jschmidb (approved):
lgtm

jschmidb (approved):
test looks good.

PR #761, opened 2024-04-09
Support new gcm API in icatoken

jschmidb:
The new libica gcm api (ica_aes_gcm_kma_*) exploits the KMA instruction on z14 and later. This results in better performance.

### DIFF
--- a/usr/lib/common/mech_aes.c
+++ b/usr/lib/common/mech_aes.c
@@ -3462,20 +3462,25 @@ CK_RV aes_gcm_encrypt_update(STDLL_TokData_t *tokdata, SESSION *sess,
     }
 
     context = (AES_GCM_CONTEXT *) ctx->context;
-    total = (context->len + in_data_len);
+    total = context->len + in_data_len;
+    remain = total % AES_BLOCK_SIZE;
+    out_len = total - remain;
 
     if (length_only) {
         if (total < AES_BLOCK_SIZE) {
             *out_data_len = 0;
             return CKR_OK;
         } else {
-            remain = (total % AES_BLOCK_SIZE);
-            out_len = total - remain;
             *out_data_len = out_len;
             TRACE_DEVEL("Length Only requested (%02ld bytes).\n",
                         *out_data_len);
             return CKR_OK;
         }
+    }
+
+    if (*out_data_len < out_len) {
+        TRACE_ERROR("%s\n", ock_err(ERR_BUFFER_TOO_SMALL));
+        return CKR_BUFFER_TOO_SMALL;
     }
 
     if (token_specific.t_aes_gcm_update == NULL) {
@@ -3507,7 +3512,6 @@ CK_RV aes_gcm_encrypt_final(STDLL_TokData_t *tokdata, SESSION *sess,
     }
 
     context = (AES_GCM_CONTEXT *) ctx->context;
-
     aesgcm = (CK_GCM_PARAMS *) ctx->mech.pParameter;
     tag_data_len = (aesgcm->ulTagBits + 7) / 8; /* round to full byte */
 
@@ -3602,25 +3606,29 @@ CK_RV aes_gcm_decrypt_update(STDLL_TokData_t *tokdata, SESSION *sess,
      * that means it's tag data, not encrypted plaintext.
      * Hence we'll keep at least tag data size in the context buffer */
 
-    context = (AES_GCM_CONTEXT *) ctx->context;
-    total = (context->len + in_data_len);
-
     aesgcm = (CK_GCM_PARAMS *) ctx->mech.pParameter;
+    context = (AES_GCM_CONTEXT *) ctx->context;
+
+    total = context->len + in_data_len;
     tag_data_len = (aesgcm->ulTagBits + 7) / 8; /* round to full byte */
+    remain = ((total - tag_data_len) % AES_BLOCK_SIZE) + tag_data_len;
+    out_len = total - remain;
 
     if (length_only) {
         if (total < AES_BLOCK_SIZE + tag_data_len) {
             *out_data_len = 0;
             return CKR_OK;
         } else {
-            remain = ((total - tag_data_len) % AES_BLOCK_SIZE)
-                + tag_data_len;
-            out_len = total - remain;
             *out_data_len = out_len;
             TRACE_DEVEL("Length Only requested (%02ld bytes).\n",
                         *out_data_len);
             return CKR_OK;
         }
+    }
+
+    if (*out_data_len < out_len) {
+        TRACE_ERROR("%s\n", ock_err(ERR_BUFFER_TOO_SMALL));
+        return CKR_BUFFER_TOO_SMALL;
     }
 
     if (token_specific.t_aes_gcm_update == NULL) {
@@ -3659,6 +3667,11 @@ CK_RV aes_gcm_decrypt_final(STDLL_TokData_t *tokdata, SESSION *sess,
             *out_data_len = context->len;
         }
         return CKR_OK;
+    }
+
+    if (*out_data_len < context->len) {
+        TRACE_ERROR("%s\n", ock_err(ERR_BUFFER_TOO_SMALL));
+        return CKR_BUFFER_TOO_SMALL;
     }
 
     if (token_specific.t_aes_gcm_final == NULL) {
--- a/usr/lib/common/mech_aes.c
+++ b/usr/lib/common/mech_aes.c
@@ -3516,11 +3516,7 @@ CK_RV aes_gcm_encrypt_final(STDLL_TokData_t *tokdata, SESSION *sess,
     tag_data_len = (aesgcm->ulTagBits + 7) / 8; /* round to full byte */
 
     if (length_only) {
-        if (context->len == 0) {
-            *out_data_len = tag_data_len;
-        } else {
-            *out_data_len = context->len + tag_data_len;
-        }
+        *out_data_len = context->len + tag_data_len;
         return CKR_OK;
     }
 
@@ -3661,11 +3657,7 @@ CK_RV aes_gcm_decrypt_final(STDLL_TokData_t *tokdata, SESSION *sess,
     context = (AES_GCM_CONTEXT *) ctx->context;
 
     if (length_only) {
-        if (context->len == 0) {
-            *out_data_len = 0;
-        } else {
-            *out_data_len = context->len;
-        }
+        *out_data_len = context->len;
         return CKR_OK;
     }
 
--- a/usr/lib/ica_s390_stdll/ica_specific.c
+++ b/usr/lib/ica_s390_stdll/ica_specific.c
@@ -3609,7 +3609,7 @@ CK_RV token_specific_aes_gcm(STDLL_TokData_t *tokdata, SESSION *sess,
     }
 
     if (rc != 0) {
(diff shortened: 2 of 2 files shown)


usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -3125,6 +3125,31 @@ static CK_RV os_specific_rsa_decrypt(STDLL_TokData_t *tokdata,
 
 }
 
+static CK_BBOOL save_data_in_context(CK_BYTE *in_data, CK_ULONG in_data_len,
+                                     CK_ULONG tag_data_len,
+                                     AES_GCM_CONTEXT *context, CK_BYTE encrypt,
+                                     CK_ULONG *total, CK_ULONG *remain)
```
ifranzki, line 3131: why is `total` a by-reference variable? It is not updated in this routine, so it can be passed by-value.

jschmidb: Right, this was a leftover from copying the inline code into the subroutine. 
Done.

usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -4639,6 +5019,10 @@ static CK_RV mech_list_ica_initialize(STDLL_TokData_t *tokdata)
         if (libica_func_list[i].mech_mode_id == AES_CBC)
             ica_data->ica_aes_available = TRUE;
 
+        /* Remember if libica supports the new AES-GCM API (z14 and later) */
+        if (libica_func_list[i].mech_mode_id == AES_GCM_KMA)
```
ifranzki, line 5019: Can it be that AES_GCM_KMA is available but AES_CBC is not available ?
If that can happen, then even if AES_GCM_KMA is available, it will always take the OpenSSL software fallback, because AES_CBC is not available.

usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -3124,6 +3125,364 @@ static CK_RV os_specific_rsa_decrypt(STDLL_TokData_t *tokdata,
...
+    context = (AES_GCM_CONTEXT *)ctx->context;
+    gcm_ctx = (kma_ctx *)context->ulClen;
+    aes_gcm_param = (CK_GCM_PARAMS *)ctx->mech.pParameter;
+    auth_data = (CK_BYTE *)aes_gcm_param->pAAD;
+    auth_data_len = aes_gcm_param->ulAADLen;
+    tag_data_len = (aes_gcm_param->ulTagBits + 7) / 8;
+    total = (context->len + in_data_len);
+
+    /* if there isn't enough data to make a block, just save it */
+    if (encrypt) {
```
ifranzki: above you make `encrypt` UNUSED (even with a comment), but here you are still using it ?

usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -3124,6 +3125,364 @@ static CK_RV os_specific_rsa_decrypt(STDLL_TokData_t *tokdata,
...
+
+    if (encrypt) {
+        if (context->len != 0) {
+            buffer = (CK_BYTE *)malloc(AES_BLOCK_SIZE);
+            if (!buffer) {
+                TRACE_ERROR("%s\n", ock_err(ERR_HOST_MEMORY));
+                rc = CKR_HOST_MEMORY;
+                goto done;
+            }
+            memcpy(buffer, context->data, context->len);
```
ifranzki, line 3412: Why copying `context->data` to the freshly allocated buffer? Why don't you directly pass `context->data` to `ica_aes_gcm_kma_update` below?

usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -3124,6 +3125,364 @@ static CK_RV os_specific_rsa_decrypt(STDLL_TokData_t *tokdata,
...
+    } else {
+        /* decrypt */
+        if (context->len > tag_data_len) {
+            buffer = (CK_BYTE *)malloc(AES_BLOCK_SIZE);
+            if (!buffer) {
+                TRACE_ERROR("%s\n", ock_err(ERR_HOST_MEMORY));
+                rc = CKR_HOST_MEMORY;
+                goto done;
+            }
+            memcpy(buffer, context->data, context->len - tag_data_len);
```
ifranzki, line 3452: Again, why copying into a buffer? you can use context->data right away with ica_aes_gcm_kma_update below.

usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -3124,6 +3125,364 @@ static CK_RV os_specific_rsa_decrypt(STDLL_TokData_t *tokdata,
...
+
+            /* Perform final update without data to calculate final tag */
+            rc = ica_aes_gcm_kma_update(NULL, NULL, 0, NULL, 0, 1, 1, gcm_ctx);
+        } else {                /* (context->len < tag_data_len) */
+            TRACE_ERROR("Incoming data are not consistent.\n");
+            rc = CKR_DATA_INVALID;
+            goto done;
+        }
+
+        final_tag_data = context->data + context->len - tag_data_len;
```
ifranzki, line 3477: Is it guaranteed that this does not overrun the `context->data` buffer? `context->data` is `2*AES_BLOCK_SIZE` in size. `context->len` is probably not larger than one `AES_BLOCK_SIZE`, so one `AES_BLOCK_SIZE `is left. Is this sufficient for `tag_data_len` ?

usr/lib/ica_s390_stdll/ica_specific.c
jschmidb: Hm, I would say no. But we could put
```
if (ica_data->ica_new_gcm_available)
        return new_gcm_specific_aes_gcm_update/final(...)
```
before 
```
if (!ica_data->ica_aes_available)
        return openssl_specific_aes_gcm_update/final(...)
```
But in the token_specific_aes_gcm_init routine we would then also have to put 
```
if (!ica_data->ica_aes_available)
        return openssl_specific_aes_gcm_init(...);
```
down, after getting the key value. What do you think? On a messed up system with KMA available, but CBC not, we would then use openssl. Would that matter?

usr/lib/ica_s390_stdll/ica_specific.c
jschmidb: Argh, the final routine ... yes, removed it.

usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -3124,6 +3125,364 @@ static CK_RV os_specific_rsa_decrypt(STDLL_TokData_t *tokdata,
...
+    auth_data = (CK_BYTE *)aes_gcm_param->pAAD;
+    auth_data_len = aes_gcm_param->ulAADLen;
+    tag_data_len = (aes_gcm_param->ulTagBits + 7) / 8;
+    total = (context->len + in_data_len);
+
+    /* if there isn't enough data to make a block, just save it */
+    if (encrypt) {
+        remain = (total % AES_BLOCK_SIZE);
+        if (total < AES_BLOCK_SIZE) {
+            memcpy(context->data + context->len, in_data, in_data_len);
```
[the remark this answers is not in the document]
jschmidb: Yes, the new subroutine just contains copied code that was at two places inline before. The logic shouldn't be changed.

usr/lib/ica_s390_stdll/ica_specific.c
ifranzki: Hm, no, leave it as is.

usr/lib/ica_s390_stdll/ica_specific.c
jschmidb: Yes, valid tag sizes are from 4 bytes up to 16 bytes.

usr/lib/ica_s390_stdll/ica_specific.c
jschmidb: ok, done.

usr/lib/ica_s390_stdll/ica_specific.c
jschmidb: done.

usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -3124,6 +3125,364 @@ static CK_RV os_specific_rsa_decrypt(STDLL_TokData_t *tokdata,
...
+                                    0, gcm_ctx);
+    }
+
+    if (rc != 0) {
+        TRACE_ERROR("ica_aes_gcm_kma_update failed, rc=%ld.\n", rc);
+        rc = CKR_FUNCTION_FAILED;
+        goto done;
+    }
+
+    (*out_data_len) = out_len;
```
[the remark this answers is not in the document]
jschmidb: Good point! This check was actually missing in the existing code. I added two additional commits providing the output buffer length check and also simplified some logic in the two gcm_final routines.

usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -3124,6 +3125,364 @@ static CK_RV os_specific_rsa_decrypt(STDLL_TokData_t *tokdata,
...
+            /* Perform final update with rest of data to calculate final tag */
+            rc = ica_aes_gcm_kma_update(buffer, out_data, context->len,
+                                        NULL, 0, 1,1, gcm_ctx);
+            if (rc != 0) {
+                TRACE_ERROR("ica_aes_gcm_kma_update failed, rc=%ld\n",rc);
+                rc = CKR_FUNCTION_FAILED;
+                goto done;
+            }
+
+            *out_data_len = context->len + tag_data_len;
```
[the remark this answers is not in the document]
jschmidb: see above, done.

usr/lib/ica_s390_stdll/ica_specific.c
ifranzki: Hm... in line 3270 in this function ist still marked as UNUSED.

usr/lib/common/mech_aes.c
```diff
@@ -3602,27 +3606,31 @@ CK_RV aes_gcm_decrypt_update(STDLL_TokData_t *tokdata, SESSION *sess,
      * that means it's tag data, not encrypted plaintext.
      * Hence we'll keep at least tag data size in the context buffer */
 
+    aesgcm = (CK_GCM_PARAMS *) ctx->mech.pParameter;
     context = (AES_GCM_CONTEXT *) ctx->context;
-    total = (context->len + in_data_len);
 
-    aesgcm = (CK_GCM_PARAMS *) ctx->mech.pParameter;
+    total = (context->len + in_data_len);
```
ifranzki, line 3608: brackets around the addition are not needed.

usr/lib/ica_s390_stdll/ica_specific.c
jschmidb: Hmm ... I don't have it in my local version of ica_specific.c ...

usr/lib/common/mech_aes.c
jschmidb: OK, removed.

usr/lib/ica_s390_stdll/ica_specific.c
ifranzki: its still there
image

ifranzki (approved):
lgtm

PR #768, opened 2024-04-26
CCA: Add support for the IBM Dilithium mechanism

ifranzki:
CCA supports different Dilithium variants with different CCA versions:

**CCA on Linux on IBM Z:**

- Dilithium (6,5) Round 2 is only available for CCA releases 7.1 and later.
- Dilithium (8,7) Round 2, Dilithium (6,5) Round 3 and Dilithium (8,7) Round 3 are only available for CCA releases 8.0 and later. 
- Dilithium (4,4) Round 3 is not supported at all by CCA on Linux on IBM Z.

**CCA on AIX, Linux on x64, Linux on IBM Power Systems:**

- Dilithium (6,5) Round 2 is only available for CCA releases 7.2 and later.
- Other variants are currently not supported at all by CCA on these platforms.

The mechanism `CKM_IBM_DILITHIUM` is only supported if at least CCA release 7.1 (Linux on IBM Z) or 7.2 (AIX, Linux on x64, Linux on IBM Power Systems) is available.

@akulhalli I think this should also work on AIX, but please give it a try. Please also double check the version checks in function `cca_pqc_strength_supported()`, this should be according to the documentation in the CCA Basic Service Guide for AIX, Linux on x64, Linux on IBM Power Systems. It checks for `#ifdef __s390__` to differentiate between the CCA on Linux on IBM Z case and the other architectures. I guess this should be sufficient for covering the AIX case, too.


usr/lib/cca_stdll/cca_stdll.h
```diff
@@ -33,7 +33,7 @@
 
 #define CCA_PRIVATE_KEY_NAME_SIZE       64
 #define CCA_REGENERATION_DATA_SIZE      64
-#define CCA_KEY_TOKEN_SIZE              2500
+#define CCA_KEY_TOKEN_SIZE              8000
```
jschmidb, line 36: Does this new size of 8000 bytes cover the rounded up max key token length of 7648 bytes of a CRYSTALS-Dilithium (8,7 Round 2) QSA token?

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -9261,6 +9274,598 @@ static CK_RV import_ec_pubkey(STDLL_TokData_t *tokdata, TEMPLATE *pub_templ)
...
+                                    sizeof(CK_BBOOL));
+        if (rc != CKR_OK) {
+            TRACE_DEVEL("build_update_attribute for CKA_SENSITIVE failed. "
+                        "rc=0x%lx\n", rc);
+            return rc;
+        }
+    } else {
+        /*
+         * This is an import of a clear IBM dilithium private key which is to be
+         * transfered into a CCA QSA private key.
```
jschmidb, line 9671: transferred

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -9261,6 +9274,598 @@ static CK_RV import_ec_pubkey(STDLL_TokData_t *tokdata, TEMPLATE *pub_templ)
...
+            return CKR_FUNCTION_FAILED;
+        }
+
+        if (analyse_cca_key_token(target_key_token, target_key_token_length,
+                                  &token_type, &token_keybitsize,
+                                  &mkvp) == FALSE ||
+            mkvp == NULL) {
+            TRACE_ERROR("Invalid/unknown cca token has been imported\n");
+            return CKR_FUNCTION_FAILED;
+        }
```
jschmidb: Is this possible? Successfully importing a key, but afterwards we realize that it's invalid/unknown?

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -9261,6 +9274,598 @@ static CK_RV import_ec_pubkey(STDLL_TokData_t *tokdata, TEMPLATE *pub_templ)
...
+
+        /* Add key token to template as CKA_IBM_OPAQUE */
+        if ((rc = build_update_attribute(priv_templ, CKA_IBM_OPAQUE,
+                                         target_key_token,
+                                         target_key_token_length))) {
+            TRACE_DEVEL("build_update_attribute(CKA_IBM_OPAQUE) failed\n");
+            return rc;
+        }
+
+        /* zero clear key values */
```
jschmidb, line 9807: zeroize (?)

usr/lib/cca_stdll/cca_stdll.h
ifranzki: The CCA doc mentions the following for parameter key_token_length on the PKA Key Token Build (CSNDPKB) verb (and possibly at other places):

_On input, a size of 8000 bytes is sufficient to contain the largest key_token created._

Thats basically where this length comes from.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -9261,6 +9274,598 @@ static CK_RV import_ec_pubkey(STDLL_TokData_t *tokdata, TEMPLATE *pub_templ)
...
+                        rc);
+            free(spki);
+            return rc;
+        }
+
+        free(spki);
+    } else {
+        /*
+         * This is an import of a clear IBM dilithium public key which is to be
+         * transfered into a CCA QSA public key.
```
jschmidb, line 9935: transferred

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -9384,6 +9989,33 @@ CK_RV token_specific_object_add(STDLL_TokData_t *tokdata, SESSION *sess, OBJECT
             return CKR_KEY_TYPE_INCONSISTENT;
         }
         break;
+    case CKK_IBM_PQC_DILITHIUM:
+        switch(keyclass) {
```
jschmidb, line 10177: space after switch ?

usr/lib/cca_stdll/cca_specific.c
ifranzki: fixed

usr/lib/cca_stdll/cca_specific.c
ifranzki: It might be a key token that `analyse_cca_key_token()` does not support. 
In any case, we need to handle the error case, even if it is unlikely that it happens.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -7797,6 +7797,190 @@ CK_RV token_specific_ibm_dilithium_generate_keypair(STDLL_TokData_t *tokdata,
...
+                    (long *)signature_len, &signature_bit_length, signature);
+    RETRY_NEW_MK_BLOB_END(tokdata, return_code, reason_code,
+                          attr->pValue, attr->ulValueLen)
+    USE_CCA_ADAPTER_END(tokdata, return_code, reason_code)
+
+    if (return_code != CCA_SUCCESS) {
+        TRACE_ERROR("CSNDDSG (QSA SIGN) failed. return:%ld,"
+                    " reason:%ld\n", return_code, reason_code);
+        return CKR_FUNCTION_FAILED;
+    } else if (reason_code != 0) {
```
jschmidb, line 7902: Are you using "0" instead of CCA_SUCCESS intentionally when comparing the reason code?

usr/lib/cca_stdll/cca_specific.c
ifranzki: done

usr/lib/cca_stdll/cca_specific.c
ifranzki: yep, fixed, (2 more places above).

usr/lib/cca_stdll/cca_specific.c
ifranzki: CCA_SUCCESS is just meant for the return code, not for the reason code.... 
Could still be used there as well though....
There are hundred of similar cases, so I guess I keep it consistent with all the other places.

testcases/misc_tests/cca_export_import_test.c
```diff
@@ -349,6 +349,62 @@ static CK_RV import_ecc_publ_key(CK_SESSION_HANDLE session,
...
+    CK_BBOOL false = FALSE;
+    CK_ATTRIBUTE template[] = {
+        {CKA_CLASS, &keyClass, sizeof(keyClass)},
+        {CKA_KEY_TYPE, &keyType, sizeof(keyType)},
+        {CKA_LABEL, (char *) label, strlen(label) + 1},
+        {CKA_SIGN, &true, sizeof(true)},
+        {CKA_TOKEN, &false, sizeof(false)},
+        {CKA_IBM_OPAQUE, ccatoken, tokenlen}
+    };
+    CK_ULONG nattr = sizeof(template)/sizeof(CK_ATTRIBUTE);
```
jschmidb, line 370: spaces around the / ?

ifranzki: Yea, fixed many other such places, too....

testcases/misc_tests/cca_export_import_test.c
```diff
@@ -1611,6 +1675,288 @@ static CK_RV cca_ecc_export_import_tests(void)
...
+            {CKA_VERIFY, &attr_verify, sizeof(CK_BBOOL)},
+            {CKA_IBM_DILITHIUM_KEYFORM,
+             (CK_BYTE *)&dilithium_variants[i].keyform, sizeof(CK_ULONG)},
+        };
+        CK_ULONG num_dilithium_attrs =
+                sizeof(dilithium_attr_public) / sizeof(CK_ATTRIBUTE);
+
+        testcase_begin("CCA export/import test with public/private IBM Dilithium %s keys",
+                       dilithium_variants[i].name);
+
```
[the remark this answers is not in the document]
ifranzki: fixed

testcases/misc_tests/cca_export_import_test.c
```diff
@@ -1611,6 +1675,288 @@ static CK_RV cca_ecc_export_import_tests(void)
...
+        // verify with re-imported public key
+
+        rc = funcs->C_VerifyInit(session, &mech, imp_publ_key);
+        if (rc != CKR_OK) {
+            testcase_error("C_VerifyInit() with re-imported pub key failed, rc=%s",
+                           p11_get_ckr(rc));
+            goto error;
+        }
+        rc = funcs->C_Verify(session, msg, msglen, sig, siglen);
+        if (rc == CKR_OK) {
```
jschmidb, line 1915: Hm, why not a switch for the rc? Would probably be better readable ...

ifranzki: Yea, again, there are many other places in that test doing it the same way....
Not sure if I really like to change all of these places in this commit....

akulhalli:
@ifranzki the first `C_Sign` testcase fails due to a zero input length error - as expected. What was the idea behind this test case?

If I disable that array entry, everything else passes as expected. I believe this might be the same glibc quirk returning a special memory address for a zero-length `malloc` call. Does that test case pass every time for you? It could have been reading arbitrary data from memory - the man page mentions dereferencing that address will result in undefined behaviour.

ifranzki:
> the first `C_Sign` testcase fails due to a zero input length error - as expected. What was the idea behind this test case?

Hm, let me look into this. 
I guess the idea was that it should also be possible to sign the empty message. If I remember correctly this works on EP11, and I don't remember seeing an error on CCA with Linux on IBM Z either.... I'll check again. 
If needed, we just need to exclude this case for CCA.... no big deal.

ifranzki:
Yes this works for me every time. The zero length input length gets through to the CSNDDSG verb, and even that seems to be OK with a zero length here. 

But yes, I do get a valid pointer from `calloc` on allocating zero bytes.... The content of that allocated buffer is irrelevant, because no one ever reads any bytes from it. 

Anyway, would the following change in dilithium_func.c help? 

```
-    data = calloc(inputlen, sizeof(CK_BYTE));
+    data = calloc(inputlen + 1, sizeof(CK_BYTE));
     if (data == NULL) {
         testcase_error("Can't allocate memory for %lu bytes",
-                       sizeof(CK_BYTE) * inputlen);
+                       sizeof(CK_BYTE) * (inputlen + 1));
         rc = -1;
         goto testcase_cleanup;
     }
```

akulhalli:
Nope, I tried something similar already - instead of doing a `malloc` for size 0, I instead skipped the malloc call altogether - that's when I get a "CKR_ARGUMENTS_BAD` return from `C_Sign`.

It doesn't seem to like an input size of 0 - though I haven't verified whether it is the actual CCA verb failing or some other malloc before it that failed.

ifranzki:
Yea, passing a NULL as input to `C_Sign` will result in CKR_ARGUMENTS_BAD. 
So even for the empty message, you need to pass a valid pointer. 

What exactly do you get if allocating a +1 length but still pass a zero length to `C_Sign` ?(i.e. like the proposed change does) ?

akulhalli:
Yeah, that works fine.

But I'd much rather assign a random "address" to the pointer if the size is 0 rather than add arbitrary numbers to bypass the `NULL` check.

The requirement of the address being valid even if the input size is 0 should probably be documented, since that relies on implicit behaviour of a specific platform - although granted that _AIX_ is the minority here since it works everywhere else :)

ifranzki:
> The requirement of the address being valid even if the input size is 0 should probably be documented

Well, I really meant **non-NULL** :-) it does not really be valid.

I am just trying to find out if the PKCS#11 standard states anything in regards of calling `C_Sign` with a NULL input data pointer and a zero input length. Haven't found anything so far....

akulhalli:
>I'd much rather assign a random "address" to the pointer if the size is 0

Disregard this, it'll be a problem for the `free` later on :-)

I'm fine with a 1-byte buffer being passed with a zero `inputlen` size, if the standard doesn't specify a behaviour.

ifranzki:
> I'm fine with a 1-byte buffer being passed with a zero `inputlen` size, if the standard doesn't specify a behaviour.

I have now made it like this:
`data = calloc(inputlen > 0 ? inputlen : 1, sizeof(CK_BYTE));`

jschmidb (approved):
lgtm

ifranzki:
@akulhalli Is this OK with you, too? If so, please approve.

akulhalli (approved):
lgtm

PR #778, opened 2024-06-05
Misc fixes

ifranzki:


jschmidb:
Lgtm. Wonder if we should also update libica and libzpc to Jammy. We updated to focal in Nov 2023.

jschmidb (approved):

ifranzki:
Fixed typos in commit message

ifranzki:
> Wonder if we should also update libica and libzpc to Jammy. 

The main reason I am changing to Jammy is because on Focal I get errors like the following on x86 and arm:
```
sudo apt-get -qq update
W: GPG error: http://repo.mongodb.org/apt/ubuntu focal/mongodb-org/4.4 Release: The following signatures were invalid: EXPKEYSIG 656408E390CFB1F5 MongoDB 4.4 Release Signing Key <packaging@mongodb.com>
E: The repository 'http://repo.mongodb.org/apt/ubuntu focal/mongodb-org/4.4 Release' is not signed.
```

PR #780, opened 2024-06-10
CCA const time fixes

ifranzki:


jschmidb:
Looks good ... just saw that in ep11token new_host.c and common/new_host.c there are similar places.

PR #781, opened 2024-06-10
Support to import RSA keys with ME components only

ifranzki:
PKCS#11 is flexible with respect to which attributes must be present in an RSA key. Keys can be specified in Chinese-Remainder format or they can be specified in modular-exponent format. Effective with version 2.40, tokens MUST also store CKA_PUBLIC_EXPONENT. This permits the retrieval of sufficient data to reconstitute the associated public key.

When creating an RSA private key via C_CreateObject: either all CRT attributes must be specified, or none of them. Specifying only some of them is invalid and causes CKR_TEMPLATE_INCONSISTENT. As per PKCS#11 v2.40, attributes CKA_MODULUS, CKA_PRIVATE_EXPONENT and CKA_PUBLIC_EXPONENT must always be specified.

### DIFF
--- a/testcases/crypto/rsa_func.c
+++ b/testcases/crypto/rsa_func.c
@@ -352,7 +352,8 @@ testcase_cleanup:
  * 5. Compare plaintext with decrypted data
  *
  */
-CK_RV do_EncryptDecryptImportRSA(struct PUBLISHED_TEST_SUITE_INFO *tsuite)
+CK_RV do_EncryptDecryptImportRSA(struct PUBLISHED_TEST_SUITE_INFO *tsuite,
+                                 CK_BBOOL me_only)
 {
     unsigned int i, j;
     CK_BYTE original[BIG_REQUEST];
@@ -373,7 +374,8 @@ CK_RV do_EncryptDecryptImportRSA(struct PUBLISHED_TEST_SUITE_INFO *tsuite)
     char *s;
 
     // begin testsuite
-    testsuite_begin("%s Encrypt Decrypt Import.", tsuite->name);
+    testsuite_begin("%s Encrypt Decrypt Import%s.", tsuite->name,
+                    me_only ? " (ME-format)" : "");
     testcase_rw_session();
     testcase_user_login();
 
@@ -426,9 +428,9 @@ CK_RV do_EncryptDecryptImportRSA(struct PUBLISHED_TEST_SUITE_INFO *tsuite)
             goto testcase_cleanup;
         }
         // begin testcase
-        testcase_begin("%s Encrypt and Decrypt Import with test vector %u."
+        testcase_begin("%s Encrypt and Decrypt Import with test vector %u%s."
                        "\npubl_exp='%s', modbits=%lu, publ_exp_len=%lu.",
-                       tsuite->name, i, s,
+                       tsuite->name, i, me_only ? " (ME-format)" : "", s,
                        tsuite->tv[i].mod_len * 8,
                        tsuite->tv[i].pubexp_len);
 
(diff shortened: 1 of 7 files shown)


usr/lib/common/key.c
```diff
@@ -2188,48 +2179,22 @@ CK_RV rsa_priv_check_required_attributes(TEMPLATE *tmpl, CK_ULONG mode)
...
+    template_attribute_get_non_empty(tmpl, CKA_EXPONENT_2, &exp2);
+    template_attribute_get_non_empty(tmpl, CKA_COEFFICIENT, &coef);
 
-    rc = template_attribute_get_non_empty(tmpl, CKA_COEFFICIENT, &attr);
-    if (rc != CKR_OK) {
-        if (mode == MODE_CREATE) {
-            TRACE_ERROR("Could not find CKA_COEFFICIENT\n");
-            return rc;
-        }
+    /* For CREATE either all CRT components or non of them must be specified */
```
jschmidb, line 2189: none

usr/lib/common/key.c
```diff
@@ -2332,6 +2332,326 @@ CK_BBOOL rsa_priv_check_exportability(CK_ATTRIBUTE_TYPE type)
...
+    CK_BYTE *buff = NULL;
+    CK_RV rc;
+
+    bn_ctx = BN_CTX_secure_new();
+    if (bn_ctx == NULL) {
+        TRACE_ERROR("BN_CTX_secure_new failed\n");
+        return CKR_FUNCTION_FAILED;
+    }
+
+    /* Get modulus a BIGNUM */
```
jschmidb, line 2355: as ?

usr/lib/common/key.c
ifranzki: fixed

usr/lib/common/key.c
ifranzki: fixed

usr/lib/common/key.c
```diff
@@ -2153,24 +2154,14 @@ CK_RV rsa_priv_check_required_attributes(TEMPLATE *tmpl, CK_ULONG mode)
...
-    // full number of attributes have been specified.  Do we revert back to
-    // modular-exponent representation?  Do we compute the missing attributes
-    // ourselves?  Do we simply return CKR_TEMPLATE_INCOMPLETE?
-    //
+
+    /*
+     * PKCS#11 is flexible with respect to which attributes must be present
+     * in an RSA key. Keys can be specified in Chinese-Remainder format or
+     * they can be specified in modular-exponent format. Effective with version
+     * 2.40, tokens MUST also store CKA_PUBLIC_EXPONENT. This permits the
```
jschmidb, line 2162: Maybe a bit more clear:
... MUST store the CKA_PUBLIC_EXPONENT in any case.

usr/lib/common/mech_openssl.c
```diff
@@ -4898,6 +4898,325 @@ CK_RV openssl_specific_hmac_final(SIGN_VERIFY_CONTEXT *ctx, CK_BYTE *signature,
     return rv;
 }
 
+CK_RV calc_rsa_crt_from_me(CK_ATTRIBUTE *modulus, CK_ATTRIBUTE *pub_exp,
+                           CK_ATTRIBUTE *priv_exp, CK_ATTRIBUTE **prime1,
+                           CK_ATTRIBUTE **prime2, CK_ATTRIBUTE **exponent1,
+                           CK_ATTRIBUTE **exponent2, CK_ATTRIBUTE **coef)
+{
```
jschmidb, line 4905: Just to double check: you already checked that either all CRT components are given, or none of them. Otherwise the calculation could either rely on already given components or check if they are identical to the calculated ones.

usr/lib/common/mech_openssl.c
```diff
@@ -4898,6 +4898,325 @@ CK_RV openssl_specific_hmac_final(SIGN_VERIFY_CONTEXT *ctx, CK_BYTE *signature,
...
+    if (BN_set_word(two, 2) != 1 ||
+        BN_sub(n_minus_1, n, BN_value_one()) != 1) {
+        TRACE_ERROR("BN_set_word/BN_sub failed\n");
+        rc = CKR_FUNCTION_FAILED;
+        goto done;
+    }
+
+    /*
+     * Prime-Factor Recovery from n, e and d described in NIST Special
+     * Publication 800-56B R1 Recommendation for Pair-Wise Key Establishment
```
jschmidb, line 4977: Looks like Revision 1 is withdrawn, see https://csrc.nist.gov/pubs/sp/800/56/b/r1/final
Maybe we should reference Rev2. The algorithm is the same.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -8386,148 +8386,200 @@ static CK_RV import_rsa_privkey(STDLL_TokData_t *tokdata, TEMPLATE * priv_tmpl)
...
-        rc = template_attribute_get_non_empty(priv_tmpl, CKA_PRIME_1, &p_prime);
-        if (rc != CKR_OK) {
-            TRACE_ERROR("CKA_PRIME_1 attribute missing for CRT.\n");
-            return rc;
+            is_me_format = TRUE;
         }
-        total += p_prime->ulValueLen;
 
-        rc = template_attribute_get_non_empty(priv_tmpl, CKA_PRIME_2, &q_prime);
+        /* Public exponent and Modulus is always required */
```
jschmidb, line 8417: are always required

testcases/crypto/rsa_func.c
```diff
@@ -373,7 +374,8 @@ CK_RV do_EncryptDecryptImportRSA(struct PUBLISHED_TEST_SUITE_INFO *tsuite)
     char *s;
 
     // begin testsuite
-    testsuite_begin("%s Encrypt Decrypt Import.", tsuite->name);
+    testsuite_begin("%s Encrypt Decrypt Import%s.", tsuite->name,
+                    me_only ? " (ME-formatpo)" : "");
```
jschmidb, line 378: Not sure about this string "ME-formatpo" ...  has this a special meaning? Or is the "po" some leftover?

usr/lib/common/mech_openssl.c
ifranzki: Correct, a key can only be created with either all or none of them.

usr/lib/common/key.c
ifranzki: fixed

usr/lib/common/mech_openssl.c
ifranzki: Fixed (to R2, also in the commit description)

usr/lib/cca_stdll/cca_specific.c
ifranzki: fixed

testcases/crypto/rsa_func.c
ifranzki: left over from copy & paste - fixed

jschmidb (approved):

Issue #782, opened 2024-06-12
multiple defects found by covscan in opencryptoki-3.23.0


ngothan:
There are multiple defects in opencryptoki-3.23.0 found by covscan. It could be some of them are false positives.
Thanks!

```
Error: RESOURCE_LEAK (CWE-772):
opencryptoki-3.23.0/usr/lib/common/shared_memory.c:301: alloc_fn: Storage is returned from allocation function "mmap".
opencryptoki-3.23.0/usr/lib/common/shared_memory.c:301: var_assign: Assigning: "addr" = storage returned from "mmap(NULL, real_len, 3, 1, fd, 0L)".
opencryptoki-3.23.0/usr/lib/common/shared_memory.c:314: var_assign: Assigning: "ctx" = "addr".
opencryptoki-3.23.0/usr/lib/common/shared_memory.c:349: leaked_storage: Variable "ctx" going out of scope leaks the storage it points to.
opencryptoki-3.23.0/usr/lib/common/shared_memory.c:349: leaked_storage: Variable "addr" going out of scope leaks the storage it points to.
# 347| free(name);
# 348|
# 349|-> return rc;
# 350| }
# 351|

Error: RESOURCE_LEAK (CWE-772):
opencryptoki-3.23.0/usr/sbin/pkcsslotd/socket_server.c:657: alloc_fn: Storage is returned from allocation function "calloc".
opencryptoki-3.23.0/usr/sbin/pkcsslotd/socket_server.c:657: var_assign: Assigning: "conn" = storage returned from "calloc(1UL, 160UL)".
opencryptoki-3.23.0/usr/sbin/pkcsslotd/socket_server.c:728: noescape: Resource "conn" is not freed or pointed-to in "proc_hangup".
opencryptoki-3.23.0/usr/sbin/pkcsslotd/socket_server.c:732: leaked_storage: Variable "conn" going out of scope leaks the storage it points to.
# 730| }
# 731|
# 732|-> return rc;
# 733| }
# 734|

Error: INTEGER_OVERFLOW (CWE-190):
opencryptoki-3.23.0/usr/lib/icsf_stdll/icsf.c:152: ineffective_check: The check "len - str_len > 0UL", which appears to be a guard against integer overflow, is not a useful guard because it is either always true, or never true. This taints "len".
opencryptoki-3.23.0/usr/lib/icsf_stdll/icsf.c:153: overflow: The expression "len - str_len" might be negative, but is used in a context that treats it as unsigned.
[...]

PR #783, opened 2024-06-13
Fix errors reported by covscan

ifranzki:
Closes: https://github.com/opencryptoki/opencryptoki/issues/782

### DIFF
--- a/usr/lib/common/loadsave.c
+++ b/usr/lib/common/loadsave.c
@@ -2848,6 +2848,14 @@ CK_RV load_public_token_objects(STDLL_TokData_t *tokdata)
             continue;
         }
 
+        /* size can not be negative if treated as signed int */
+        if (size >= 0x80000000) {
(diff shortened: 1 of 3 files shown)


usr/lib/common/loadsave.c
```diff
@@ -2848,6 +2848,14 @@ CK_RV load_public_token_objects(STDLL_TokData_t *tokdata)
             continue;
         }
 
+        /* size can not be negative if treated as signed int */
```
jschmidb, line 2851: as unsigned int

ifranzki: no, the problem arises if it is treated as **signed** int, because then its value would be negative.

jschmidb (approved):
ok, lgtm as discussed.

ifranzki:
Close + reopen to trigger Travis

PR #784, opened 2024-06-14
TRAVIS: Add one more job to build with OpenSSL 1.1.1

ifranzki:
Ubuntu 'jammy' (22.04) comes with OpenSSL 3.0, which is what most distributions use nowadays as well. However, it must still be ensured that it builds with OpenSSL 1.1.1. Add one job to run on Ubuntu 'focal' (20.04) that still comes with OpenSSL 1.1.1. Build this on s390x only, since this covers the largest set of source files to be built. If it builds on s390x, then it should also build on other architectures.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,11 @@ dist: jammy
 language: c
 
 before_install:
-    - sudo apt-get -qq update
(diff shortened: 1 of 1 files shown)


jschmidb (approved):
ok, looks good.

PR #785, opened 2024-06-14
TESTCASES: Remove unnecessary SKIP messages for test that can never run

ifranzki:
Some tests can never run, regardless of what the token supports or not. Don't produce SKIP messages for those to reduce the output size.


jschmidb (approved):
lgtm

PR #789, opened 2024-06-19
COMMON: correct rc handling with s390_pcc function

jschmidb:
Let s390_pcc return the condition code of the PCC instruction. Also fix the error handling in the calling functions.

### DIFF
--- a/usr/lib/common/pkey_utils.c
+++ b/usr/lib/common/pkey_utils.c
@@ -234,27 +234,30 @@ static int s390_kmac(unsigned long func, void *param,
 
 /**
  * s390_pcc:
- * @func: the function code passed to KM; see s390_pcc_functions
+ * @func: the function code passed to PCC; see s390_pcc_functions
  * @param: address of parameter block; see POP for details on each func
  *
  * Executes the PCC operation of the CPU.
  *
- * Returns -1 for failure, 0 for the query func, number of processed
- * bytes for encryption/decryption funcs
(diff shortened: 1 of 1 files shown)


usr/lib/common/pkey_utils.c
```diff
@@ -234,27 +234,30 @@ static int s390_kmac(unsigned long func, void *param,
...
-
-    __asm__ volatile (
-        "0: .long   %[opc] << 16\n"
-        "   brc 1,0b\n"
-        :
-        : [fc] "d" (r0), [param] "a" (r1), [opc] "i" (0xb92c)
-        : "cc", "memory");
-
-    return 0;
+	register unsigned long r0 __asm__("0") = (unsigned long)func;
```
ifranzki, line 246: Please use OCK coding style, i.e. 4 blanks instead of tabs.

jschmidb: Fixed.

jschmidb:
Yes, I ran the aes tests on ep11 token with 'PKEY_MODE ENABLE4ALL', and on CCA token with 'PKEY_MODE = ENABLED'. I got calls to s390_pcc from CCA when performing the xts tests. Cannot say why I'm not getting calls from the cmac tests in both tokens.

jschmidb:
> But for EP11, I guess you should get calls to s390_pcc for CMAC.

I'm getting the calls now! It was a config problem on my test system.

PR #790, opened 2024-06-24
CCA: Also treat RSA-OAEP error 8/2053 as CKR_ENCRYPTED_DATA_INVALID

ifranzki:
CCA error 8/2053 means "No message found in the OAEP-decoded data" and is an additional error (in addition to the already handled 8/2054 error: "There is a non-valid RSA Enciphered Key cryptogram: OAEP optional encoding parameters failed validation") that could be returned by CCA verb CSNDPKD in case of RSA-OAEP on invalid encoded messages.

Translate this error also to CKR_ENCRYPTED_DATA_INVALID to not allow any distinction between these two error situations. Until now 8/2053 was translated to CKR_FUNCTION_FAILED, like any other CCA error codes.


jschmidb (approved):
lgtm

PR #793, opened 2024-07-09
TESTCASES: Generate unique label for do_EncryptDecryptAESPkey() test

ifranzki:
To allow running the testcase concurrently against the same token, generate a unique name, so that one process does not find another process's token object, because this would cause the testcase to fail, although its not a real failure, just an effect of running concurrently.

### DIFF
--- a/testcases/crypto/aes_func.c
+++ b/testcases/crypto/aes_func.c
@@ -12,6 +12,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <memory.h>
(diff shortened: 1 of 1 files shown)


GrgoMariani (approved):
I've checked out this PR locally.
Compilation was successful and without any warnings.
The test was executed by running concurrent 16 threads separated by a 0.5 second time window each for 128 iterations.
No problems encountered.

Source code changes look good. I would say this is ready to merge.

PR #800, opened 2024-07-31
Support live guest relocations for protected key ops

jschmidb:
This pr adds support for  live guest relocations (LGR) for protected key operations to the CCA and EP11 tokens.

### DIFF
--- a/usr/lib/cca_stdll/cca_specific.c
+++ b/usr/lib/cca_stdll/cca_specific.c
@@ -3309,8 +3309,9 @@ CK_RV token_specific_aes_xts(STDLL_TokData_t *tokdata, SESSION *session,
     rc = ccatok_pkey_check(tokdata, session, key_obj, &mech);
     switch (rc) {
     case CKR_OK:
-        rc = pkey_aes_xts(key_obj, init_v, in_data, in_data_len,
-                          out_data, out_data_len, encrypt, initial, final, iv);
+        rc = pkey_aes_xts(tokdata, session, key_obj, init_v,
+                          in_data, in_data_len, out_data, out_data_len,
+                          encrypt, initial, final, iv);
         goto done;
     default:
         goto done;
@@ -5771,7 +5772,7 @@ CK_RV token_specific_aes_ecb(STDLL_TokData_t * tokdata,
     rc = ccatok_pkey_check(tokdata, session, key, &mech);
     switch (rc) {
     case CKR_OK:
-        rc = pkey_aes_ecb(key, in_data, in_data_len,
+        rc = pkey_aes_ecb(tokdata, session, key, in_data, in_data_len,
                           out_data, out_data_len, encrypt);
         goto done;
     case CKR_FUNCTION_NOT_SUPPORTED:
@@ -5911,7 +5912,7 @@ CK_RV token_specific_aes_cbc(STDLL_TokData_t * tokdata,
     rc = ccatok_pkey_check(tokdata, session, key, &mech);
     switch (rc) {
     case CKR_OK:
-        rc = pkey_aes_cbc(key, init_v, in_data, in_data_len,
+        rc = pkey_aes_cbc(tokdata, session, key, init_v, in_data, in_data_len,
                           out_data, out_data_len, encrypt);
         goto done;
     case CKR_FUNCTION_NOT_SUPPORTED:
--- a/usr/lib/cca_stdll/cca_specific.c
+++ b/usr/lib/cca_stdll/cca_specific.c
@@ -3289,7 +3289,7 @@ static CK_RV ccatok_pkey_check_attrs(STDLL_TokData_t *tokdata,
         return CKR_OK;
 
     /*
-     * At this point, the key has CKA_IBM_PROKEY_EXTRACTABLE = true and it has
+     * At this point, the key has CKA_IBM_PROTKEY_EXTRACTABLE = true and it has
      * a secure key token, so let's check if the secure key token is
      * CPACF-exportable.
      */
@@ -3305,7 +3305,8 @@ static CK_RV ccatok_pkey_check_attrs(STDLL_TokData_t *tokdata,
          */
         if (pkey_op_ec_curve_supported_by_cpacf(templ) &&
             !ccatok_token_is_cpacf_exportable(sec_key, sec_len)) {
-            TRACE_ERROR("ECC secure key is CKA_IBM_PROTKEY_EXTRACTABLE, but token is not CPACF-exportable.\n");
+            TRACE_ERROR("ECC secure key is CKA_IBM_PROTKEY_EXTRACTABLE, "
+                        "but token is not CPACF-exportable.\n");
             return CKR_TEMPLATE_INCONSISTENT;
         }
         break;
@@ -3314,7 +3315,6 @@ static CK_RV ccatok_pkey_check_attrs(STDLL_TokData_t *tokdata,
     }
 
     return CKR_OK;
-
 }
 
 CK_RV ccatok_pkey_check_aes_xts(TEMPLATE *tmpl)
--- a/usr/lib/cca_stdll/cca_specific.c
+++ b/usr/lib/cca_stdll/cca_specific.c
@@ -5981,7 +5981,8 @@ CK_RV token_specific_aes_ecb(STDLL_TokData_t * tokdata,
     switch (rc) {
     case CKR_OK:
         rc = pkey_aes_ecb(tokdata, session, key, in_data, in_data_len,
-                          out_data, out_data_len, encrypt);
+                          out_data, out_data_len, encrypt,
+                          ccatok_pkey_convert_key);
         goto done;
     case CKR_FUNCTION_NOT_SUPPORTED:
         /* fallback */
@@ -6121,7 +6122,8 @@ CK_RV token_specific_aes_cbc(STDLL_TokData_t * tokdata,
     switch (rc) {
     case CKR_OK:
         rc = pkey_aes_cbc(tokdata, session, key, init_v, in_data, in_data_len,
-                          out_data, out_data_len, encrypt);
+                          out_data, out_data_len, encrypt,
+                          ccatok_pkey_convert_key);
         goto done;
     case CKR_FUNCTION_NOT_SUPPORTED:
         /* fallback */
--- a/usr/lib/cca_stdll/cca_specific.c
+++ b/usr/lib/cca_stdll/cca_specific.c
@@ -3511,7 +3511,8 @@ CK_RV token_specific_aes_xts(STDLL_TokData_t *tokdata, SESSION *session,
     case CKR_OK:
         rc = pkey_aes_xts(tokdata, session, key_obj, init_v,
                           in_data, in_data_len, out_data, out_data_len,
-                          encrypt, initial, final, iv);
+                          encrypt, initial, final, iv,
+                          ccatok_pkey_convert_key);
         goto done;
     default:
         goto done;
--- a/usr/lib/cca_stdll/cca_specific.c
+++ b/usr/lib/cca_stdll/cca_specific.c
@@ -2695,10 +2695,20 @@ static CK_RV ccatok_pkey_get_firmware_wkvp(STDLL_TokData_t *tokdata)
     }
 
     /* Save WKVP in token data */
+    if (pthread_rwlock_wrlock(&cca_data->pkey_rwlock)) {
+        TRACE_ERROR("%s Failed to lock pkey lock\n", __func__);
+        ret = CKR_CANT_LOCK;
+        goto done;
+    }
+
     memcpy(&cca_data->pkey_mk_vp, (CK_BYTE *)pkey_attr->pValue + AES_KEY_SIZE_256,
            PKEY_MK_VP_LENGTH);
 
     __sync_or_and_fetch(&cca_data->pkey_wrap_supported, 1);
+
+    if (pthread_rwlock_unlock(&cca_data->pkey_rwlock)) {
+        TRACE_ERROR("%s Failed to unlock pkey lock\n", __func__);
+    }
 
 done:
 
@@ -2733,20 +2743,29 @@ static CK_BBOOL ccatok_pkey_is_valid(STDLL_TokData_t *tokdata, OBJECT *key_obj)
     struct cca_private_data *cca_data = tokdata->private_data;
     CK_ATTRIBUTE *pkey_attr = NULL;
     int vp_offset;
+    CK_BBOOL ret = CK_FALSE;
 
     if (template_attribute_get_non_empty(key_obj->template, CKA_IBM_OPAQUE_PKEY,
                                          &pkey_attr) == CKR_OK) {
         if (pkey_attr->ulValueLen >= AES_KEY_SIZE_128 + PKEY_MK_VP_LENGTH) {
             vp_offset = pkey_attr->ulValueLen - PKEY_MK_VP_LENGTH;
+            if (pthread_rwlock_rdlock(&cca_data->pkey_rwlock)) {
+                TRACE_ERROR("%s Failed to lock pkey lock\n", __func__);
+                goto done;
+            }
             if (memcmp((CK_BYTE *)pkey_attr->pValue + vp_offset,
                        &cca_data->pkey_mk_vp,
                        PKEY_MK_VP_LENGTH) == 0) {
-                return CK_TRUE;
+                ret = CK_TRUE;
+            }
+            if (pthread_rwlock_unlock(&cca_data->pkey_rwlock)) {
+                TRACE_ERROR("%s Failed to unlock pkey lock\n", __func__);
             }
         }
     }
 
-    return CK_FALSE;
+done:
+    return ret;
 }
 
 /**
@@ -2761,6 +2780,7 @@ static CK_RV ccatok_pkey_update(STDLL_TokData_t *tokdata, OBJECT *key_obj,
     CK_ATTRIBUTE *pkey_attr = NULL;
     CK_RV ret;
     int vp_offset;
+    int num_retries = 0;
 
     /* Get secure key from obj */
     if (template_attribute_get_non_empty(key_obj->template, CKA_IBM_OPAQUE,
@@ -2770,6 +2790,7 @@ static CK_RV ccatok_pkey_update(STDLL_TokData_t *tokdata, OBJECT *key_obj,
         goto done;
     }
 
+retry:
     /* Transform the secure key into a protected key */
     ret = ccatok_pkey_skey2pkey(tokdata, skey_attr, &pkey_attr, aes_xts);
     if (ret != CKR_OK) {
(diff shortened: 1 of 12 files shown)


usr/lib/cca_stdll/cca_specific.c
```diff
@@ -2695,8 +2695,16 @@ static CK_RV ccatok_pkey_get_firmware_wkvp(STDLL_TokData_t *tokdata)
...
+        ret = CKR_CANT_LOCK;
+        goto done;
+    }
     memcpy(&cca_data->pkey_mk_vp, (CK_BYTE *)pkey_attr->pValue + AES_KEY_SIZE_256,
            PKEY_MK_VP_LENGTH);
+    if (pthread_rwlock_unlock(&cca_data->pkey_rwlock)) {
+        TRACE_ERROR("%s Failed to unlock pkey lock\n", __func__);
+    }
 
     __sync_or_and_fetch(&cca_data->pkey_wrap_supported, 1);
```
ifranzki, line 2709: Please include this line also into the lock.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -2770,26 +2788,78 @@ static CK_RV ccatok_pkey_update(STDLL_TokData_t *tokdata, OBJECT *key_obj,
...
+        vp_offset = pkey_attr->ulValueLen / 2 - PKEY_MK_VP_LENGTH;
+        if (memcmp(&cca_data->pkey_mk_vp, (CK_BYTE *)pkey_attr->pValue + vp_offset,
+                   PKEY_MK_VP_LENGTH) != 0) {
+            TRACE_ERROR("vp of this pkey does not match with the one in cca_data\n");
+            ret = CKR_FUNCTION_FAILED;
+        }
+    }
+
+    if (pthread_rwlock_unlock(&cca_data->pkey_rwlock))
+        TRACE_ERROR("%s Failed to unlock pkey lock\n", __func__);
```
ifranzki, line 2834: Please use { ... } around the TRACE_ERROR stmt. Its a single line block, so { } would not be needed, but TRACE_ERROR is a macro and you don't know to what it expands. If it expands to multiple lines, then only the very first expanded line would be part of the if stmt, but the remaining lines not.

usr/lib/cca_stdll/cca_specific.c
```diff
@@ -2939,6 +2939,125 @@ static CK_RV ccatok_pkey_check(STDLL_TokData_t *tokdata, SESSION *session,
...
+        }
+    }
+
+    /* Update wkvp in CCA private data. */
+    memcpy(&cca_private->pkey_mk_vp,
+           (CK_BYTE *)pkey_attr->pValue + pkey_attr->ulValueLen - PKEY_MK_VP_LENGTH,
+           PKEY_MK_VP_LENGTH);
+
+    /* Pass back new protkey. Need to do this before unlocking the obj. */
+    memcpy(protkey, pkey_attr->pValue, pkey_attr->ulValueLen);
```
ifranzki, line 3116: Instead of checking for `*protkey_len < MAXPROTKEYSIZE - 16` at the beginning of the function, I would rather check here for  `*protkey_len < pkey_attr->ulValueLen` and return here with CKR_BUFFER_TOO_SMALL if it is too small.
While we know that `pkey_attr->ulValueLen` will most likely not be larger than `MAXPROTKEYSIZE - 16`, ist still better to check here.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -1297,6 +1300,141 @@ CK_RV ep11tok_pkey_add_protkey_attr_to_tmpl(TEMPLATE *tmpl)
...
+    }
+    memcpy(&ep11_data->pkey_mk_vp,
+           (CK_BYTE *)pkey_attr->pValue + pkey_attr->ulValueLen - PKEY_MK_VP_LENGTH,
+           PKEY_MK_VP_LENGTH);
+    if (pthread_mutex_unlock(&ep11_data->pkey_mutex)) {
+        TRACE_ERROR("%s Failed to lock pkey lock\n", __func__);
+    }
+
+    /* Pass back new protkey. Need to do this before unlocking the obj. */
+    memcpy(protkey, pkey_attr->pValue, pkey_attr->ulValueLen);
```
ifranzki, line 1471: same as for the CCA token: Check here for length, instead of at the beginning of the function.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -16,6 +16,9 @@
 #include <regex.h>
 #include <dirent.h>
 #include <libgen.h>
+#ifndef NO_PKEY
+#include <asm/pkey.h>
```
ifranzki, line 20: For what exactly do you need this include here? Is it for `MAXPROTKEYSIZE` only ? If so, then you might not need that anymore when changing the length check as suggested below.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -1297,6 +1300,141 @@ CK_RV ep11tok_pkey_add_protkey_attr_to_tmpl(TEMPLATE *tmpl)
...
+     * the attribute when we have the write lock here.
+     */
+    if (template_attribute_get_non_empty(key_obj->template,
+                                         CKA_IBM_OPAQUE, &skey_attr) != CKR_OK) {
+        TRACE_ERROR("This key has no blob: should not occur!\n");
+        rc = CKR_FUNCTION_FAILED;
+        goto done;
+    }
+
+    /* Convert secure key to protected key. */
```
ifranzki, line 1399: Move that comment to below the retry label, it seems to fit better there.

usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -1021,34 +1032,78 @@ static CK_RV ep11tok_pkey_update(STDLL_TokData_t *tokdata, SESSION *session,
...
+    /*
+     * Recreating the firmware mkvp via ep11tok_pkey_get_firmware_mk_vp below
+     * does only work when pkey_wrap_support_checked is 0. Set it to 0 while
+     * the mutex lock is still obtained.
+     */
+    if (ret != CKR_OK)
+        __sync_and_and_fetch(&ep11_data->pkey_wrap_support_checked, 0);
+
+    if (pthread_mutex_unlock(&ep11_data->pkey_mutex))
+        TRACE_ERROR("%s Failed to unlock pkey lock\n", __func__);
```
ifranzki, line 1078: Please use { .. } (TRACE_ERROR is a macro).

usr/lib/common/pkey_utils.c
```diff
@@ -546,14 +549,28 @@ CK_RV pkey_aes_ecb(STDLL_TokData_t *tokdata, SESSION *session,
 
     /* Call CPACF */
     memcpy(param.key, pkey_attr->pValue, pkey_attr->ulValueLen);
-    bytes_processed = s390_km(fc, &param, out_data, in_data, in_data_len);
-    if (bytes_processed <= 0) {
+    while (len > 0 && num_tries < PKEY_CONVERT_KEY_RETRIES) {
+        bytes_processed = s390_km(fc, &param, out_pos, in_pos, len);
+        if (bytes_processed < len) {
+            ret = convert_key(tokdata, session, key_obj, CK_FALSE,
```
ifranzki, line 555: I just noticed that we do not see any trace messages when such a convert_key case happens. The convert_key routine does not itself issue any trace messages (unless an error occurs). It would be good to see this in the trace in case we ever need to debug an LGR case. So please add trace messages here before calling convert_key(), and at all other places where the convert function is called (also in the next 2 commits). You may include the bytes_processed and len variables into the trace message.

usr/lib/cca_stdll/cca_specific.c
jschmidb: fixed.

usr/lib/cca_stdll/cca_specific.c
jschmidb: fixed.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: fixed.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: fixed.

usr/lib/ep11_stdll/ep11_specific.c
jschmidb: fixed.

usr/lib/common/pkey_utils.c
jschmidb: ok, trace msgs are added.

usr/lib/common/pkey_utils.c
```diff
@@ -908,18 +908,25 @@ struct aes_xts_param {
...
-    int offset, rc;
+    int offset, rc = 1, num_tries = 0;
 
     offset = AES_XTS_PCC_I(param->keylen * 8);
     memcpy(param->param_pcc + offset, tweak, AES_BLOCK_SIZE);
 
-    rc = s390_pcc(param->fc & 0x7f, param->param_pcc);
+    while (rc != 0 && num_tries < PKEY_CONVERT_KEY_RETRIES) {
+        rc = s390_pcc(param->fc & 0x7f, param->param_pcc);
+        num_tries++;
```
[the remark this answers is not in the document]
ifranzki: Shouldn't there also be a call to `convert_key()` here? Just retrying `s390_pcc()` alone will not make any difference and willfail again and again ...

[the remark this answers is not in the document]
jschmidb: Arghh, yes, definitely. Will fix.

usr/lib/common/pkey_utils.c
```diff
@@ -908,18 +908,43 @@ struct aes_xts_param {
...
+            ret = convert_key(param->tokdata, param->session, param->key_obj,
+                              CK_TRUE, protkey, &protkey_len);
+            if (ret != CKR_OK)
+                break;
+            /*
+             * Copy XTS key 2 into CPACF parmblock for PCC instruction. Key 2
+             * is used to create the iv from tweak via PCC. The key object
+             * itself now contains both keys in rederived form.
+             */
+            memcpy(param->param_pcc, protkey + protkey_len / 2, protkey_len / 2);
```
ifranzki, line 944: I guess you should also copy the re-derived key part 1 into `param->param_km` here.

Initially both params blocks are setup with the key material in `pkey_aes_xts()` before calling `aes_xts_cipher()`.
`aes_xts_cipher()` then first calls `pkey_aes_xts_iv_from_tweak()` and after that calls `pkey_aes_xts_cipher_blocks()` multiple times.

So if the key must be re-derived inside `pkey_aes_xts_iv_from_tweak()` (due to LGR), then you must also update the key in the `param_km` so that the subsequent calls to `pkey_aes_xts_cipher_blocks()` work on the re-derived key as well.

If an LGR happens during `pkey_aes_xts_cipher_blocks()`, it is OK to only update the key in the `param_km`, since `pkey_aes_xts_iv_from_tweak()` is not called anymore.

jschmidb: Yes, that makes sense. Added the memcpy for key 1 and some comment.

usr/lib/common/pkey_utils.c
```diff
@@ -908,18 +908,50 @@ struct aes_xts_param {
...
+                              CK_TRUE, protkey, &protkey_len);
+            if (ret != CKR_OK)
+                break;
+            /*
+             * Copy XTS key 1 into CPACF parmblock for KM instruction. After
+             * re-deriving the iv from tweak, the subsequent calls to
+             * pkey_aes_xts_cipher_blocks will then use the re-derived key
+             * as well.
+             */
+            memcpy(param->param_km, protkey, protkey_len / 2);
```
ifranzki, line 945: There are many places in the PKEY code where the protected key is copied to a CPACF param block without any length checks. While it is very unlikely that `protkey_len` is larger than `param_km`, it is not impossible. `param_km` is 64+16 bytes, but the protkey buffer is MAXPROTKEYSIZE+2 in size, which is most likely larger than 64+16.

This does not only affect this place, but many places. Most of the time the protected key is copied right from the attribute, e.g. like this:
`memcpy(param.key, pkey_attr->pValue, pkey_attr->ulValueLen);`
No check if `pkey_attr->ulValueLen` is larger than the target of the copy.....

Since attributes can theoretically be set/updated by a malicious user program with data of any size (i.e. larger size), just copying this without a length check can cause memory overlays, etc.  

We should add explicit length checks at all these places, or at least do something like this to prevent copying too many bytes:

`memcpy(param.key, pkey_attr->pValue, MIN(pkey_attr->ulValueLen, sizeof(param.key)));`

Having a too sort or otherwise invalid protected key in the parm block will cause a failure later, and a possible rederive of the prot key. So this will be handled anyway.  

Please do so in the commits if this PR where such memcpy's are added, and add a new commit on top to fix all the other places not anyway touched by the commits of this PR.

jschmidb: Done.

usr/lib/common/pkey_utils.c
```diff
@@ -1200,7 +1200,7 @@ CK_RV pkey_ec_sign(STDLL_TokData_t *tokdata, SESSION *session,
             hash_len = 32;
         off = 32 - hash_len;
         memcpy(param.P256.hash + off, hash, hash_len);
-        memcpy(param.P256.priv, pkey_attr->pValue, 32);
+        memcpy(param.P256.priv, pkey_attr->pValue, MIN(pkey_attr->ulValueLen - 32, 32));
```
ifranzki, line 1203: please use `sizeof(param.P256.priv)` instead of the hardcoded 32 for the second MIN parameter.
Same for all the other places.

usr/lib/common/pkey_utils.c
```diff
@@ -1200,7 +1200,7 @@ CK_RV pkey_ec_sign(STDLL_TokData_t *tokdata, SESSION *session,
...
         memcpy(param.P256.vp, (char *)pkey_attr->pValue + 32, 32);
```
ifranzki, line 1204: Use `MIN(..., sizeof(param.P256.vp))` here as well.
Also, did you check anywhere that `pkey_attr->ulValueLen` is at least 32+32 bytes in size? Otherwise you access memory after the actual allocation.

usr/lib/common/pkey_utils.c
```diff
@@ -1579,7 +1579,7 @@ struct {                          \
         hash_off = 32 - hash_len;
         memcpy(param.P256.sig_r, sig, sig_len);
         memcpy(param.P256.hash + hash_off, hash, hash_len);
-        memcpy(param.P256.pub_x, ecpoint, ecpoint_len);
+        memcpy(param.P256.pub_x, ecpoint, MIN(ecpoint_len, sig_len));
```
ifranzki, line 1582: has sig_len be checked to be valid ? This also comes right from the user application an can thus be anything....

ifranzki: Further below theris the following:

```
    case curve_ed448:
        rc = s390_kdsa(fc, edparam.buff, msg, msg_len);
        if (sig[113] != 0) {
            /* KDSA doesn't check last byte */
            rc = 1;
        }
```

Did you check that `sig_len` is at least 112 bytes ?

usr/lib/common/pkey_utils.c
```diff
@@ -1196,31 +1196,52 @@ CK_RV pkey_ec_sign(STDLL_TokData_t *tokdata, SESSION *session,
...
+        }
         if (hash_len > 32)
             hash_len = 32;
         off = 32 - hash_len;
         memcpy(param.P256.hash + off, hash, hash_len);
-        memcpy(param.P256.priv, pkey_attr->pValue, 32);
-        memcpy(param.P256.vp, (char *)pkey_attr->pValue + 32, 32);
+        memcpy(param.P256.priv, pkey_attr->pValue,
+               MIN(pkey_attr->ulValueLen - 32, sizeof(param.P256.priv)));
+        memcpy(param.P256.vp, (char *)pkey_attr->pValue + 32, sizeof(param.P256.vp));
```
ifranzki, line 1211: Since you have the explicit check above, you can omit the MIN here. 

However, instead of the hardcoded +/- 32, I would use `sizeof(xxxx)` like in the following:

```
memcpy(param.P256.priv, pkey_attr->pValue, pkey_attr->ulValueLen - sizeof(param.P256.vp), sizeof(param.P256.priv)));
memcpy(param.P256.vp, (char *)pkey_attr->pValue + sizeof(param.P256.priv), sizeof(param.P256.vp));
```

Similar for the `32` for the hash, can we use a `sizeof(something)` here ? Isn't `sizeof(param.P256.hash)` exactly what you need ?

jschmidb: It's the vp that must be subtracted. I removed the MIN(...) for p256, p384, and p521 and made 1 single line then. Let me know if you want to have a line break before col 80 .... it's always some tradeoff between readability and format conventions.

usr/lib/common/pkey_utils.c
```diff
@@ -1573,35 +1605,65 @@ struct {                          \
...
+            TRACE_ERROR("Public key has an invalid length of %ld bytes.\n", ecpoint_len);
+            ret = CKR_ATTRIBUTE_VALUE_INVALID;
+            goto done;
+        }
         if (hash_len > sig_len / 2)
             hash_len = sig_len / 2;
         fc = KDSA_ECDSA_VERIFY_P384;
         hash_off = 48 - hash_len;
-        memcpy(param.P384.sig_r, sig, sig_len);
+        memcpy(param.P384.sig_r, sig, MIN(sig_len, 2 * sizeof(param.P384.sig_r)));
```
ifranzki, line 1641: so here you assume `sig_r` and `sig_s` are the same length (which is probably true). But in above length check you add sizeof of each of them individually ? I reader wonders why is this done differently ????
I would use either `2 * xxx` everywhere, or add the individual sizes everywhere, but don't mix it.

Same below for `pub_x` and `pub_y`.

usr/lib/common/pkey_utils.c
ifranzki: Yes please no lines linger than 80.

But you still have a +32 in the second memcpy:
`memcpy(param.P256.vp, (char *)pkey_attr->pValue + 32, sizeof(param.P256.vp));`
I guess that one should be
`memcpy(param.P256.vp, (char *)pkey_attr->pValue + sizeof(param.P256.priv), sizeof(param.P256.vp));`

jschmidb: ok, done.

usr/lib/common/pkey_utils.c
jschmidb: ok, done

usr/lib/common/pkey_utils.c
```diff
@@ -1404,15 +1428,29 @@ CK_RV pkey_ibm_ed_sign(STDLL_TokData_t *tokdata, SESSION *session,
...
+                        pkey_attr->ulValueLen);
+            ret = CKR_ATTRIBUTE_VALUE_INVALID;
+            goto done;
+        }
         fc = KDSA_ENCRYPTED_EDDSA_SIGN_ED25519;
-        memcpy(edparam.ED25519.priv, pkey_attr->pValue, 32);
-        memcpy(edparam.ED25519.vp, (char *)pkey_attr->pValue + 32, 32);
+        memcpy(edparam.ED25519.priv, pkey_attr->pValue,
+               MIN(pkey_attr->ulValueLen, sizeof(edparam.ED25519.priv)));
+        memcpy(edparam.ED25519.vp, (char *)pkey_attr->pValue + 32, sizeof(edparam.ED25519.vp));
```
ifranzki, line 1440: here is still a `+ 32`

usr/lib/common/pkey_utils.c
```diff
@@ -1573,35 +1611,65 @@ struct {                          \
...
         fc = KDSA_ECDSA_VERIFY_P384;
         hash_off = 48 - hash_len;
-        memcpy(param.P384.sig_r, sig, sig_len);
+        memcpy(param.P384.sig_r, sig, MIN(sig_len, 2 * sizeof(param.P384.sig_r)));
         memcpy(param.P384.hash + hash_off, hash, hash_len);
-        memcpy(param.P384.pub_x, ecpoint, ecpoint_len);
+        memcpy(param.P384.pub_x, ecpoint, MIN(ecpoint_len, 2 * sizeof(param.P384.pub_x)));
         break;
     case curve_p521:
+        if (sig_len != 2 * (sizeof(param.P521.sig_r) - 14)) {
```
ifranzki, line 1652: Why `- 14` ? Where does the 14 come from?

usr/lib/common/pkey_utils.c
```diff
@@ -1573,35 +1611,65 @@ struct {                          \
...
+            TRACE_ERROR("Signature has an invalid length of %ld bytes.\n", sig_len);
+            ret = CKR_ARGUMENTS_BAD;
+            goto done;
+        }
+        if (ecpoint_len != 2 * (sizeof(param.P521.pub_x) - 14)) {
```
ifranzki, line 1657: same here

usr/lib/common/pkey_utils.c
```diff
@@ -1681,16 +1749,26 @@ struct {                            \
...
+            ret = CKR_ARGUMENTS_BAD;
+            goto done;
+        }
         fc = KDSA_EDDSA_VERIFY_ED25519;
         s390_flip_endian_32(edparam.ED25519.sig_r, sig);
         s390_flip_endian_32(edparam.ED25519.sig_s, sig + (sig_len / 2));
         s390_flip_endian_32(edparam.ED25519.pub, ecpoint);
         break;
     case curve_ed448:
+        if (sig_len != 2 * (sizeof(edparam.ED448.sig_r) - 7)) {
```
ifranzki, line 1763: Why `- 7` ?

jschmidb: The 2 buffers for r and s each have 64 bytes, but an EdDSA signature has only 114 bytes. Similar to the 80 bytes buffers for p521, but signatures of 132 bytes.

ifranzki: Ah ok, yes, makes sense. Please add a comment here to explain that, otherwise we will wonder in a few weeks again :-)

jschmidb: ok, fixed all places and re-pushed.

jschmidb:
Replaced some more numbers by their related buffer sizes.

usr/lib/common/pkey_utils.c
```diff
@@ -1196,31 +1196,55 @@ CK_RV pkey_ec_sign(STDLL_TokData_t *tokdata, SESSION *session,
     curve_type = get_cpacf_curve_type(privkey->template);
     switch (curve_type) {
     case curve_p256:
+        if (pkey_attr->ulValueLen != sizeof(param.P256.priv) + sizeof(param.P256.vp)) {
+            TRACE_ERROR("Protected key has an invalid length of %ld bytes.\n",
+                        pkey_attr->ulValueLen);
+            ret = CKR_ATTRIBUTE_VALUE_INVALID;
+            goto done;
+        }
         if (hash_len > 32)
```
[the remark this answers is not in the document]
ifranzki: same for the other curves below

usr/lib/common/pkey_utils.c
```diff
@@ -1196,31 +1196,55 @@ CK_RV pkey_ec_sign(STDLL_TokData_t *tokdata, SESSION *session,
...
             hash_len = 32;
-        off = 32 - hash_len;
+        off = sizeof(param.P256.hash) - hash_len;
         memcpy(param.P256.hash + off, hash, hash_len);
-        memcpy(param.P256.priv, pkey_attr->pValue, 32);
-        memcpy(param.P256.vp, (char *)pkey_attr->pValue + 32, 32);
+        memcpy(param.P256.priv, pkey_attr->pValue,
```
ifranzki, line 1209: Since you checked above that `pkey_attr->ulValueLen` is exactly `sizeof(param.P256.priv) + sizeof(param.P256.vp)` you could simply copy `sizeof(param.P256.priv)` bytes here. No need for the calculation....

Similar below for the other curves

usr/lib/common/pkey_utils.c
```diff
@@ -1196,31 +1196,55 @@ CK_RV pkey_ec_sign(STDLL_TokData_t *tokdata, SESSION *session,
...
+               pkey_attr->ulValueLen - sizeof(param.P256.vp));
+        memcpy(param.P256.vp, (char *)pkey_attr->pValue + sizeof(param.P256.priv),
+               sizeof(param.P256.vp));
         *sig_len = 2 * 32;
```
ifranzki, line 1213: `*sig_len = 2 * sizeof(param.P256.sig_r)` here ? 

same for the other curves below.

jschmidb: ok, fixed and re-pushed.

ifranzki (approved):
lgtm

PR #801, opened 2024-08-05
Add support for CKA_ALWAYS_AUTHENTICATE

ifranzki:
This PR adds support for the CKA_ALWAYS_AUTHENTICATE attribute.
CKA_ALWAYS_AUTHENTICATE can be set for a private key object (i.e. an object of class CLO_PRIVATE_KEY), and if it si TRUE, then the user must authenticate on each 'use' of that key. 

From https://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.html#_Toc416959719:

> The CKA_ALWAYS_AUTHENTICATE attribute can be used to force re-authentication (i.e. force the user to provide a PIN) for each use of a private key. “Use” in this case means a cryptographic operation such as sign or decrypt. This attribute may only be set to CK_TRUE when CKA_PRIVATE is also CK_TRUE.
> 
> Re-authentication occurs by calling C_Login with userType set to CKU_CONTEXT_SPECIFIC immediately after a cryptographic operation using the key has been initiated (e.g. after C_SignInit). In this call, the actual user type is implicitly given by the usage requirements of the active key. If C_Login returns CKR_OK the user was successfully authenticated and this sets the active key in an authenticated state that lasts until the cryptographic operation has successfully or unsuccessfully been completed (e.g. by C_Sign, C_SignFinal,..). A return value CKR_PIN_INCORRECT from C_Login means that the user was denied permission to use the key and continuing the cryptographic operation will result in a behavior as if C_Login had not been called. In both of these cases the session state will remain the same, however repeated failed re-authentication attempts may cause the PIN to be locked. C_Login returns in this case CKR_PIN_LOCKED and this also logs the user out from the token. Failing or omitting to re-authenticate when CKA_ALWAYS_AUTHENTICATE is set to CK_TRUE will result in CKR_USER_NOT_LOGGED_IN to be returned from calls using the key. C_Login will return CKR_OPERATION_NOT_INITIALIZED, but the active cryptographic operation will not be affected, if an attempt is made to re-authenticate when CKA_ALWAYS_AUTHENTICATE is set to CK_FALSE.

Yes, it is strange that this is attribute defined only for private keys, but not also for secret keys, although one can also decrypt (AES, 3DES, etc) and sign (HMAC) with a secret key. Nevertheless, the PKCS#11 standard only defines this attribute for private keys. 

Furthermore, the re-authentication only works for operations that have an C_XXXInit call, becaue the C_Login must happen after the C_XXXInit and before any C_XXX, C_XXXUpdate, or C_XXXFinal calls. However operations like Wrap/Unwrap and Derive do not have an Init call, so those operations can be performed without extra authentication.

Anyway that's how the PKCS#11 standard defines it.

### DIFF
--- a/usr/include/pkcs11types.h
+++ b/usr/include/pkcs11types.h
@@ -314,6 +314,8 @@ typedef CK_ULONG CK_USER_TYPE;
 #define CKU_SO    0
 /* Normal user */
 #define CKU_USER  1
+/* Context specific */
+#define CKU_CONTEXT_SPECIFIC    2
 
 
 /* CK_STATE enumerates the session states */
@@ -563,8 +565,8 @@ typedef CK_ULONG CK_ATTRIBUTE_TYPE;
 #define CKA_EC_PARAMS          0x00000180
 #define CKA_EC_POINT           0x00000181
 /* The following are new for v2.11 */
-#define CKA_SECONDARY_AUTH     0x00000200
-#define CKA_AUTH_PIN_FLAGS     0x00000201
+#define CKA_SECONDARY_AUTH     0x00000200 /* Deprecated since 2.11 */
+#define CKA_AUTH_PIN_FLAGS     0x00000201 /* Deprecated since 2.11 */
 #define CKA_ALWAYS_AUTHENTICATE 0x00000202
 #define CKA_WRAP_WITH_TRUSTED  0x00000210
 #define CKA_HW_FEATURE_TYPE    0x00000300
--- a/usr/lib/common/new_host.c
+++ b/usr/lib/common/new_host.c
@@ -2322,7 +2322,7 @@ CK_RV SC_DecryptInit(STDLL_TokData_t *tokdata, ST_SESSION_HANDLE *sSession,
 
     sess->decr_ctx.count_statistics = TRUE;
     rc = decr_mgr_init(tokdata, sess, &sess->decr_ctx, OP_DECRYPT_INIT,
-                       pMechanism, hKey, TRUE);
+                       pMechanism, hKey, TRUE, TRUE);
     if (rc != CKR_OK)
         TRACE_DEVEL("decr_mgr_init() failed.\n");
 
@@ -2816,7 +2816,7 @@ CK_RV SC_SignInit(STDLL_TokData_t *tokdata, ST_SESSION_HANDLE *sSession,
 
     sess->sign_ctx.count_statistics = TRUE;
     rc = sign_mgr_init(tokdata, sess, &sess->sign_ctx, pMechanism, FALSE, hKey,
-                       TRUE);
+                       TRUE, TRUE);
     if (rc != CKR_OK)
         TRACE_DEVEL("sign_mgr_init() failed.\n");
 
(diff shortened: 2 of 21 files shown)


ifranzki:
The Travis failures are unrelated, as with all PRs at the moment .....

usr/lib/common/new_host.c
```diff
@@ -1508,17 +1540,118 @@ CK_RV SC_Login(STDLL_TokData_t *tokdata, ST_SESSION_HANDLE *sSession,
...
+        break;
+
+    case CKU_CONTEXT_SPECIFIC:
+        if (*flags & CKF_USER_PIN_LOCKED) {
+            TRACE_ERROR("%s\n", ock_err(ERR_PIN_LOCKED));
+            rc = CKR_PIN_LOCKED;
+            goto done;
+        }
+
+        /* Check if token has a specific handler for this, otherwise
```
jschmidb, line 1555: standard multiline comment?

usr/lib/ep11_stdll/new_host.c
```diff
@@ -1493,24 +1525,108 @@ CK_RV SC_Login(STDLL_TokData_t *tokdata, ST_SESSION_HANDLE *sSession,
         }
 
         rc = load_masterkey_so(tokdata);
-        if (rc != CKR_OK)
+        if (rc != CKR_OK) {
             TRACE_DEVEL("Failed to load SO's masterkey.\n");
+        }
+        break;
+    case CKU_CONTEXT_SPECIFIC:
```
jschmidb, line 1532: Looks like this code block is identical in common/new_host.c and ep11stdll/new_host.c. Is there a way to move this to a common subroutine?

usr/include/pkcs11types.h
```diff
@@ -563,8 +565,8 @@ typedef CK_ULONG CK_ATTRIBUTE_TYPE;
 #define CKA_EC_PARAMS          0x00000180
 #define CKA_EC_POINT           0x00000181
 /* The following are new for v2.11 */
-#define CKA_SECONDARY_AUTH     0x00000200
-#define CKA_AUTH_PIN_FLAGS     0x00000201
+#define CKA_SECONDARY_AUTH     0x00000200 /* Deprecated since 2.11 */
```
jschmidb, line 568: I also found the note that these attributes are now deprecated. But I didn't find any place saying they are deprecated since 2.11. So they were introduced in 2.11 and then immediately deprecated after 2.11? Interesting.

jschmidb (approved):
Besides the few minor comments, lgtm.

usr/lib/ep11_stdll/new_host.c
ifranzki: EP11's new_host.c is a copy of the common new_host.c (similar for ICSF's new_host.c). So they are 95% equal.... I don't like this too much, but extracting everything common into subroutines is a major task. I have it on my long-term TODO list, but I am not sure if this will ever be done :-) 
Currently the goal is to keep these files as equal as possible, where only the token specific things differ.

usr/include/pkcs11types.h
ifranzki: In v2.20 header file it states:

```
/* CKA_SECONDARY_AUTH, CKA_AUTH_PIN_FLAGS,
 * are new for v2.10. Deprecated in v2.11 and onwards. */
#define CKA_SECONDARY_AUTH     0x00000200
#define CKA_AUTH_PIN_FLAGS     0x00000201
```

ifranzki:
> Besides the few minor comments, lgtm.

No further typos, not even in commit description? How can this be ? ;-)

jschmidb:
> No further typos, not even in commit description? How can this be ? ;-)
Ha ha, yes I didn't see any

PR #815, opened 2024-10-29
proposed patch to fix resource leaks

ngothan:
There are 2 resource leaks in opencryptoki-3.24.0 found by covscan:

```
1. Defect type: RESOURCE_LEAK
  4. opencryptoki-3.24.0/usr/sbin/pkcscca/pkcscca.c:740:5: alloc_fn: Storage is returned from allocation function "malloc".
  5. opencryptoki-3.24.0/usr/sbin/pkcscca/pkcscca.c:740:5: var_assign: Assigning: "new_key->opaque_attr" = storage returned from "malloc(attrs[0].ulValueLen)".
  7. opencryptoki-3.24.0/usr/sbin/pkcscca/pkcscca.c:748:5: noescape: Resource "new_key->opaque_attr" is not freed or pointed-to in "memcpy". [Note: The source code implementation of the function has been overridden by a builtin model.]
  9. opencryptoki-3.24.0/usr/sbin/pkcscca/pkcscca.c:752:9: leaked_storage: Freeing "new_key" without freeing its pointer field "opaque_attr" leaks the storage that "opaque_attr" points to.
   750|       if (!new_key->label) {
   751|           print_error("Malloc of %lu bytes failed!", attrs[2].ulValueLen + 1);
   752|->         free(new_key);
   753|           return 2;
   754|       }

2. Defect type: RESOURCE_LEAK
   15. opencryptoki-3.24.0/usr/lib/common/mech_ec.c:1140:5: alloc_arg: "object_mgr_create_skel" allocates memory that is stored into "temp_obj".
   21. opencryptoki-3.24.0/usr/lib/common/mech_ec.c:1182:5: leaked_storage: Variable "temp_obj" going out of scope leaks the storage it points to.
     1180|       free(derived_key);
     1181|
     1182|->     return rc;
     1183|   }
     1184|
```
@ifranzki ,  i made a proposed patch

### DIFF
--- a/usr/lib/common/mech_ec.c
+++ b/usr/lib/common/mech_ec.c
@@ -1152,6 +1152,8 @@ CK_RV ecdh_pkcs_derive(STDLL_TokData_t *tokdata, SESSION *sess,
         TRACE_ERROR("template_update_attribute failed\n");
         free(value_attr);
         free(vallen_attr);
+        object_free(temp_obj);
+        temp_obj = NULL;
         goto end;
     }
 
@@ -1160,6 +1162,8 @@ CK_RV ecdh_pkcs_derive(STDLL_TokData_t *tokdata, SESSION *sess,
         if (rc != CKR_OK) {
             TRACE_ERROR("template_update_attribute failed\n");
             free(vallen_attr);
+            object_free(temp_obj);
+            temp_obj = NULL;
             goto end;
         }
     }
--- a/usr/lib/common/mech_dh.c
+++ b/usr/lib/common/mech_dh.c
@@ -124,6 +124,8 @@ CK_RV dh_pkcs_derive(STDLL_TokData_t *tokdata,
     if (rc != CKR_OK) {
         TRACE_ERROR("template_update_attribute failed\n");
         free(new_attr);
+        object_free(temp_obj);
+        temp_obj = NULL;
         return rc;
     }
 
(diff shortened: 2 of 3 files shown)


ngothan:
@jschmidb

usr/lib/common/mech_ec.c
```diff
@@ -1178,6 +1178,8 @@ CK_RV ecdh_pkcs_derive(STDLL_TokData_t *tokdata, SESSION *sess,
 
 end:
     free(derived_key);
+    if (temp_obj)
```
ifranzki, line 1181: We can not unconditionally free `temp_obj` here. When `object_mgr_create_final()` succeeds, then `temp_obj` must NOT be freed, because it has then be added to the btree and thus must stay available. 

I would suggest to free `temp_obj` and set `temp_obj` to NULL in lines 1153 and 1162 just like it is done in line 1172.  So just free it in the error case before going to `end`.

ifranzki (changes requested):

usr/lib/common/mech_ec.c
ngothan: You're right, I've improved the patch. Could you please have a look?
Thank you!

ngothan: It's done. Could you please have a look?
Thank you!

ifranzki (approved):
looks good now.

ifranzki:
There is probably a similar error in `dh_pkcs_derive()` in mech_dh.c, there is also an error exit after `object_mgr_create_skel()`. I'll take a look later and fix that, too (unless you want to do it as well?).

ngothan:
I also fixed that in latest commit.
Thank you!

ifranzki:
Thanks. Looks good, too.

PR #817, opened 2024-11-11
opencryptoki.conf: Fix typo and enhance example slot definition

ifranzki:


jschmidb (approved):
;-) another typo we always oversaw

PR #818, opened 2024-11-12
testcases: Wrapping CCA AES CIPHER keys can not work, skip it

ifranzki:
When the CCA token is configure to use CCA AES CIPHER keys, then wrapping such keys is not possible, and an attempt to do so fails with CKR_KEY_NOT_WRAPPABLE. Don't treat this as an error but skip the test.


jschmidb (approved):
Besides of a small typo in the commit msg lgtm.

PR #819, opened 2024-11-18
EP11: Fix handling of bPrepend flag in CK_IBM_KYBER_PARAMS

ifranzki:
The bPrepend flag in CK_IBM_KYBER_PARAMS can not be used as an indication if the hSecret key handle is valid or not. The bPrepend flag only determines if the hybrid secret is prepended to the Kyber secret, or appended.

To check if a hybrid secret is specified check if the hSecret handle is CK_INVALID_HANDLE or a valid object handle value.


jschmidb (approved):
lgtm

PR #820, opened 2024-11-18
SOFT: Correct check for PQC algorithm support

ifranzki:
The newest OQS-provider does no longer support the round 2 and 3 dilithium algorithms, only the standardized variants (i.e. ML-DSA). However, the current check only checked if the algorithm OID is known by OpenSSL (which is still the case), but not if the algorithm is really supported.

Add a check to fetch the algorithm to verify if the algorithm is really supported to avoid failures.

Don't report the CKM_IBM_DILITHIUM mechanism supported if none of the round 3 variants are supported by the OQS-provider.

Note that OpenCryptoki currently does not yet support the standardized ML-DSA algorithms, only the round 2 and 3 variants.

### DIFF
--- a/usr/lib/soft_stdll/soft_specific.c
+++ b/usr/lib/soft_stdll/soft_specific.c
@@ -276,6 +276,9 @@ CK_RV token_specific_init(STDLL_TokData_t *tokdata, CK_SLOT_ID SlotNumber,
                           char *conf_name)
 {
     struct soft_private_data *soft_private;
+#if OPENSSL_VERSION_PREREQ(3, 0)
+    const struct pqc_oid *oid;
+#endif
     CK_RV rc;
 
     UNUSED(conf_name);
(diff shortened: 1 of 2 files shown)


usr/lib/soft_stdll/soft_specific.c
```diff
@@ -304,12 +307,22 @@ CK_RV token_specific_init(STDLL_TokData_t *tokdata, CK_SLOT_ID SlotNumber,
...
     soft_private->oqs_provider = OSSL_PROVIDER_load(NULL, "oqsprovider");
     if (soft_private->oqs_provider == NULL) {
         TRACE_DEVEL("OSSL_PROVIDER_load for 'oqsprovider' failed, no quantum "
                     "safe mechanisms are supported.\n");
         ERR_pop_to_mark();
+    } else {
+        oid = find_pqc_by_keyform(dilithium_oids,
+                                  CK_IBM_DILITHIUM_KEYFORM_ROUND3_44);
+        if (oid == NULL || openssl_get_pqc_oid_name(oid) == NULL) {
+            OSSL_PROVIDER_unload(soft_private->oqs_provider);
```
jschmidb, line 323: We could add a TRACE_DEVEL msg here, similar to the one above in line 316, saying that the oqs_provider could be loaded, but does not support round 3 variants.

ifranzki: Good idea! Added and force pushed.

PR #823, opened 2024-11-27
Misc fixes

ifranzki:


usr/lib/cca_stdll/cca_specific.c
```diff
@@ -11497,11 +11497,12 @@ CK_RV token_specific_ecdh_pkcs_derive_kdf(STDLL_TokData_t *tokdata,
...
         TRACE_ERROR("CSNDEDH (EC Diffie-Hellman) failed. return:%ld,"
                     " reason:%ld\n", return_code, reason_code);
+        rc = CKR_FUNCTION_FAILED;
         if (return_code == 8 &&
             reason_code >= 2243 && reason_code <= 2246)
             rc = CKR_KEY_FUNCTION_NOT_PERMITTED;
-        else
-           rc = CKR_FUNCTION_FAILED;
+        if (return_code == 8 && reason_code == 90)
+            rc = CKR_FUNCTION_CANCELED; /* Control point prohibits function */
```
jschmidb, line 11505: Is there a space char before the rc?
And: is there a special reason for rc CKR_FUNCTION_CANCELED? Why not CKR_ACTION_PROHIBITED, i.e. "prohibited by ACP setting"?

ifranzki: > Is there a space char before the rc?

Yes there (now) is, it is highlighted by github because before there were only 3 spaces before rc... ;-) 

> is there a special reason for rc CKR_FUNCTION_CANCELED? ...

Yes, there are multiple reasons :-) 

First of all this is what EP11 retunes for the same situation. So it makes sense that the CCA token behaves the same.
Second, there is no really good PKCS#11 return code for such cases. 

CKR_ACTION_PROHIBITED has a very district meaning in the PKCS#11 standard and is limited to specific functions: 
_This value can only be returned by C_CopyObject, C_SetAttributeValue and C_DestroyObject. It denotes that the action may not be taken, either because of underlying policy restrictions on the token, or because the object has the the relevant CKA_COPYABLE, CKA_MODIFIABLE or CKA_DESTROYABLE policy attribute set to CK_FALSE._

CKR_FUNCTION_CANCELED of course also has a distinct meaning, but ist much more special, so reusing this one does not cause too much trouble:
_CKR_FUNCTION_CANCELED: The function was canceled in mid-execution. This happens to a cryptographic function if the function makes a CKN_SURRENDER application callback which returns CKR_CANCEL (see CKR_CANCEL). It also happens to a function that performs PIN entry through a protected path. The method used to cancel a protected path PIN entry operation is device dependent._

CKR_FUNCTION_REJECTED would also be a candidate, but it does not fit either:
_CKR_FUNCTION_REJECTED: The signature request is rejected by the user._

Long story short: I stick with what is done by EP11.....

jschmidb (approved):
ok, looks good then.

PR #826, opened 2024-12-05
CCA: Fix endless retry loop in cca_pkey_skey2pkey

ifranzki:
In case the pkey ioctl fails, but no MK change is active, break the loop also, otherwise we end up in an endless retry loop.

Fixes: 1d87606113c8ccc673fdc70436da6a94c7b981db


vkanjur (approved):
Looks good

jschmidb (approved):
lgtm

PR #828, opened 2024-12-12
Avoid incorrect copyright claims

bmwiedemann:
When building this in 2025, it claims Copyright 2025 without any copyright-worthy work happening in that year.

Also note that the number does not have to be updated every year as copyright does not expire for 70+ years.

This patch was done while working on [reproducible builds](https://reproducible-builds.org/) for openSUSE, sponsored by the NLnet NGI0 fund.

### DIFF
--- a/tools/tableidxgen.c
+++ b/tools/tableidxgen.c
@@ -596,8 +596,7 @@ static void generatelicense(FILE *fp)
     /* Remove trailing newline character */
     tstr[strlen(tstr) - 1] = 0;
     fputs("/*\n", fp);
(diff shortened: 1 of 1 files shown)


ifranzki (approved):
Yes, makes sense to hard-code the copyright year here.

ifranzki (changes requested):
Can you please add a `Signed-off-by:` tag to your commit, as described here: 
https://github.com/opencryptoki/opencryptoki/blob/master/CONTRIBUTING.md#making-changes
Other than that, your commit looks good.

bmwiedemann:
done.

ifranzki (approved):
Looks good now.

PR #829, opened 2024-12-12
Allow to override build date with SOURCE_DATE_EPOCH

bmwiedemann:
Allow to override build date with `SOURCE_DATE_EPOCH` to make builds reproducible.
See https://reproducible-builds.org/ for why this is good and https://reproducible-builds.org/specs/source-date-epoch/ for the definition of this variable.

Alternative: drop the timestamp

This patch was done while working on reproducible builds for openSUSE, sponsored by the NLnet NGI0 fund.

### DIFF
--- a/tools/tableidxgen.c
+++ b/tools/tableidxgen.c
@@ -589,8 +589,11 @@ static void generatelicense(FILE *fp)
     time_t t;
     struct tm *tm;
     char *tstr;
+    char *source_date_epoch;
 
-    t = time(NULL);
+    if ((source_date_epoch = getenv("SOURCE_DATE_EPOCH")) == NULL ||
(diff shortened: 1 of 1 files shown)


ifranzki (approved):
Looks good.

I would be also OK to drop the timestamp. I would like to still keep the `"AUTO-GENERATED, DO NOT EDIT!"` note anyway, but without a timestamp.

Let me know if you want to keep it that way, or remove the timestamp.

bmwiedemann:
I opened #832 for the alternative. Please merge the one you prefer. You are the maintainers.

ifranzki:
I think we go with https://github.com/opencryptoki/opencryptoki/pull/832. Why making things more complicated than necessary.

PR #831, opened 2024-12-12
TRAVIS: Set 'pipefail' to catch the return code of the testcases

ifranzki:
Running the testcases via ock_tests.sh uses a pipe construct to adjust the output. Without 'set -o pipefail' the return code of this is the return code of the last command in the pipe, but not the return code of ock_tests.sh.


ifranzki:
The Travis failure is intended, because now it uncovers a testcase failure that was not discovered without this change.

PR #832, opened 2024-12-12
Drop timestamp from autogenerated files

bmwiedemann:
for reproducible builds

This is an alternative to #829


ifranzki (approved):

ifranzki:
The Travis failures are unrelated (and kind of expected at the moment).

PR #834, opened 2024-12-16
COMMON: Fix AES-KEY-WRAP with OpenSSL 1.1.1

ifranzki:
With OpenSSL 1.1.1 flag EVP_CIPHER_CTX_FLAG_WRAP_ALLOW must be set in the cipher context before initializing the cipher (not after), when using EVP_aes_nnn_wrap() or EVP_aes_nnn_wrap_pad() ciphers.

Since OpenSSl 3.x the flag is no longer required, but can still be set (for compatibility reasons).


ifranzki:
Travis run on OpenSSL 1.1.1 shows `No space left on device` during build.... I don't think that this is related to the change in this PR. Will give it a try again later.

ifranzki:
Travis now runs clean.

vkanjur (approved):
looks good to me

PR #836, opened 2025-01-24
COMMON: Fix unwrap private keys to supply correct public attributes

ifranzki:
When unwrapping a private key, the public attributes are also supplied to the key. However, not all public key attributes that can be extracted from the SPKI must be added to the private key.

Add a function parameter to the nnn_priv_unwrap_get_data() functions so that the caller can tell if the public attributes are to be supplied to the template of a private or public key object.


jschmidb (approved):
lgtm

PR #837, opened 2025-01-27
COMMON: Fix setting of CKA_IBM_PROTKEY_NEVER_EXTRACTABLE and CKA_NEVER_EXTRACTABLE

ifranzki:
Attribute CKA_IBM_PROTKEY_EXTRACTABLE is set to FALSE as default for key objects, so consequently set CKA_IBM_PROTKEY_NEVER_EXTRACTABLE to TRUE as default as well.

When CKA_EXTRACTABLE or CKA_IBM_PROTKEY_EXTRACTABLE is set to TRUE, then make sure CKA_NEVER_EXTRACTABLE or CKA_IBM_PROTKEY_NEVER_EXTRACTABLE respectively is set to FALSE.

### DIFF
--- a/usr/lib/cca_stdll/cca_specific.c
+++ b/usr/lib/cca_stdll/cca_specific.c
@@ -3245,7 +3245,7 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
     CK_ATTRIBUTE *sensitive_attr = NULL;
 #ifndef NO_PKEY
     CK_ATTRIBUTE *pkey_attr = NULL, *ecp_attr = NULL;
-    CK_BBOOL add_pkey_extractable = CK_FALSE;
+    CK_BBOOL add_pkey_extractable = CK_FALSE, bfalse = CK_FALSE;
 #endif
     CK_BBOOL sensitive;
     CK_BBOOL btrue = CK_TRUE;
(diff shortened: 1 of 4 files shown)


usr/lib/cca_stdll/cca_specific.c
```diff
@@ -3245,7 +3245,7 @@ CK_RV token_specific_set_attrs_for_new_object(STDLL_TokData_t *tokdata,
     CK_ATTRIBUTE *sensitive_attr = NULL;
 #ifndef NO_PKEY
     CK_ATTRIBUTE *pkey_attr = NULL, *ecp_attr = NULL;
-    CK_BBOOL add_pkey_extractable = CK_FALSE;
+    CK_BBOOL add_pkey_extractable = CK_FALSE, bfalse = FALSE;
```
jschmidb, line 3248: Looks like in the past we used CK_FALSE, but you are now consequently using just FALSE (which is the same). So you could even change the CK_FALSE in this line to FALSE, while at it. But yes, this would not directly belong to the fix.

ifranzki:
@jschmidb can you please check again? I fixed the one bug you reported via email.

jschmidb:
Checked again on my test system: the fix works. AES tests now complete with 0 fails and 0 errors.

PR #839, opened 2025-02-18
Add p11kmip, a tool for import and exporting pkcs#11 keys to a kmip server

john-craig:
The P11KMIP tool may be used to export AES secret keys from a PKCS#11 token repository to a KMIP server by wrapping them under a shared RSA key pair. This tool may also import AES secret keys from a KMIP server to a PKCS#11 token repository using the same method.

The pull request also includes testcases for the P11KMIP tool.


ifranzki (approved):
Already reviewed this code prior to this PR.

ifranzki:
Note that the build if p11kmip is disabled by default for now. It must be explicitly enabled during configure step.

PR #840, opened 2025-02-19
P11KMIP: Add KMIP client source files from s390-tools

ifranzki:
Build the KMIP client library as part of Opencryptoki, and don't depend on libkmipclient from the s390-tools package. The s390-tools package is only available on the s390x architecture, but Opencryptoki can be built on other architectures as well.

Disable the use of HTTPS transport as well as JSON and XML encoding. The p11kmip tool only uses TTLV encoding over plain TLS.


ifranzki:
@john-craig FYI

ifranzki:
With this the build of p11kmp is enabled by default, and it can be built on any platform that Opencryptoki supports.

PR #843, opened 2025-03-07
EP11: Check session-bound keys when CKA_TOKEN or CKA_PRIVATE is changed

ifranzki:
In case attributes CKA_TOKEN and/or CKA_PRIVATE are changed (only allowed during C_CopyObject, but not via C_SetAttributeValues), special checking is required in regards of session bound keys. In case strict session mode, VHSM mode or FIPS session mode is active, the changed attributes may cause the key object to be expected to be bound to a different EP11 session as it currently is. Check if the key object is bound to the expected session, and fail if not. This prevents to have inconsistent sessions of key objects.

Note: As of now, EP11 does not support/allow to change key objects in regards of the session they are bound to, so the only option currently is to fail.

### DIFF
--- a/usr/lib/ep11_stdll/ep11_specific.c
+++ b/usr/lib/ep11_stdll/ep11_specific.c
@@ -15334,6 +15334,59 @@ retry:
     return CKR_OK;
 }
 
+/**
+ * This function is called when the attributes of a key object are changed, or
+ * if the object is copied by means of C_CopyObject. In case attributes
+ * CKA_TOKEN and/or CKA_PRIVATE are changed (only allowed during C_CopyObject),
+ * special checking is required in regards of session bound keys.
+ * In case strict session mode, VHSM mode or FIPS session mode is active,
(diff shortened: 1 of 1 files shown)


usr/lib/ep11_stdll/ep11_specific.c
```diff
@@ -15334,6 +15334,59 @@ static CK_RV ep11tok_setup_single_target(STDLL_TokData_t *tokdata,
     return CKR_OK;
 }
 
+/**
+ * This function is called when the attributes of a key object are changed, or
+ * if the object is copied by means of C_CopyObject. In case attributes
+ * CKA_TOKEN and/or CKA_PRIVATE is changed (only allowed during C_CopyObject),
```
jschmidb, line 15340: are changed

jschmidb (approved):
lgtm

usr/lib/ep11_stdll/ep11_specific.c
ifranzki: fixed

PR #845, opened 2025-03-14
TESTCASES: Fix KMIP server setup for p11kmip test

ifranzki:
Fix the logic to re-get the login authorization ID from SKLM in case the login token was expired. This can happen if another system concurrently obtains a login authorization ID from the same server. This revokes all previously login tokens.

Thus, in case a operation with SKLM returns message CTGKM6004E, a new login token must be obtained, and the failed setup step must be repeated with the new login token.

While at it, make sure that the retry loop terminates in case of unrecoverable error situations and make sure RC is set in all such cases.


jschmidb (approved):
lgtm

PR #852, opened 2025-04-03
CCA: Adjust CCA host library version detection for newer CCA versions

ifranzki:
Newer CCA versions might report the version string with CSUACFV or CSUACFQ with keyword STATCCA using a different indicator character after the version information. Ignore the indication character and the remaining data entirely. Only the version information as such is of interest.


jschmidb (approved):
lgtm

PR #853, opened 2025-04-07
EP11: Support key blob import with multiple domains in a Secure Execution guest

ifranzki:
When a Secure Execution guest has multiple domains of the crypto cards, then there is an Ultravisor session per domain. Thus it can not be just checked that the key blob to import is bound the 'the one' UV session, but it can be any of the UV sessions of the domains.

Generate a temporary AES key on each APQN, and check if at least one of the sessions of that temporary keys match the session of the blob to import.

Note that all those UV sessions are logged into all domains of the cards, so it is OK to use a key bound to one of the UV sessions on other domains as well.

### DIFF
--- a/usr/lib/ep11_stdll/ep11_specific.c
+++ b/usr/lib/ep11_stdll/ep11_specific.c
@@ -5248,6 +5248,79 @@ out:
     return rc;
 }
 
+struct check_uv_session_data {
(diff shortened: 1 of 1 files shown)


jschmidb (approved):
lgtm

PR #854, opened 2025-04-15
EP11: Handle concurrent MK change for Kyber-KEM using C_DeriveKey

ifranzki:
Special handling for mechanisms using C_DeriveKey where the mechanism parameter contains an EP11 secure key blob (i.e. Kyber-KEM): If the EP11 key blob in the mechanism parameter causes m_DeriveKey to return CKR_IBM_WKID_MISMATCH, because the master keys have been changed, then retry the whole operation including the preprocessing. This will then cause the preprocessing to set the EP11 key blob in the mechanism parameter to the re-enciphered secure key blob one.


jschmidb (approved):
lgtm

ifranzki:
Fixed wording of commit message.
Also changed the retry logic to retry only once (i.e. only when `retry_mech_param_blob == 1`). It does not make sense to retry more often. If it does not work with the re-enciphered blob on the first retry, it will never work.

PR #855, opened 2025-04-24
CCA: Fix retrieval of ACP for non-s390x platforms

ifranzki:
The output of the CSUAACM verb with the GET-ROLE keyword is in big endian format, regardless of the architecture. Use endiness conversion functions for integer type fields to make it work on little endian architectures.

Fixes: aa9e7461b70539b99f02b73bbb37997716cd0f3d

### DIFF
--- a/usr/lib/cca_stdll/cca_specific.c
+++ b/usr/lib/cca_stdll/cca_specific.c
@@ -4596,10 +4596,10 @@ static CK_BBOOL cca_pqc_strength_supported(STDLL_TokData_t * tokdata,
     struct cca_private_data *cca_private = tokdata->private_data;
 #ifdef __s390__
     const struct cca_version cca_v7_1 = { .ver = 7, .rel = 1, .mod = 0 };
-    const struct cca_version cca_v8_0 = { .ver = 8, .rel = 0, .mod = 0 };
 #else
     const struct cca_version cca_v7_2_43 = { .ver = 7, .rel = 2, .mod = 43 };
 #endif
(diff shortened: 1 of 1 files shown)


usr/lib/cca_stdll/cca_specific.c
```diff
@@ -4511,11 +4512,15 @@ static CK_RV cca_get_acp_info_handler(STDLL_TokData_t *tokdata,
...
         /* Identify default role */
-        if ((memcmp(role_name, "DFLT", 4) != 0 &&    /* since z13 */
+        if ((memcmp(role_name, "DFLT", 4) != 0 &&    /* since z13 + non-s390x */
+#if defined(__s390__)
              memcmp(role_name, "DEFALT", 6) != 0) || /* before z13 */
-            role_info->lower_time_limit != 0x0000 ||
-            (role_info->upper_time_limit != 0x0000 &&
-             role_info->upper_time_limit != 0x173b) || /* 23:59 */
+#else
+             memcmp(role_name, "DEFAULT", 7) != 0) || /* non-s390x */
```
jschmidb, line 4519: Just check: in line 4515/4517 the keyword is "DEFALT", here it is "DEFAULT".

ifranzki: Yes thats the reason for the #if statement :-) 
For s390x its 'DEFALTxx' where xx is the domain, while for non#s390 its just 'DEFAULT ' (no domain).

jschmidb (approved):
lgtm then.

PR #859, opened 2025-04-28
CCA: Fix handling of concurrent master key change for ECDH

ifranzki:
Performing the CCA CSNDEDH verb must be enclosed in a RETRY_NEW_MK_BLOB_START/RETRY_NEW_MK_BLOB_END loop to handle the situation where the EC private key token is already enciphered with the new master key, when a concurrent master key change is ongoing.

While at it, fix other minor findings in the code.

Fixes: a40e798c35ed6904f99e7a1766db916e40642505


jschmidb (approved):
lgtm

PR #863, opened 2025-05-09
CCA: Set CKA_IBM_PROTKEY_EXTRACTABLE during import of AES DATA key blob

ifranzki:
When importing a CCA secure key blob for an AES or AES-XTS key, set the attributes CKA_IBM_PROTKEY_EXTRACTABLE=TRUE and CKA_EXTRACTABLE=TRUE when the secure key blob is a CCA DATA key. CCA DATA keys are always extractable and protected key extractable.

When CIPHER keys are imported, these attributes depends on the flags from within the secure key blob, and are already set accordingly during import.

With this change, AES keys now behave the same as HMAC keys already behave. Both CKA_EXTRACTABLE and CKA_IBM_PROTKEY_EXTRACTABLE are now always set during import according to the flags of the secure key blob (and AES DATA keys are always extractable, because they can not be export restricted).

This inconsistency between import behavior of AES DATA vs CIPHER keys was noticed in the CI runs after commit b1509dd525e8823b55602b8db39a1fa70b410518. For cases where no CCA config file is available, the PKEY_MODE now defaults to DEFAULT (and no longer DISABLED) which caused the cca_ep11_export_import_test test to fail, because CKA_IBM_PROTKEY_EXTRACTABLE defaults to FALSE for the key object imported from the key blob, but AES-XTS keys are required to be protkey extractable.

The code checked the CKA_IBM_PROTKEY_EXTRACTABLE at the very beginning of the import function and failed right away because the AES-XTS key template had CKA_IBM_PROTKEY_EXTRACTABLE=FALSE, even though the code would have changed CKA_IBM_PROTKEY_EXTRACTABLE to TRUE (at least for CIPHER key) later on, because the secure key blob is flagged to be protected key extractable.

With PKEY_MODE being DISABLED (due to the bug fixed by commit b1509dd525e8823b55602b8db39a1fa70b410518) the test was skipped because the CCA token does not support the AES-XTS mechanism at all in that case, so this inconsistent behavior was not noticed.

### DIFF
--- a/usr/lib/cca_stdll/cca_specific.c
+++ b/usr/lib/cca_stdll/cca_specific.c
@@ -3688,11 +3688,6 @@ static CK_RV import_aes_xts_key(STDLL_TokData_t *tokdata,
         return CKR_MECHANISM_INVALID;
     }
 
-    if (!ccatok_pkey_check_aes_xts(object->template)) {
-        TRACE_ERROR("%s CCA AES XTS is not supported\n", __func__);
-        return CKR_TEMPLATE_INCONSISTENT;
-    }
-
     rc = template_attribute_find(object->template, CKA_IBM_OPAQUE, &opaque_attr);
     if (rc == TRUE) {
         /*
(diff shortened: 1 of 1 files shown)


jschmidb (approved):
lgtm

ifranzki:
Fixed typo in commit msg, no code change.

PR #864, opened 2025-05-14
COMMON: Always allow to modify CKA_IBM_OPAQUE[_*] attributes

ifranzki:
When an object has CKA_MODIFIABLE=FALSE, then it is essentially read-only. This means that changing attributes via C_SetAttributeValues is rejected with CKR_ACTION_PROHIBITED.

For secure key tokens, such as the CCA and EP11 tokens, the secure keys that are contained in the key objects might need to be reenciphered, when the HSM master keys are changed. For this, the offline reencipher tools 'pkcscca' and 'pkcsep11_migrate' can be used. These tools use the PKCS#11 API to iterate over the key objects, perform the secure key reencipherment and then update the secure key in CKA_IBM_OPAQUE via C_SetAttributeValues. If a key object has CKA_MODIFIABLE=FALSE this fails.

Allow to change the CKA_IBM_OPAQUE[_*] attributes via C_SetAttributeValues for secure key tokens, even if the object has CKA_MODIFIABLE=FALSE. This allows to perform offline reencryption for all objects.

### DIFF
--- a/usr/lib/common/obj_mgr.c
+++ b/usr/lib/common/obj_mgr.c
@@ -1664,6 +1664,38 @@ done:
     return rc;
 }
 
+static CK_BBOOL modifiable_override(CK_ATTRIBUTE *pTemplate, CK_ULONG ulCount)
+{
+    CK_ULONG i, k;
+    CK_BBOOL found;
+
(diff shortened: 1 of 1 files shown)


jschmidb (approved):
lgtm

PR #866, opened 2025-05-15
P11KMIP: Fix Coverity scan findings

ifranzki:
Mostly trivial changes.

@john-craig Please double check the change in function `get_kmip_usage_mask_p11()` in p11kmip.c.
The Coverity scan reported the following for the construct there:
```
CID 551165: (#1 of 1): Operands don't affect result (CONSTANT_EXPRESSION_RESULT)
result_independent_of_operands: keytype->encrypt_decrypt & (12 /* KMIP_CRY_USAGE_MASK_ENCRYPT | KMIP_CRY_USAGE_MASK_DECRYPT */) is always 0 regardless of the values of its operands. This occurs as the bitwise first operand of "|".
```

I fixed this using a series of explicit if statements. Please check if this produces the intended result.

### DIFF
--- a/usr/sbin/p11kmip/kmipclient/attribute.c
+++ b/usr/sbin/p11kmip/kmipclient/attribute.c
@@ -2432,7 +2432,7 @@ out:
 	kmip_node_free(iv);
 	kmip_node_free(tag);
 	kmip_node_free(ffl);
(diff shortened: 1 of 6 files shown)


jschmidb (approved):
yes, looks reasonable.

PR #869, opened 2025-05-27
EP11: Reject strict-session of VHSM mode in a Secure Execution guest

ifranzki:
Strict-session of VHSM mode uses the m_Login() function to login the respective EP11 sessions. In an Secure Execution for Linux guest only m_LoginExtended() with XCP_LOGIN_ALG_F2021 is supported, and can be correctly intercepted by the Ultravisor.

Reject the use of strict-session of VHSM mode when running in a Secure Execution for Linux guest.


jschmidb (approved):
lgtm

PR #872, opened 2025-06-04
EP11: Version checks for strict-session or VHSM mode in a Secure Execution guest

ifranzki:
In an Secure Execution guest strict-session or VHSM mode only works when using an EP11 host library version 4.1 or later and when all APQNs configured for the token support API ordinal 6 or later.

An EP11 host library version 4.1 or later translates a m_Login() call into a m_LoginExtended() call with XCP_LOGIN_ALG_PRE_F2021 when the API ordinal used is >= 6 and this can be correctly intercepted by the Ultravisor.

Only reject the use of strict-session or VHSM mode in a Secure Execution guest if the requirements are not fulfilled.

Fixes: f24ae99bc150bb9556370bca3668670549c5fafe


jschmidb (approved):
lgtm

PR #873, opened 2025-06-17
Allow a token to provide token info like hardware or firmware version or serial number and do so for CCA, Soft and ICA token

ifranzki:


usr/lib/ica_s390_stdll/ica_specific.c
```diff
@@ -6661,3 +6661,25 @@ CK_RV token_specific_object_add(STDLL_TokData_t *tokdata, SESSION *sess,
...
+
+CK_RV token_specific_get_token_info(STDLL_TokData_t *tokdata,
+                                    CK_TOKEN_INFO_PTR pInfo)
+{
+    libica_version_info ver;
+    int rc;
+
+    UNUSED(tokdata);
+
+    rc = ica_get_version(&ver);
```
jschmidb, line 6673: Hmm, just a thought ... you could obtain the libica fips status via ica_fips_status() and put this info into the token info as well ...

vkanjur (approved):
looks ok to me

PR #874, opened 2025-06-23
Support larger RSA key sizes

ifranzki:


jschmidb (approved):
lgtm

vkanjur (commented):
Do we need to update p11sak_test.sh to include the key size changes?

PR #876, opened 2025-06-27
Correct mechanism flags für EC mechanisms

ifranzki:
The PKCS#11 standard states that '_A Cryptoki library that can perform EC mechanisms must set either or both of these flags for each EC mechanism_'. Make sure these flags are set correctly for all tokens.


vkanjur (approved):
lgtm

Issue #881, opened 2025-07-31
Bits/bytes confusion in AES-CTR implementation for software token


MatthiasValvekens:
Hi,

There appears to be a silly bug in the handling of `counter_width` for the software token.

- At the PKCS#11 level, `counter_width` is expressed in bits (as per the spec): https://github.com/opencryptoki/opencryptoki/blob/2fe148560aa9934142de5b16bc470714e4149398/usr/lib/common/mech_aes.c#L4075-L4077
- ...but the software token implementation with openssl includes this check, which ostensibly seems to assume that `counter_width` is expressed in bytes:
https://github.com/opencryptoki/opencryptoki/blob/master/usr/lib/common/mech_openssl.c#L3973-L3976

This check ends up off by a factor 8: if the width of the counter is 16 _bits_ or less, the check passes (by accident) and the resulting output appears to be correct.

FWIW: I found out about this issue by running the test vectors from RFC 3686 through opencryptoki in https://github.com/pyauth/python-pkcs11.

PR #882, opened 2025-08-01
COMMON: Fix AES-CTR counter width range checking

ifranzki:
The range check for the AES-CTR counter width is off by a factor of 8.

According to the PKCS#11 standard, counter width is expressed in bits, and the range is supposed to be: 0 < ulCounterBits ≤ 128.

This affects the soft token, as well as the ICA token if the software fallback path is taken.

Closes: https://github.com/opencryptoki/opencryptoki/issues/881


vkanjur (approved):
lgtm

PR #886, opened 2025-08-08
COMMON: Fix decoding of DH private key

ifranzki:
Use the correct object ID length as offset to the alg-id parameters.

Fixes: aa32d071cd46 ("EP11: Add key import for DH keys")


vkanjur (approved):
looks good

PR #896, opened 2025-10-15
EP11: Fix concurrent MK change for Kyber-KEM using C_DeriveKey

ifranzki:
When the operation is repeated because of CKR_IBM_WKID_MISMATCH, then the secret key object has already been acquired and locked by the first try. Unlock the object first, before attempting to acquired and lock it again. Otherwise the object might get locked twice, but is unlocked only once, thus it stays locked after the derive processing has finished.

Fixes: 91ecd8214aafd585a26cd6a6b8ab38885911c004


vkanjur (approved):
approved.

PR #898, opened 2025-10-28
PKCSCONF: Fix memory leak in case of error and other fixes

ifranzki:
When C_GetMechanismList or C_GetMechanismInfo fails after allocating the mechanism list buffer, free the buffer before returning.

And other fixes.


vkanjur (approved):
lgtm

PR #904, opened 2025-11-10
EP11: Fix private secure key blob import

ifranzki:
Newer EP11 firmware versions do not allow to get attribute CKA_VALUE_LEN for private keys, because CKA_VALUE_LEN is not defined for these key types. This now causes error CKR_ATTRIBUTE_TYPE_INVALID. Older firmware versions ignored that and did not fail if CKA_VALUE_LEN was attempted to be retrieved.

Use separate templates for getting the attributes of public, private or secret keys, and get CKA_VALUE_LEN only for secret keys.

Fixes: 169ab762c283ca341350580c2c080720e62967e8


vkanjur (approved):
lgtm

PR #936, opened 2026-04-14
TESTCASES: Fix ECDH en-/decapsulation tests for older EP11 cards

ifranzki:
Older EP11 firmware does not support ECDH with KDFs other than CKD_NULL, or with non-empty shared data. This fails with CKR_MECHANISM_PARAM_INVALID and must be skipped as unsupported properly. This error might only be reported for the non-size-query call to C_EncapsulateKey.

Add a test vector with CKD_NULL and empty shared data to have at least one supported test for such older EP11 cards.


vkanjur (approved):
lgtm

PR #937, opened 2026-04-15
EP11: Fix syslog message printing about different CPs

ifranzki:
Syslog message "Adapter xx.yyyy has different control points than adapter xx.yyyy, using minimum." is erroneously issued even if the control points are the same.

The control points of subsequent APQNs are compared against a modified control point bit string, where the unknown CPS have already been turned ON. Comparing this with the queried control point bit string will always show differences, even if the used/known control points are the same.

Fix this by remembering the original, unmodified control point bit string instead of the modified one, and compare the control point bit strings of subsequent APQNs with that.

Fixes: 97248f73495695436f11fafd74c2ec41a5a6f796


vkanjur (approved):
lgtm

PR #938, opened 2026-04-16
CCA: Support additional RSA public exponent values

ifranzki:
Since CCA v5.2 CCA supports public exponent values of 5, 17, or 257 in addition to 0 (random), 3 or 65537.

The CCA token requires CCA v7.1 or later at minimum, so those public exponent values can be assumed to be supported in any case.


vkanjur (approved):
lgtm

PR #941, opened 2026-04-20
More PKCS#11 v3.2 enhancements

ifranzki:
New mechanisms introduced with PKCS#11 v3.2:
- Add support for CKM_ECDH_X_AES_KEY_WRAP and CKM_ECDH_COF_AES_KEY_WRAP
- Add support for CKM_PUB_KEY_FROM_PRIV_KEY


usr/lib/api/mechtable.inc
```diff
@@ -75,6 +75,8 @@ const struct mechrow mechtable_rows[] =
      { "CKM_ECDH1_COFACTOR_DERIVE",   CKM_ECDH1_COFACTOR_DERIVE,     0, MC_INFORMATION_UNAVAILABLE, MCF_DERIVE | MCF_ENDECAPS | MCF_NEEDSPARAM },
      { "CKM_ECDH1_DERIVE",            CKM_ECDH1_DERIVE,              0, MC_INFORMATION_UNAVAILABLE, MCF_DERIVE | MCF_ENDECAPS | MCF_NEEDSPARAM },
      { "CKM_ECDH_AES_KEY_WRAP",       CKM_ECDH_AES_KEY_WRAP,         0,           MC_KEY_DEPENDENT, MCF_WRAPUNWRAP },
+     { "CKM_ECDH_COF_AES_KEY_WRAP",   CKM_ECDH_COF_AES_KEY_WRAP,     0,           MC_KEY_DEPENDENT, MCF_WRAPUNWRAP },
```
vkanjur: Do we need MCF_NEEDSPARAM too?

usr/lib/api/mechtable.inc
```diff
@@ -75,6 +75,8 @@ const struct mechrow mechtable_rows[] =
...
+     { "CKM_ECDH_X_AES_KEY_WRAP",     CKM_ECDH_X_AES_KEY_WRAP,       0,           MC_KEY_DEPENDENT, MCF_WRAPUNWRAP },
```
vkanjur: Do we need MCF_NEEDSPARAM too?

vkanjur (changes requested):

usr/lib/api/mechtable.inc
ifranzki: Yes it does. Will fix,

usr/lib/api/mechtable.inc
ifranzki: Yes it does. Will fix,
Mech CKM_ECDH_AES_KEY_WRAP also needs it, will add it as well (although that was added some time back).

ifranzki: I don't think these flags are anywhere really used, but for consistency, lets make it correct.

vkanjur (approved):
lgtm

PR #942, opened 2026-04-20
Fix missing unlock errors

ifranzki:


usr/lib/ep11_stdll/ep11_session.c
vkanjur, line 1: Minor typo: unlockes -> unlocks

usr/lib/common/utility.c
```diff
@@ -222,27 +222,35 @@ CK_RV XProcLock(STDLL_TokData_t *tokdata)
 
 CK_RV XProcUnLock(STDLL_TokData_t *tokdata)
 {
+    CK_RV rc = CKR_OK;
+    CK_RV rc2;
```
vkanjur, line 226: Isn't it better to initialize rc2 also like rc?

vkanjur (changes requested):

usr/lib/common/utility.c
ifranzki: Why? rc2 is only set by XThreadUnLock below and used right afterwards. SO no need to initialize it. It can never be used uninitialized.

usr/lib/ep11_stdll/ep11_session.c
ifranzki: fixed

vkanjur (approved):
lgtm

PR #943, opened 2026-04-22
COMMON: Fix typo in ML-KEM OID define and constant

ifranzki:
External define OCK_ML_KEM_786 has a typo, it must be OCK_ML_KEM_768. Fix this, but keep the old define for backward compatibility.

Also constant ml_kem_786 has the same typo, it must be ml_kem_768. Fix this also (not externally visible).

Fixes: 1440e408f84e0ebba15eedfeb6c7191cdba239aa


vkanjur (approved):
lgtm

PR #948, opened 2026-05-22
CCA: Enable SHA3 mechanisms conditionally for Linux on x64/Power/AIX

ifranzki:
CCA for Linux on x64 and Power as well as CCA for AIX checks the SHA3 Access Control Points 0x03BD-0x03C0 when performing SHA3 operations. By default these ACPs are OFF. To enable SHA3 mechanisms for the CCA token, all 4 of them must be turned ON.

CCA on Linux on IBM Z does not check these control points.


vkanjur (approved):
lgtm

PR #950, opened 2026-06-16
ICA: Support recent changes in libica regarding FIPS 140-3

ifranzki:
When libica is in FIPS mode it might block RSA key sizes below 128 bit of security strength, which means it blocks RSA keys less than 3072 bits.

The libica method list also reports a properties bit field to indicate what RSA key sizes are supported. This was extended to also report 3072 bit RSA key sizes.

Update the ICA token to support that new RSA 3072 bit property flag, and report the supported key size for the RSA mechanisms accordingly.

While at it, use the defines provided in the libica header file, and only define them if they are not yet defined by the libica header (i.e. when built against older libica versions).


vkanjur (approved):
lgtm

PR #953, opened 2026-06-18
Update CKM_PUB_KEY_FROM_PRIV_KEY handling after PKCS#11 spec update

ifranzki:
PKCS#11 spec update https://github.com/oasis-tcs/pkcs11/commit/1f2af1234e8b891ff9a91b46261372ce401c132f clarified the behavior of CKM_PUB_KEY_FROM_PRIV_KEY in regards of the CKA_DERIVE attribute:

>    Although this is technically a derivation mechanism, it can always be used with any CKO_PRIVATE_KEY object regardless of its CKA_DERIVE attribute value.

Update the key_mgr_derive_key() function and p11sak command 'extract-pubkey' accordingly.


vkanjur (approved):
lgtm

PR #954, opened 2026-06-19
CCA: Fix required cipher text length for ML-KEM

ifranzki:
The cipher text length used for the 3 ML-KEM variants is slightly smaller than the public key length, except for ML-KEM-1024 where it is the same.

The CCA token checks the cipher text buffer length to be at least the public key length. If it is smaller it reject it with CKR_BUFFER_TOO_SMALL.

For a size query call, it would return a slightly too large size, but this does not matter. On the final call it sets the cipher text length to the correct length (which might be slightly smaller than the buffer).

However, if one performs an ML-KEM operation and supplies a buffer with the exact cipher text length as needed by the MK-KEM variant, it might fail with CKR_BUFFER_TOO_SMALL although the buffer length is actually correct.

Fix this by checking against the correct cipher text length, not the public key length.

Fixes: bf499edb51f0d26aa67d6780d22665989d5ea3f4


vkanjur (approved):
lgtm

PR #956, opened 2026-06-24
Fix build error on 32 bit systems: size_t vs CK_ULONG pointer incompatibility

### LINKED ISSUE #955: Please verify patch for 32bit systems
Hi,
I intended to update the opencryptoki Debian package.  When I tried to build on i386 the build failed.  When I apply [this patch] the build succeeds.  However, I would be really happy if you would check whether the patch makes sense and I
hope all is fine.
Kind regards, Anndreas.

ifranzki:
On i386, `size_t` is `unsigned int` while `CK_ULONG` is `unsigned long`.

Closes: https://github.com/opencryptoki/opencryptoki/issues/955


ifranzki:
@tillea FYI

vkanjur (approved):
lgtm

PR #957, opened 2026-07-01
EP11: Add support for CKM_IBM_ETH_DERIVE mechanism 

ifranzki:
The CKM_IBM_ETH_DERIVE mechanism is an IBM specific mechanism. It can be used to perform Ethereum 2.0 key derivation of BLS12-381 EC keys using the mechanism described in EIP-2333. It is used with C_DeriveKey.

Ethereum 2.0 key derivation is designed to derive a master key from a single seed. From that master key, a hierarchical tree of keys can be derived.

Three types of key derivation are supported:
  - MASTERK: A master key (a BLS private key) is derived from a generic secret key, that is used as the seed.
  - PRV2PRV: A private BLS child key is derived from a private BLS key.
  - PRV2PUB: A public BLS child key is derived from a private BLS key.


testcases/crypto/ec_func.c
```diff
@@ -3578,6 +3578,276 @@ CK_RV run_BLSAggregation(void)
...
+                      eth_child_key_index[k].index,
+                      eth_child_key_index[k].key_info_len, pkey);
+
+        /* Testcase #4: Test if derived keys are usable */
+        testcase_new_assertion();
+
+        mech.mechanism = CKM_ECDSA;
+        mech.ulParameterLen = 0;
+        mech.pParameter = NULL;
+
```
vkanjur, line 3778: Dont we need a testcase_begin() here?

usr/lib/api/policy.c
```diff
@@ -1154,6 +1154,31 @@ static CK_RV policy_is_mech_allowed(policy_t p, CK_MECHANISM_PTR mech,
...
+        case CKM_IBM_ETH_DERIVE:
+            if (mech->ulParameterLen != sizeof(CK_IBM_ETH_DERIVE_PARAMS) ||
+                mech->pParameter == NULL) {
+                TRACE_ERROR("Invalid mechanism parameter\n");
+                rv = CKR_MECHANISM_PARAM_INVALID;
+                break;
+            }
+            if (((CK_IBM_ETH_DERIVE_PARAMS *)mech->pParameter)->version !=
+                                        CK_IBM_ETH_DERIVE_PARAMS_VERSION_1)
+                break;
```
vkanjur, line 1166: Do we need to check CK_IBM_ETH_SIG_VERSION_4 and return CKR_FUNCTION_FAILED?

usr/lib/api/statistics.c
```diff
@@ -228,6 +228,26 @@ static CK_RV statistics_increment(struct statistics *statistics,
             break;
         }
         break;
+    case CKM_IBM_ETH_DERIVE:
+        if (mech->pParameter == NULL ||
+            mech->ulParameterLen != sizeof(CK_IBM_ETH_DERIVE_PARAMS))
+            return CKR_MECHANISM_PARAM_INVALID;
+        if (((CK_IBM_ETH_DERIVE_PARAMS *)mech->pParameter)->version !=
+                                        CK_IBM_ETH_DERIVE_PARAMS_VERSION_1)
```
vkanjur, line 236: Do we need to check CK_IBM_ETH_SIG_VERSION_4 here?

vkanjur (changes requested):

usr/lib/api/policy.c
ifranzki: The `version` field of the `CK_IBM_ETH_DERIVE_PARAMS` structure defines the version of the structure itself. So we do check it here to be sure to use the right structure overlay. 

The `sigVersion` field however specifies the version of the EIP-2333 specification, which is currently at version 4. So its more of the version of the algorithm internally used. I would assume that EP11 internally checks that `sigVersion` field to be valid. For the policy and for the statistics it should not make any difference. 

The only impact `sigVersion` could have is if a newer version uses a different digest that SHA256 internally. This would affect the policy and statistics code. However, I would say we only care once/if a new version is there.

usr/lib/api/statistics.c
ifranzki: see above

usr/lib/api/policy.c
vkanjur: Ok Ingo..Thanks for the clarification.

testcases/crypto/ec_func.c
ifranzki: Its a bit strange, but the `testcase_begin()` is inside the `run_GenerateSignVerifyECC()` function (but not the `testcase_new_assertion()` and not the `testcase_pass()`)....

So no `testcase_begin()` needed here.

vkanjur: Ok

vkanjur:
lgtm

vkanjur (approved):
lgtm

PR #959, opened 2026-07-22
Fixes for the kmipclient library

ifranzki:
Replicate the fixes for the kmipclient library also provided in the s390-tools project (https://github.com/ibm-s390-linux/s390-tools) to Opencryptoki.'s p11kmip tool. The code is intentionally kept almost the same as in s390-tools, except for a few #ifdefs to account for the different environment.


vkanjur (commented):
lgtm

PR #961, opened 2026-07-24
TESTCASES: Fix build warning with set but not used variable

ifranzki:
Fix the following by removing the variable and its assignments entirely. The function's actual pass/fail logic uses the global tests_failed counter.

```
testcases/unit/asn1fuzztest.c:846:9: warning: variable 'result' set but
                                     not used [-Wunused-but-set-variable]
  846 |     int result = 0;
      |         ^
```


vkanjur (approved):
lgtm

PR #963, opened 2026-08-17
Various fixes

ifranzki:


ifranzki:
Pushed directly to master.

